PR #9, opened 2019-08-06
Invoke capnp compile from mpgen

ryanofsky:
This simplifies multiprocess build code somewhat (https://github.com/bitcoin/bitcoin/pull/16367) so it only has to invoke one code generator instead of two, and is probably a little more future proof in that is should make it possible to change capnp options without any requiring changes outside libmultiprocess.


hebasto:
A commit fro the future:

Screenshot from 2022-05-01 17-27-22

:smiley:

ryanofsky:
> :smiley: 

Yeah, I messed this up, and there isn't a good way to fixi it without rewriting all the hashes in commit history. Could do if I find a convenient tool that does this.

ryanofsky:
> :smiley: 

Yeah, I messed this up, and there isn't a convenient way to fix it without rewriting all the hashes in commit history. Could do if I find a convenient tool that does this

src/mp/gen.cpp
```diff
@@ -475,13 +524,13 @@ void Generate(kj::StringPtr output_stem, kj::StringPtr input_schema, kj::ArrayPt
 int main(int argc, char** argv)
 {
     if (argc < 3) {
-        fprintf(stderr, "Usage: " PROXY_BIN " OUTPUT_STEM INPUT_SCHEMA [IMPORT_PATH...]\n");
+        fprintf(stderr, "Usage: " PROXY_BIN " SRC_PREFIX SRC_FILE [IMPORT_PATH...]\n");
```
hebasto, line 527: Should it be
```diff
-        fprintf(stderr, "Usage: " PROXY_BIN " SRC_PREFIX SRC_FILE [IMPORT_PATH...]\n");
+        fprintf(stderr, "Usage: " PROXY_BIN " SRC_PREFIX INCLUDE_PREFIX SRC_FILE [IMPORT_PATH...]\n");
```
?

ryanofsky: > Should it be

Good catch. Should be fixed in #80

PR #17, opened 2020-01-15
Avoid using boost::optional in PassField()

vasild:
Refactor PassField() to an equivalent one that does not use
boost::optional.

### DIFF
--- a/include/mp/proxy-types.h
+++ b/include/mp/proxy-types.h
@@ -955,22 +955,26 @@ void CustomBuildField(TypeList<LocalType> local_type,
 template <typename Accessor, typename LocalType, typename ServerContext, typename Fn, typename... Args>
 void PassField(TypeList<LocalType*>, ServerContext& server_context, const Fn& fn, Args&&... args)
 {
-    InvokeContext& invoke_context = server_context;
-    boost::optional<Decay<LocalType>> param;
     const auto& params = server_context.call_context.getParams();
     const auto& input = Make<StructField, Accessor>(params);
-    bool want = input.want();
-    if (want) {
-        MaybeReadField(std::integral_constant<bool, Accessor::in>(), TypeList<LocalType>(), invoke_context, input,
-            Emplace<decltype(param)>(param));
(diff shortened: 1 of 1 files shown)


vasild:
@ryanofsky as a followup to https://github.com/chaincodelabs/libmultiprocess/pull/16 I crudely stripped down `absl::optional` to about 4k lines of code (:-O). With finer stripping maybe it could go down to below 1000 lines, but it does not look good - what if a bug is introduced due to the stripping? Also, changes from upstream would be very tedious to merge and it would be one black box inside the source code. Its better to require the user to install `absl` externally.

While playing with it the patch in this PR materialized - we could remove some more boost without adding a `boost::optional` replacement.

ryanofsky:
This seems like a good approach, and I can merge this PR after a little testing. I agree that replacing optional isn't good if it's going to add a lot of new code. I was thinking a small optional class with basic emplace/get functionality could be implemented in something like ~100 lines, but if this isn't the case it's probably better to avoid adding a new optional class. And maybe it's better to avoid using optional altogether, since most of the changes you've made so far do seem like simplifications.

include/mp/proxy-types.h
```diff
@@ -955,22 +955,26 @@ void CustomBuildField(TypeList<LocalType> local_type,
...
     const auto& input = Make<StructField, Accessor>(params);
-    bool want = input.want();
-    if (want) {
-        MaybeReadField(std::integral_constant<bool, Accessor::in>(), TypeList<LocalType>(), invoke_context, input,
-            Emplace<decltype(param)>(param));
-        if (!param) param.emplace();
+
+    if (!input.want()) {
+        fn.invoke(server_context, std::forward<Args>(args)..., nullptr);
+        server_context.call_context.getResults();
```
ryanofsky: Should be safe to drop this line. Can be done in a followup though since keeping it makes the refactoring more obvious.

vasild: https://github.com/chaincodelabs/libmultiprocess/pull/18

Cheerz!

PR #21, opened 2020-01-18
Refactor ThreadName() to improve its portability

vasild:
* Test for the presence of pthread_getname_np() during cmake time and
  only use it if it is available. At least FreeBSD does not have it.
  When not available, fall back to (the first 15 chars of) the
  executable name, as if pthread_setname_np() has never been called.

* Use std::this_thread::get_id() to retrieve the thread id in a portable
  way instead of pthread_threadid_np() or syscall() (none of which is
  available on FreeBSD but it has its own pthread_getthreadid_np()).
  C++11 is here to help.

* According to pthread_getname_np(3) the thread name length is
  restricted to 16 characters, including the terminating null byte.
  So reduce our thread_name[] from 17 to 16 chars.


CMakeLists.txt
```diff
@@ -5,9 +5,19 @@
 cmake_minimum_required(VERSION 3.0)
 project("Libmultiprocess" CXX)
 include(CTest)
+include(CheckCXXSourceCompiles)
 find_package(Boost)
 find_package(CapnProto)
 
+check_cxx_source_compiles("
```
ryanofsky: This check was failing for me due to a link error (linux requires pthreads programs to be linked with a pthreads library), so I wasn't seeing thread names after this. Following fix worked for me:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 16ecf50..cf87baf 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,11 +4,15 @@
 
 cmake_minimum_required(VERSION 3.0)
 project("Libmultiprocess" CXX)
+include(CMakePushCheckState)
 include(CTest)
 include(CheckCXXSourceCompiles)
 find_package(Boost)
 find_package(CapnProto)
+find_package(Threads REQUIRED)
 
+cmake_push_check_state()
+set(CMAKE_REQUIRED_LIBRARIES Threads::Threads)
 check_cxx_source_compiles("
   #include <pthread.h>
   int main(int argc, char** argv)
@@ -17,6 +21,7 @@ check_cxx_source_compiles("
     return pthread_getname_np(pthread_self(), thread_name, sizeof(thread_name));
   }"
   HAVE_PTHREAD_GETNAME_NP)
+cmake_pop_check_state()
 
 capnp_generate_cpp(MP_PROXY_SRCS MP_PROXY_HDRS include/mp/proxy.capnp)
 
```

src/mp/util.cpp
```diff
@@ -37,14 +39,17 @@ size_t MaxFd()
 
 std::string ThreadName(const char* exe_name)
 {
-    char thread_name[17] = {0};
+    char thread_name[16] = {0};
```
ryanofsky: Note: This change looks right. Previous code was trying to make room for an extra null byte, but pthread_getname_np documentation does say 16 byte max includes the null byte like PR description says

src/mp/util.cpp
```diff
@@ -37,14 +39,17 @@ size_t MaxFd()
...
+#ifdef HAVE_PTHREAD_GETNAME_NP
     pthread_getname_np(pthread_self(), thread_name, sizeof(thread_name));
-    uint64_t tid = 0;
-#if __linux__
-    tid = syscall(SYS_gettid);
```
ryanofsky: I'd like keep using `gettid` on linux if possible. On linux, tid thread numbers are different from pthread numbers, and match up with the thread numbers shown in gdb, so it's easier to match up logs with gdb backtraces. Tid numbers are also shorter by default and easier to visually distinguish (`20080` vs `140369161336576`)

src/mp/util.cpp
```diff
@@ -37,14 +39,17 @@ size_t MaxFd()
...
 #else
-    pthread_threadid_np(NULL, &tid);
-#endif
+    // By default, all the threads inherit the program name. So use that one
+    // if we don't have pthread_getname_np().
+    snprintf(thread_name, sizeof(thread_name), "%s", exe_name);
```
ryanofsky: I think this isn't safe because exe_name can be null here. Probably better in any case just to leave thread_name empty if it's not known, so exe_name isn't repeated twice below.

CMakeLists.txt
vasild: Excellent, done!

src/mp/util.cpp
vasild: Right, the doc also says that the output name will be null terminated so it may seem ok to remove the initialization `= {0}` but I left it because I suspect that if `pthread_getname_np()` fails it may leave the name untouched.

src/mp/util.cpp
vasild: I see. Done. And also resurrected the other call to `pthread_threadid_np()` (+ a proper check for its presence).

I checked that on FreeBSD `std::this_thread::get_id()` returns some long id that is different from what gdb prints (similar to what you say for linux), but [`thr_self()`](https://www.freebsd.org/cgi/man.cgi?query=thr_self) and [`pthread_getthreadid_np()`](https://www.freebsd.org/cgi/man.cgi?query=pthread_getthreadid_np) print the same id as gdb. So I added `pthread_getthreadid_np()` too, and if none is available then fallback to `std::this_thread::get_id()`.

src/mp/util.cpp
vasild: Removed. In this case we would return something like `prog-35291/-20080` which looks a bit strange and may break some tool that tries to parse it.

ryanofsky: re: https://github.com/chaincodelabs/libmultiprocess/pull/21#discussion_r368248257

> Removed. In this case we would return something like `prog-35291/-20080` which looks a bit strange and may break some tool that tries to parse it.

I'm not too concerned about appearance of the dash, but omitting it when the thread name is empty could be a possible improvement

vasild: https://github.com/chaincodelabs/libmultiprocess/pull/24 Don't print a dash if thread name is not known

PR #22, opened 2020-01-19
Handle fork(2) failures

vasild:
Generate() and SpawnProcess() call fork(2) but did not check if it
failed (return -1) and continued execution with possible unexpected
results.

Handle fork(2) failure by throwing an exception.

### DIFF
--- a/src/mp/util.cpp
+++ b/src/mp/util.cpp
@@ -4,6 +4,7 @@
 
 #include <mp/util.h>
 
+#include <errno.h>
 #include <kj/array.h>
 #include <pthread.h>
 #include <sstream>
@@ -13,6 +14,7 @@
 #include <sys/types.h>
 #include <sys/un.h>
 #include <sys/wait.h>
+#include <system_error>
 #include <unistd.h>
 
 #if __linux__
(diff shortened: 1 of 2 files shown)


src/mp/util.cpp
```diff
@@ -83,6 +85,9 @@ int SpawnProcess(int& pid, FdToArgsFn&& fd_to_args)
     }
 
     pid = fork();
+    if (pid == -1) {
+        throw std::system_error(errno, std::system_category(), "fork");
```
ryanofsky: Interesting, I didn't know system_error took a third string argument. It'd be good to start using this more places.

vasild: Yes, it also occurred to me that if `std::system_category()` modifies `errno` then `std::system_error()` may get the wrong value of `errno`:
```
#include <errno.h>
#include <stdio.h>

int f1() {
    errno = 3;
    return 20;
}   

void f2(int e, int x) {
    printf("%d %d\n", e, x);
}   

int main(int, char**) {
    errno = 2; // resulted from fork() or whatever
    f2(errno, f1());
    return 0;
}
```
produces:
```
# clang 9.0.1
$ clang++90 t.cc -o t && ./t
2 20

# gcc 9.2.0
$ g++9 t.cc -o t && ./t
3 20
```

ryanofsky: It's a possible concern, but in practice if that constructor was setting errno, it would probably break a lot of things. Also in the worst case error message text would just be less clear

vasild: Yes, safe to ignore in this case. Just the pattern `func1(x, func2_may_possibly_modify_x());` is better to be avoided.

vasild: > system_error took a third string argument. It'd be good to start using this more places.

https://github.com/chaincodelabs/libmultiprocess/pull/23 Tell std::system_error() which function failed

Cheerz!

PR #31, opened 2020-03-20
Unify ReadFieldNew / ReadFieldUpdate

ryanofsky:
Replace with ReadField wrapper and CustomReadField hook to be consistent with
BuildField / CustomBuildField. This

- Makes it more straightforward pass custom types with symmetric
  CustomReadField / CustomBuildField overloads

- Simplifies debugging and overload resolution, removing hacks that allowed
  ReadFieldNew calls to forward to ReadFieldUpdate calls and vice versa

- Enables more flexibility and efficiency within CustomReadField
  implementations allowing emplacing or updating in the same function based on
  which is simpler or more efficient. Removes a lot of std::enable_if overload
  hacks

- Should allow clientInvoke to use RVO and construct return values in place
  instead of copying or moving. This will allow more types to be for supported
  as IPC return values even if they don't support default construction,
  copying, or moving.

- Improves separation of concerns, avoids need for ReadField callers to know
  whether callee will emplace or update, avoids need for emplace ReadDest
  object to have hardcoded support for map, vector, optional, shared_ptr,
  reference_wrapper types, so they now are implemented the same as any other
  custom type.

### DIFF
--- a/include/mp/proxy-types.h
+++ b/include/mp/proxy-types.h
@@ -175,257 +175,292 @@ auto PassField(TypeList<>, ServerContext& server_context, const Fn& fn, const Ar
         kj::mv(future.promise));
 }
 
+template <typename LocalType, typename EmplaceFn>
+struct ReadDestEmplace
+{
+    ReadDestEmplace(TypeList<LocalType>, EmplaceFn&& emplace_fn) : m_emplace_fn(emplace_fn) {}
+
+    template <typename... Args>
+    auto& construct(Args&&... args)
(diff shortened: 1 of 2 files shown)


ryanofsky:
Corresponding updates for [bitcoin](https://github.com/bitcoin-core/bitcoin-devwiki/wiki/Process-Separation) were https://github.com/ryanofsky/bitcoin/commit/cc91cff97548f24a169eea0307590be8d967c5c8

PR #37, opened 2020-05-13
Add "make check" target to build and run tests

ryanofsky:
fanquake reported https://github.com/chaincodelabs/libmultiprocess/issues/36 that instructions for running tests in the README are wrong, and there isn't a convenient way to build and run tests, so this PR adds a "make check" custom target to make it more straightforward.

The new "make check" target builds and runs all available tests, while the previous "make test" target (provided by CTest) only runs existing test binaries without building anything.

This PR also:

- Moves all test code and build rules to a new test/ subdirectory to separate test code from non-test code
- Adds a new "make tests" target used internally by "make check" to build tests without running them


fanquake:
Thanks. Tested 037835320bd28d0416913d597f1f6298e8eaea00 and that running `make check` works for me.

PR #45, opened 2021-03-15
Comments and LogEscape

jamesob:
Just a few little changes made while reviewing. Feel free to discard or change any of the documentation edits.

### DIFF
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -237,8 +237,8 @@ struct Waiter
     {
         m_cv.wait(lock, [&] {
             // Important for this to be "while (m_fn)", not "if (m_fn)" to avoid
-            // a lost-wakeup bug. A new m_fn and m_cv notification might be set
-            // after then fn() call and before the lock.lock() call in this loop
+            // a lost-wakeup bug. A new m_fn and m_cv notification might be sent
+            // after the fn() call and before the lock.lock() call in this loop
             // in the case where a capnp response is sent and a brand new
             // request is immediately received.
             while (m_fn) {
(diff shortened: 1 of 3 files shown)


src/mp/util.cpp
```diff
@@ -81,7 +81,7 @@ std::string LogEscape(const kj::StringTree& string)
     string.visit([&](const kj::ArrayPtr<const char>& piece) {
         if (result.size() > MAX_SIZE) return;
         for (char c : piece) {
-            if ('c' == '\\') {
```
ryanofsky, line 84: 🤦

include/mp/proxy-io.h
```diff
@@ -338,6 +338,13 @@ class Connection
 };
 
 //! Vat id for server side of connection. Required argument to RpcSystem::bootStrap()
+//!
+//! "Vat" is Cap'n Proto nomenclature for a host of various objects that facilitates
+//! bidirectional communication with other vats; it is often but not always 1-1 with
+//! processes. Cap'n Proto doesn't reference clients or servers per se; instead everything
+//! is just a vat.
+//!
+//! See also: https://github.com/capnproto/capnproto/blob/9021f0c722b36cb11e3690b0860939255ebad39c/c%2B%2B/src/capnp/rpc.capnp#L42-L56
```
ryanofsky, line 347: Further back, it's from E https://en.wikipedia.org/wiki/E_(programming_language)#Computational_model

src/mp/util.cpp
jamesob: Surprised there isn't a compiler warning for these kinds of things.

ryanofsky:
Thank you!

PR #69, opened 2022-08-26
Fix CMake minimum required version

hebasto:
The "C++17" value of the [`CXX_STANDARD`](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html) target property, which was introduced in chaincodelabs/libmultiprocess#25, is available in CMake 3.8 and newer.

Tested with old CMake versions available [here](https://cmake.org/files/).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.8)
 project("Libmultiprocess" CXX)
 include(CMakePushCheckState)
 include(CheckCXXSourceCompiles)


CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.8)
```
hebasto, line 5: Considering https://repology.org/project/cmake/versions, maybe 3.10 is also justified?

ryanofsky: > Considering https://repology.org/project/cmake/versions, maybe 3.10 is also justified?

I'm not actually sure what specifically here requires 3.10, but would welcome any followup PR that changes the minimum version or adds a comment explaining the current minimum.

ryanofsky (approved):
Code review ACK 6902bfd40e33a0007f75bce8af0fc0b10a147e86. Thanks for the PR!

CMakeLists.txt
hebasto: > I'm not actually sure what specifically here requires 3.10

Nothing requires 3.10. Just thinking about availability of CMake 3.8 and 3.9 on real systems.

PR #70, opened 2022-08-31
Fix check for `pthread_threadid_np`

hebasto:
Otherwise, the test always fails with "error: ‘uint64_t’ was not declared in this scope".


ryanofsky (approved):
Code review ACK 373d0d7823c39b30b318f91886f0041d581213cb. Would be curious to know what platform(s) the problem happens on, but thanks for the fix!

hebasto:
> Code review ACK 373d0d7823c39b30b318f91886f0041d581213cb. Would be curious to know what platform(s) the problem happens on, but thanks for the fix!

macOS at least.

PR #71, opened 2022-08-31
Use defined `CXX_STANDARD` for environment introspection

hebasto:
It doesn't look robust to use different C++ standards for configuration and building.

This PR, for example, will apply the `-std=gnu++17` compiler option for all checks during configuration when building with GCC.


ryanofsky (approved):
Code review ACK 2eff5da83c5dfcb13d28df7c2c108178bbcba956. The original reason this was implemented this way was that some versions of capnproto required c++14 while bitcoin was still using c++11, so there really was a mix of versions being used together. But it is simpler to set a default version now, and good to use the same version for checks.

PR #72, opened 2022-08-31
Make `mpgen` target independent from `multiprocess` one

hebasto:
Instead of dependency `mpgen` on `multiprocess`, both of them can share the `util` object library.

This is a step towards ability to build and install `mpgen` and `multiprocess` independently, especially when [cross](https://github.com/bitcoin/bitcoin/pull/25941) building.


ryanofsky (approved):
Code review ACK 665d1fdc676a9dee36e6b17261448e56f225c0e4. I like this change. No reason the code generator and runtime need to share all the same dependencies.

PR #73, opened 2022-08-31
Drop unneeded include directory

hebasto:
No headers reside in the `${CMAKE_CURRENT_SOURCE_DIR}/src>` directory.

No changes in behavior.


ryanofsky (approved):
Code review ACK d36ebb9aceaa9a72cc3dde277f4db3405e89c5e3. I'm not sure about this change in the long term since in the future maybe the code generator should have private headers that are not installed. But this is ok for now, and maybe code would move around anyway and make this obsolete.

PR #74, opened 2022-09-07
Add `install-lib` and `install-bin` build targets

hebasto:
This PR adds new `install-lib` and `install-bin` build targets which allow to build and install only required artifacts during different stages of cross-building.

A demo, which uses this branch in Bitcoin Core, is [here](https://github.com/hebasto/bitcoin/commits/220907-mp-demo).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,15 @@ target_link_libraries(multiprocess PRIVATE CapnProto::kj)
 target_link_libraries(multiprocess PRIVATE CapnProto::kj-async)
 set_target_properties(multiprocess PROPERTIES
     PUBLIC_HEADER "${MP_PUBLIC_HEADERS}")
-install(TARGETS multiprocess EXPORT Multiprocess ARCHIVE DESTINATION lib PUBLIC_HEADER DESTINATION include/mp)
+install(TARGETS multiprocess EXPORT Multiprocess
+  ARCHIVE DESTINATION lib COMPONENT lib
+  PUBLIC_HEADER DESTINATION include/mp COMPONENT lib)
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/pkgconfig/libmultiprocess.pc"
+  DESTINATION "lib/pkgconfig" COMPONENT lib)
+add_custom_target(install-lib
+  COMMAND ${CMAKE_COMMAND} -DCOMPONENT=lib -P ${CMAKE_CURRENT_BINARY_DIR}/cmake_install.cmake
+  VERBATIM)
+add_dependencies(install-lib multiprocess)
 
 add_executable(mpgen src/mp/gen.cpp $<TARGET_OBJECTS:util>)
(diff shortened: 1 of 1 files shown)


hebasto:
Drafted until Concept (N)ACK.

ryanofsky:
Few thoughts:

- PR title "Make build system cross-building-friendly" is vague and seems to imply that current build system does not support cross-building, which is untrue. Would suggest "Add codegen-only and library-only CMake build options"

- I think it would be better to add new build targets, not new build options. Ideally in addition to current `make install` target there would be new `make install-lib` and `make install-bin` targets that automatically build and install the minimal dependencies. Current approach confuses the role of CMake and Make. Cmake is great at determining _how_ things should be built but Make does a much better job of determining _what_ to build. If you go down the road of using CMake to decide what to build, it is a big mistake because instead of specifying dependencies explicitly you specify them implicitly in nest of `IF()` statements that don't scale well and can become unmaintainable: wastefully building more dependencies than the user actually requires, or failing to rebuild needed dependencies when different combination of options are used. It is much better to let Make figure out what to build automatically, than to try to do it manually in CMake with `IF()` statements.
  
  Instead of changing the build rules, I think it would would be better to take advantage of cmake install components and `add_custom_target` as described https://stackoverflow.com/questions/9190098/for-cmakes-install-command-what-can-the-component-argument-do

hebasto:
Updated cd9389f1b5074ecfd55218d530ae91cb6f2f228c -> fa130db398ef12752e976468eb12ad36dfd49b99 ([pr74.01](https://github.com/hebasto/libmultiprocess/commits/pr74.01) -> [pr74.02](https://github.com/hebasto/libmultiprocess/commits/pr74.02), [diff](https://github.com/hebasto/libmultiprocess/compare/pr74.01..pr74.02)).

@ryanofsky 

Thank you for your suggestions!

> Ideally in addition to current `make install` target there would be new `make install-lib` and `make install-bin` targets that automatically build and install the minimal dependencies.

Implemented.

The only downside of the current implementation is that new `install-lib` and `install-bin` targets do not export the underlying targets.

ryanofsky (approved):
Code review ACK fa130db398ef12752e976468eb12ad36dfd49b99. This does exactly what I'd expect now, so will merge it. But you do bring up an important point about no longer exporting cmake targets.

> The only downside of the current implementation is that new `install-lib` and `install-bin` targets do not export the underlying targets.

Can confirm neither of the new install commands create `$PREFIX/cmake/Multiprocess/Multiprocess.cmake` and `$PREFIX/lib/cmake/Multiprocess/Multiprocess-debug.cmake` files.

Probably should update `install()` `EXPORT Multiprocess` arguments to `EXPORT libmultiprocess-bin` and `EXPORT libmultiprocess-lib` so the two installs can be independent. And maybe there is a way to get the new `install-bin` and `install-lib` commands to actually export these targets.

But I am now thinking maybe a better long-term solution is to split up the cmake project into two projects. It could also clarify the source organization to split the parts of the code that are needed for code generation from the parts that are used at runtime.

PR #80, opened 2022-09-19
doc: Fix mpgen usage string

ryanofsky:
Also add comment to describe argument meanings.

Problem with string was reported by hebasto in https://github.com/chaincodelabs/libmultiprocess/pull/9#discussion_r973146130


hebasto (commented):
Post-merge ACK fcad5fbe5359f887e18624972694487a505e35d9.

PR #82, opened 2022-09-29
build: Fix missing Cap'n Proto include directories

ryanofsky:
Two bugs configuring Cap'n Proto include directories were reported by sjors in https://github.com/chaincodelabs/libmultiprocess/issues/75

They appear to have gone undetected because they were masked in other contexts by `CMAKE_INCLUDE_PATH` and `NIX_CFLAGS_COMPILE` environment variable settings (https://github.com/chaincodelabs/libmultiprocess/issues/75#issuecomment-1262334271)

The first bug is missing `${CAPNP_INCLUDE_DIRECTORY}` target_include_directories in the `util` target. This resulted in a compile error `include/mp/util.h:8:10: fatal error: 'capnp/schema.h' file not found`

The second bug was misspelling `CAPNP_INCLUDE_DIRS` as `capnp_INCLUDE_DIRS` in the configure check for `kj/filesystem.h`, which caused `HAVE_KJ_FILESYSTEM` to be unset and could result in runtime errors like:

```bash
capnp compile: --import-path=/prefix/include: no such directory
Try 'capnp compile --help' for more information.
terminate called after throwing an instance of 'std::runtime_error'
  what():  Invoking capnp failed
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,7 +52,8 @@ capnp_generate_cpp(MP_PROXY_SRCS MP_PROXY_HDRS include/mp/proxy.capnp)
 add_library(util OBJECT src/mp/util.cpp)
 target_include_directories(util PRIVATE
   $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/include>
-  $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include>)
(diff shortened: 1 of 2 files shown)


Sjors:
tACK c75362392f6b94134cca0089a5ef1c06796b833d

ryanofsky:
Thanks for reporting this and testing the patch

PR #96, opened 2024-03-16
cmake: Introduce packages

hebasto:
This PR adds configurations for the `Libmultiprocess` and `LibmultiprocessGen` packages.

The Bitcoin Core branch, with uses this PR branch, is available [here](https://github.com/hebasto/bitcoin/tree/240329-cmake-CI-mpgen).

As a suggestion for a follow-up, it seems worth disabling `mpgen` target for cross-compiling (see https://github.com/bitcoin/bitcoin/pull/29665#issuecomment-2002027957 and CMake [docs](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html#running-executables-built-in-the-project)).

Based on https://github.com/chaincodelabs/libmultiprocess/pull/95.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,6 +77,7 @@ add_library(multiprocess STATIC
   ${MP_PUBLIC_HEADERS}
   src/mp/proxy.cpp
   $<TARGET_OBJECTS:util>)
+add_library(Libmultiprocess::multiprocess ALIAS multiprocess)
 target_include_directories(multiprocess PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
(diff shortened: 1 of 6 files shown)


ryanofsky:
Looks nice! Will review this. I remembered you also posted a proof of concept branch defining packages last year in https://github.com/chaincodelabs/libmultiprocess/pull/79#issuecomment-1407430661 that I guess this supercedes?

hebasto:
> I remembered you also posted a proof of concept branch defining packages last year in [#79 (comment)](https://github.com/chaincodelabs/libmultiprocess/pull/79#issuecomment-1407430661) that I guess this supercedes?

Yes, it does.

ryanofsky (approved):
Code review ACK 4e70ad47191f85188da869400690c22f6f81e1e0. This should make it easier to depend on libmultprocess in cmake projects, using a find_package call and not having to include files directly.

I think I would like to rename a few things in the PR, to make new package names match existing component names (bin and lib). I also don't like the redundant "Libmultiprocess/Libmultiprocess" in some installed filenames, and the way one package name is a prefix of the other, since they are side-by-side dependencies. I can post a followup PR with some renames though.

PR #98, opened 2024-04-01
cmake: Combine installed packages

ryanofsky:
This change combines previous installed:

- `cmake/LibmultiprocessLibConfig.cmake`
- `cmake/LibmultiprocessBinConfig.cmake`

files into a single:

- `cmake/Libmultiprocess/LibmultiprocessConfig.cmake`

file, so it can be imported with `find_package(Libmultiprocess)`.

The previous locations which were set in https://github.com/chaincodelabs/libmultiprocess/pull/97 were not compatible with `find_package` search behavior by default.

The change also adds some documentation about using the new package to doc/install.md.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,9 @@ project("Libmultiprocess" CXX)
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED YES)
 
+find_package(CapnProto REQUIRED)
+find_package(Threads REQUIRED)
+
 option(Libmultiprocess_ENABLE_CLANG_TIDY "Run clang-tidy with the compiler." OFF)
 if(Libmultiprocess_ENABLE_CLANG_TIDY)
   find_program(CLANG_TIDY_EXECUTABLE NAMES clang-tidy)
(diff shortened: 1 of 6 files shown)


hebasto:
I did not test this PR yet, but have a question regarding the combined package configuration. 

In the Bitcoin Core depends, how does it suppose to work considering that the mpgen tool is installed in the `native` subdirectory of the depends prefix?

ryanofsky:
I'm not sure exactly, but I'd imagine you'd just find the package twice with different components instead of finding two packages separately. Even before this PR you would need to point cmake at the native directory to find the code generation package, so I don't think that aspect is changing.

ryanofsky:
Or if the issue is that cmake doesn't allow importing two packages with the same name, maybe a good solution would be to add an option here controlling the installed package name.

If there multiple libmultiprocess builds are installed (native and cross-compiled) it seems good if they are all allowed to be full installs, and downstream cmake projects can choose which components of each install to use. Even though depends system only does half-builds for efficiency (executable only or library only), this doesn't seem like something that should be required for other builds.

hebasto:
> Or if the issue is that cmake doesn't allow importing two packages with the same name, maybe a good solution would be to add an option here controlling the installed package name.

Once package found, the `${PACKAGE}_DIR` variable is cached.

Therefore, to work with Bitcoin Core's depends, this code should be used:
```cmake
find_package(Libmultiprocess COMPONENTS Lib)
unset(Libmultiprocess_DIR CACHE)
find_package(Libmultiprocess COMPONENTS Bin
  NO_CMAKE_FIND_ROOT_PATH
  PATHS ${CMAKE_FIND_ROOT_PATH}/native
)
```

But this approach undermines the user's ability to specify non-standard package component locations because the `Libmultiprocess_DIR` variable is overloaded.

ryanofsky:
re: https://github.com/chaincodelabs/libmultiprocess/pull/98#issuecomment-2241236657

@hebasto can you use the `find_package` `NAMES` option to give the packages different names? From the [find_package documentation](https://cmake.org/cmake/help/latest/command/find_package.html#full-signature):

> A cache entry called `<PackageName>_DIR` is created to hold the directory containing the file. By default, the command searches for a package with the name `<PackageName>`. If the `NAMES` option is given, the names following it are used instead of `<PackageName>`.

So modifying your example above, you could write:

```cmake
find_package(Libmultiprocess COMPONENTS Lib)
find_package(LibmultiprocessNative COMPONENTS Bin
  NAMES Libmultiprocess
  NO_CMAKE_FIND_ROOT_PATH
  PATHS ${CMAKE_FIND_ROOT_PATH}/native
)
```

This should result in separate `Libmultiprocess_DIR` and `LibmultiprocessNative_DIR` variables

hebasto:
@ryanofsky 

> can you use the `find_package` `NAMES` option to give the packages different names? From the [find_package documentation](https://cmake.org/cmake/help/latest/command/find_package.html#full-signature):

Thank you! The suggested approach looks nice to me.

It only requires a bugfix suggested in https://github.com/chaincodelabs/libmultiprocess/pull/103.

PR #103, opened 2024-07-23
cmake: Fix package configuration file

hebasto:
When loading a package configuration file, `find_package` defines variables to provide information about the call arguments. In particular, `CMAKE_FIND_PACKAGE_NAME` represents the package name.

This PR uses this variable instead of a hardcoded name, making the use case suggested in https://github.com/chaincodelabs/libmultiprocess/pull/98#issuecomment-2243751206 workable:
```cmake
find_package(Libmultiprocess COMPONENTS Lib)
find_package(LibmultiprocessNative COMPONENTS Bin
  NAMES Libmultiprocess
)
```

### DIFF
--- a/cmake/Config.cmake.in
+++ b/cmake/Config.cmake.in
@@ -7,24 +7,24 @@
 set(_Libmultiprocess_supported_components Bin Lib)
 
 # If no components specified, include all components.
-list(LENGTH Libmultiprocess_FIND_COMPONENTS Libmultiprocess_FIND_COMPONENTS_len)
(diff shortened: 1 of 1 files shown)


hebasto:
@ryanofsky 

FWIW, the recent [push](https://github.com/hebasto/bitcoin/tree/pr29790-0723.2.mp) to https://github.com/bitcoin/bitcoin/pull/29790 uses this PR implementation.

ryanofsky (approved):
Code review ACK c373a94d5b4b3f9c3ee582d15db61d17525bb9bf

Thanks for the fix!

PR #109, opened 2024-08-21
example: Add missing thread.join() call so example can exit cleanly

ryanofsky:
Problem was pointed out by TheCharlatan in https://github.com/chaincodelabs/libmultiprocess/pull/107#issuecomment-2301338227


TheCharlatan (approved):
ACK 70b2d8794f3b78c2028dd6e8d00444b178e76893

I tried adding `thread.join()` too before, but forgot to apply my prior patch and it never exited, because it still had a client. Thank you!

PR #111, opened 2024-09-09
doc: Add internal design section

TheCharlatan:
I am not sure if the descriptions here are correct, or even all that useful, but I think some of the finer points are useful to document to serve as an anchor for future developers reading the code. Maybe I also missed some documentation in the headers and the other docs that already describe these things better.

~What could potentially also be useful are some diagrams to show the wrapping hierarchy as done below (again, not sure if actually correct, and given the polymorphism and templating going on it is also questionable if such a hierarchy can accurately be depicted as a diagram).~

Given the points made by ryanofsky below, I don't think anymore than an object diagram will be particularly helpful.

 A sequence diagram that follows a call from its c++ callsite to the client invocation to the server receiving and processing it, the client resolving it and finally returning a c++ type again would probably be more useful.


ryanofsky:
Thanks, this all seems helpful and correct reading it at a first pass. I think I would just tweak a few things and probably add a few additional details. It think it's especailly helpful to note that libmultiprocess is just acting as a wrapper to make it easier to write and call cap'nproto interfaces from c++. But it is completely possible to call or implement the same interfaces without using libmultiprocess (especially from other languages like rust python and javascript).

One thing that is a little off in 86415b456bb3307afae406c2b1b5cbc5d0c143b3 is that it makes it sound like ProxyClient and ProxyServer objects wrap ends of socket, when it would be more accurate just to say that they communicate over sockets. Specifically when there is a socket connection there is not just one ProxyClient and one ProxyServer object sitting at either end of the socket, but one ProxyClient and one ProxyServer for each instantiated interface, and probably dozens or hundreds of ProxyClient and ProxyService objects all using the same socket.

If you want to mention the class that does sit at either end of a socket, that would just be the `mp::Connection` class defined in `mp/proxy-io.h`

Similarly, in diagrams above it could give wrong impression to show AsyncioContext containing AsyncIoStream, because the same context can be used to communicate across many streams. For example, if there is a bitcoin-node process, it may spawn a bitcoin-wallet and communicate across one stream with but also accept bitcoin-mine connections and use other streams to connect to them. All the streams will use the same EventLoop and AsyncioContext.

It is true in both of these 1:N relationships though, that not all of the N are equal. Even though there will be many ProxyClient/ProxyServer clients using the same socket, and many AsyncIoStreams users the same AsyncioContext and EventLoop, users are not undifferentiated and identical. While the EventLoop uses many AsyncIoStreams it does have one special `wait_stream` that is used to receive requests posted to the event loop. Similarly, when a connection is first established, there is a initial ProxyClient and ProxyServer pair wrapping an Init interface, which is special because it is the first interface, and it what is used to create all other interfaces and ProxyClient and ProxyServer objects wrapping the interfaces that are created.

TheCharlatan:
Thank you for the feedback, I tweaked the docs a bit to make the relationship between connections, proxy objects and capnp objects a bit clearer. One thing I have been wondering about is the naming and usage of `Field`. I am guessing this comes from capnp proto naming scheme, but is that always entirely appropriate here?

doc/design.md
```diff
@@ -11,6 +11,16 @@ If the wrapped C++ object inherits from an abstract base class declaring virtual
...
+
+When a generated method on the `ProxyClient` is called, it calls `clientInvoke` with the capnp-translated types. `clientInvoke` creates a self-executing promise (`kj::TaskSet`) that drives the execution of the request and gives ownership of it to the `EventLoop`.
```
ryanofsky, line 20: In commit "doc: Add internal design section" (86415b456bb3307afae406c2b1b5cbc5d0c143b3)

Might be worth adding that after `clientInvoke` sends the request, it blocks waiting until a response is received, or until there is a call from the server that needs to run on the same client thread, using a `Waiter` object.

doc/design.md
```diff
@@ -11,6 +11,16 @@ If the wrapped C++ object inherits from an abstract base class declaring virtual
...
+
+On the server side, the `capnp::RpcSystem` receives the capnp request and invokes the corresponding c++ method through the corresponding `ProxyServer` and the heavily templated `serverInvoke` triggering a `ServerCall`. Its result is then converted back by invoking `ServerRet`. The two are connected through `ServerField`. The main method driving execution of a request is `PassField`, which is invoked through `ServerField`. Instantiated interfaces, or capabilities in capnp speak, are tracked and owned by the server's `capnp::RpcSystem`.
```
ryanofsky, line 22: In commit "doc: Add internal design section" (86415b456bb3307afae406c2b1b5cbc5d0c143b3)

Could expand "Its result is then converted back by invoking `ServerRet`." Maybe say "Return values from the c++ methods are copied into Cap'nProto responses by `ServerRet` and exceptions are caught and copied by `ServerExcept`".

Also from this description it sounds like ServerField should probably be renamed to ServerArg or ServerParam. But none of these names are very good, they should probably be named more like ServerMethodArgumentHandler ServerMethodReturnValueHandler ServerMethodExceptionHandler

ryanofsky (approved):
ACK 86415b456bb3307afae406c2b1b5cbc5d0c143b3

I had some ideas reading this so left a bunch of suggestions, but I'm happy to merge it as, just let me know. I think everything you wrote is accurate and this is definitely an improvement

TheCharlatan:
Thanks for the comments, will pour over this some more and hopefully come up with more improvements.

doc/design.md
TheCharlatan: These suggested names sound good to me, More verbose, but they convey the meaning more clearly.

TheCharlatan:
Took your improvements @ryanofsky, and added some minor corrections.

ryanofsky:
ACK 1fa2ca7cd804d097d02eb7ad0927832e40c3f4a5. Looks good, and nice job incorporating the extra information, especially about the waiter object. I'll go ahead and merge this now but feel free to make more changes or suggest improvements.

PR #112, opened 2024-09-23
cmake: Introduce `Libmultiprocess_INSTALL_PKGCONFIG_FILE` build option

hebasto:
This PR makes the installation of the `pkgconfig/libmultiprocess.pc` file optional.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,8 @@ if(Libmultiprocess_ENABLE_CLANG_TIDY)
   set(CMAKE_CXX_CLANG_TIDY "${CLANG_TIDY_EXECUTABLE}")
 endif()
 
+option(Libmultiprocess_INSTALL_PKGCONFIG_FILE "Install pkg-config module to build against libmultiprocess library." ON)
+
 include("cmake/capnp_compat.cmake")
 include("cmake/pthread_checks.cmake")
 include(GNUInstallDirs)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -88,9 +90,11 @@ install(FILES "include/mpgen.mk"
   DESTINATION ${CMAKE_INSTALL_INCLUDEDIR} COMPONENT bin)
 
 # pkg-config module to build against libmultiprocess library, for downstream autoconf projects
-configure_file(pkgconfig/libmultiprocess.pc.in "${CMAKE_CURRENT_BINARY_DIR}/pkgconfig/libmultiprocess.pc" @ONLY)
-install(FILES "${CMAKE_CURRENT_BINARY_DIR}/pkgconfig/libmultiprocess.pc"
-  DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig COMPONENT lib)
+if(Libmultiprocess_INSTALL_PKGCONFIG_FILE)
+  configure_file(pkgconfig/libmultiprocess.pc.in "${CMAKE_CURRENT_BINARY_DIR}/pkgconfig/libmultiprocess.pc" @ONLY)
+  install(FILES "${CMAKE_CURRENT_BINARY_DIR}/pkgconfig/libmultiprocess.pc"
```
hebasto, line 95: Also, the `pkgconfig` subdirectory seems redundant here, no?

hebasto:
> Is there a particular reason to not install it by default? Note that `pkg-config` is just a generic tool, and it's usage is not specific to Autotools. Meson and other build systems also use it.

The default value of the new option has been changed to `ON`.

ryanofsky:
What's the motivation for this change? It seems like an unusual thing for a package to have a configuration option to not fully install itself. And if installing a .pc file is optional, should installing .cmake package files be optional also? I think I'd prefer the cmake code here to be as vanilla as possible and not have a non-standard option like this, unless there is a real reason why it's useful.

Also it seems would seem better to just use cmake components to control what is installed instead of using both cmake components and cmake options to control what is installed. Maybe installed metadata files could be moved into different components to provide finer-grained control over what to install if this is necessary.

TheCharlatan:
> Alternatively, it can be deleted during postprocessing, as we do with other packages.

I think I would prefer that.

PR #119, opened 2024-10-15
cmake: avoid libatomic not found error on debian

### LINKED ISSUE #68: cmake .. Returns error libatomic not found
Hi I cannot install libmultiprocess

I am returned the error:

CMake Error at /usr/lib/x86_64-linux-gnu/cmake/CapnProto/CapnProtoConfig.cmake:71 (message):
  libatomic not found
Call Stack (most recent call first):
  CMakeLists.txt:9 (find_package)

I have installed capnproto from master tree c++ here:

https://github.com/capnproto/capnproto/tree/master/c%2B%2B

using autoreconf -i && ./configure && make -j6 check && sudo make install

Running CMake .. on build directory still did not work.

I double checked libatomic-ops and installed from master tree at

https://github.com/ivmai/debian-libatomic-ops

I updated libstdc++ with sudo apt-get install libstdc++ and apt -fix-broken install for safe measure

I again ran for capnproto on master/capnproto/c++

autoreconf -i && ./configure && make -j6 check && sudo make install

to see if the reconfiguration would link libatomic

Again testing CMake .. on libmultiprocess/build I am returned:

CMake Error at /usr/lib/x86_64-linux-gnu/cmake/CapnProto/CapnProtoConfig.cmake:71 (message):
  libatomic not found
Call Stack (most recent call first):
  CMakeLists.txt:9 (find_package)

Why is CMake not finding libatomic and connecting it for capnproto?

Under capnproto issues another user had a similar problem with Raspbian. In my case I am running Kali with Debian Bullseye repositories overtop. My system is well built with GNU libraries as well, and runs as a full fleshed out Debian distro inside a Kali. I am on linux x86_64.

https://github.com/capnproto/capnproto/issues/1448

In his case capnproto did not want to build at all, but it passes with all tests succeeded in my case, however it is not liking to configure with libatomic-ops, even though it is already installed.

Because of this the CMake for libmultiprocess cannot be done.

Just looking to set libmultiprocess in place for installing bitcoin and master tree using --enable-libmultiprocess for added compatibility in advance of future needs. I don't want to have to recomp[ile with enable multiprocess later on, it could be frustrating. I am aware that this is entirely to assist in the C++ libraries for compiling purposes only, and doesn't really have any major affect on bitcoin-core installation, but I would prefer using libmultiprocess with bitcoin for any compatibility issues post installation or with accessory uses.

I would appreciate the time looked into this.

Satori Hoshi

ryanofsky:
Add workaround to prevent "libatomic not found" when building on debian against debian capnproto package which installs a buggy CapnProtoConfig.cmake file. The bug was introduced as part of debian's attempt to fix a different issue: https://github.com/capnproto/capnproto/issues/1448 and their fix works for autoconf projects but not cmake projects like libmultiprocess.

Fixes #68

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,8 @@ project("Libmultiprocess" CXX)
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED YES)
 
+include("cmake/compat_find.cmake")
+
 find_package(CapnProto REQUIRED)
 find_package(Threads REQUIRED)
 
@@ -20,7 +22,7 @@ if(Libmultiprocess_ENABLE_CLANG_TIDY)
   set(CMAKE_CXX_CLANG_TIDY "${CLANG_TIDY_EXECUTABLE}")
 endif()
 
-include("cmake/capnp_compat.cmake")
+include("cmake/compat_config.cmake")
 include("cmake/pthread_checks.cmake")
 include(GNUInstallDirs)
 
--- a/cmake/compat_config.cmake
+++ b/cmake/compat_config.cmake
@@ -2,6 +2,9 @@
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
+# compat_config.cmake -- compatibility workarounds meant to be included after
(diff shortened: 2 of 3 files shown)


hebasto (commented):
Post-merge ACK 296c380a4bba6887f267fd84600172d614d01a74.

PR #120, opened 2024-11-12
proxy-types.h: add static_assert to detect int/enum size mismatch

ryanofsky:
Add static_assert to detect when an int field is too small to hold an enum value

This catches the bug TheCharlatan pointed out in https://github.com/bitcoin/bitcoin/pull/29409#discussion_r1834362068

### DIFF
--- a/include/mp/proxy-types.h
+++ b/include/mp/proxy-types.h
@@ -963,6 +963,9 @@ LocalType BuildPrimitive(InvokeContext& invoke_context,
     TypeList<LocalType>,
     typename std::enable_if<std::is_enum<Value>::value>::type* enable = nullptr)
 {
+    using E = std::make_unsigned_t<std::underlying_type_t<Value>>;
+    using T = std::make_unsigned_t<LocalType>;
+    static_assert(std::numeric_limits<T>::max() >= std::numeric_limits<E>::max(), "mismatched integral/enum types");
     return static_cast<LocalType>(value);
 }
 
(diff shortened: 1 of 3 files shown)


TheCharlatan (approved):
lgtm ACK bbc80abc20300dca7481e01e20709ec0d1a6bd5b

ryanofsky:
> Maybe add a test to illustrate the fix?

Good idea. There was no coverage for enum/int conversions so I pushed a new commit 110349f6268dc95179c0a36b6fc3b00c6e3f1751 to add some.

I don't think there's a good way to add test coverage for the new `static_assert` since it just triggers compile errors if enum and int types aren't compatible. But it is possible to trigger the static assert manually by changing:

```diff
--- a/test/mp/test/foo.capnp
+++ b/test/mp/test/foo.capnp
@@ -27,7 +27,7 @@ interface FooInterface $Proxy.wrap("mp::test::FooImplementation") {
     passEmpty @12 (arg :FooEmpty) -> (result :FooEmpty);
     passMessage @13 (arg :FooMessage) -> (result :FooMessage);
     passMutable @14 (arg :FooMutable) -> (arg :FooMutable);
-    passEnum @15 (arg :Int32) -> (result :Int32);
+    passEnum @15 (arg :Int16) -> (result :Int32);
 }
 
 interface FooCallback $Proxy.wrap("mp::test::FooCallback") {
```

Then there should be a compile error if you run `make check`:

```c++
include/mp/proxy-types.h:968:19: error: static assertion failed due to requirement 'std::numeric_limits<unsigned short>::max() >= std::numeric_limits<unsigned int>::max()': mismatched integral/enum types
  968 |     static_assert(std::numeric_limits<T>::max() >= std::numeric_limits<E>::max(), "mismatched integral/enum types");
[...]
include/mp/proxy-types.h:968:49: note: expression evaluates to '65535 >= 4294967295'
  968 |     static_assert(std::numeric_limits<T>::max() >= std::numeric_limits<E>::max(), "mismatched integral/enum types");

```

PR #126, opened 2025-01-15
ProxyClientBase: avoid static_cast to partially destructed object

### LINKED ISSUE #125: Bitcoin Core UBSan failure for interfaces::Echo
When running `ipc_tests.cpp` with the UBSan fix from #121, it still fails:

```
export UBSAN_OPTIONS="suppressions=$(pwd)/test/sanitizer_suppressions/ubsan:print_stacktrace=1:halt_on_error=1:report_error_type=1"
cmake -B build -DWITH_MULTIPROCESS=ON -DSANITIZERS=undefined
cmake --build build

build/src/test/test_bitcoin --run_test=ipc_tests --log_level=all | grep -v disabled
...
test/ipc_test.cpp:143: info: check remote_echo->echo("echo test") == "echo test" has passed
/usr/local/include/mp/proxy.h:95:45: runtime error: downcast of address 0x600001f28700 which does not point to an object of type 'ProxyClient<Interface>' (aka 'ProxyClient<ipc::capnp::messages::Echo>')
0x600001f28700: note: object is of type 'mp::ProxyClientBase<ipc::capnp::messages::Echo, interfaces::Echo>'
...
SUMMARY: UndefinedBehaviorSanitizer: dynamic-type-mismatch /usr/local/include/mp/proxy.h:95:45 in 
```

ryanofsky:
This is a bugfix that should fix the UBSan failure reported https://github.com/chaincodelabs/libmultiprocess/issues/125

In practice there is no real bug here, just like there was no real bug fixed in the recent "ProxyClientBase: avoid static_cast to partially constructed object" change in https://github.com/chaincodelabs/libmultiprocess/pull/121. This is because in practice, ProxyClient subclasses only inherit ProxyClientBase state and do not define any state of their own, so there is nothing bad that could actually happen from static_cast-ing a ProxyClientBase pointer to a ProxyClient pointer inside the ProxyClientBase constructor or destructor.

However, using static_cast this way technically triggers undefined behavior, and that causes UBSan to fail, so this commit moves ProxyClientBase cleanup out of the ProxyClientBase destructor, into ProxyClient subclass destructors to prevent this.

An alternate fix could maybe avoid the need to do this by making ProxyClient::construct() and ProxyClient::destroy() methods into static methods taking a ProxyClientBase& self argument instead of instance methods using *this. This would avoid the need to use static_cast at all. But it would also require changes to the ProxyClient class code generator so unclear if it would be a simpler fix.

Fixes #125

### DIFF
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -63,16 +63,16 @@ struct ProxyClient<Thread> : public ProxyClientBase<Thread, ::capnp::Void>
     ProxyClient(const ProxyClient&) = delete;
     ~ProxyClient();
 
-    void setCleanup(std::function<void()> cleanup);
+    void setCleanup(std::function<void()> fn);
 
     //! Cleanup function to run when the connection is closed. If the Connection
     //! gets destroyed before this ProxyClient<Thread> object, this cleanup
     //! callback lets it destroy this object and remove its entry in the
     //! thread's request_threads or callback_threads map (after resetting
-    //! m_cleanup so the destructor does not try to access it). But if this
+    //! m_cleanup_it so the destructor does not try to access it). But if this
     //! object gets destroyed before the Connection, there's no need to run the
     //! cleanup function and the destructor will unregister it.
-    std::optional<CleanupIt> m_cleanup;
+    std::optional<CleanupIt> m_cleanup_it;
 };
 
 template <>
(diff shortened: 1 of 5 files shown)


ryanofsky:
Closing this in favor of alternative fix in #127 which should be simpler and less fragile.

PR #130, opened 2025-01-23
refactor: Add CleanupRun function to dedup clean list code

ryanofsky:
Add CleanupRun function to dedup clean list code. Also rename "cleanup" variables to distinguish between cleanup iterators and cleanup callback functions.

These changes were originally part of https://github.com/chaincodelabs/libmultiprocess/pull/126 which was closed for other reasons, but I think they are still helpful and make code easier to navigate.

Since this renames a public struct member, it will require a change to bitcoin core when the depends version is bumped:

```c++
diff --git a/src/ipc/capnp/protocol.cpp b/src/ipc/capnp/protocol.cpp
index 4b67a5bd1e36..691bdf5f9242 100644
--- a/src/ipc/capnp/protocol.cpp
+++ b/src/ipc/capnp/protocol.cpp
@@ -73,7 +73,7 @@ public:
     }
     void addCleanup(std::type_index type, void* iface, std::function<void()> cleanup) override
     {
-        mp::ProxyTypeRegister::types().at(type)(iface).cleanup.emplace_back(std::move(cleanup));
+        mp::ProxyTypeRegister::types().at(type)(iface).cleanup_fns.emplace_back(std::move(cleanup));
     }
     Context& context() override { return m_context; }
     void startLoop(const char* exe_name)
```

### DIFF
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -63,16 +63,16 @@ struct ProxyClient<Thread> : public ProxyClientBase<Thread, ::capnp::Void>
     ProxyClient(const ProxyClient&) = delete;
     ~ProxyClient();
 
-    void setCleanup(std::function<void()> cleanup);
+    void setCleanup(std::function<void()> fn);
 
     //! Cleanup function to run when the connection is closed. If the Connection
     //! gets destroyed before this ProxyClient<Thread> object, this cleanup
     //! callback lets it destroy this object and remove its entry in the
     //! thread's request_threads or callback_threads map (after resetting
-    //! m_cleanup so the destructor does not try to access it). But if this
+    //! m_cleanup_it so the destructor does not try to access it). But if this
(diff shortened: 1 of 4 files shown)


ryanofsky:
Updated 7d08d83b7a7895a8ea070f16426a8527509f237d -> 700085f9a8e5f8bf1b200c4c09c8314b186c150a ([`pr/cleanupcleanup.1`](https://github.com/ryanofsky/libmultiprocess/commits/pr/cleanupcleanup.1) -> [`pr/cleanupcleanup.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/cleanupcleanup.2), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/cleanupcleanup.1..pr/cleanupcleanup.2)) tweaking to delete the cleanup functions after running them, which might be useful in case they are holding on to any state.

PR #136, opened 2025-01-27
cmake: Support being included with add_subdirectory

ryanofsky:
This implements changes needed to make the cmake project work well when it is not the top level project and has been included with `add_subdirectory`.

- Avoids defining "tests" and "check" targets when project is not at the top level to avoid clashes. Add new "mptests" and "mpcheck" alternatives instead.
- Renames "example" target to "mpexamples" 
- Renames "util" target to "mputil"
- Makes various internal changes needed to resolve other issues. See commit messages for details.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,20 @@ include("cmake/compat_config.cmake")
 include("cmake/pthread_checks.cmake")
 include(GNUInstallDirs)
 
+# Set convenience variables for subdirectories.
 set(MP_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/include")
+if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+  set(MP_STANDALONE TRUE)
+else()
+  # Set MP_INCLUDE_DIR for parent directories too, so target_capnp_sources calls
+  # in parent directories can use it and not need to specify include directories
+  # manually or see capnproto error "error: Import failed: /mp/proxy.capnp"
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -26,17 +26,32 @@ include("cmake/compat_config.cmake")
...
 include(GNUInstallDirs)
 
+# Set convenience variables for subdirectories.
+set(MP_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/include")
+if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+  set(MP_STANDALONE TRUE)
+else()
+  # Set MP_INCLUDE_DIR for parent directories too, so target_capnp_sources calls
+  # in parent directories can use it and not need to specify include directories
+  # manually or see capnproto error "error: Import failed: /mp/proxy.capnp"
```
hebasto, line 36: This comment does not clarify why the propagating a variable to the parent namespace is better than specifying "include directories manually".

I think that the latter is better. I do think that `add_subdirectory` calls should not introduce any side effects in the call sites.

ryanofsky: I do think specifying include directories is better in general. But I don't believe libmultiprocess exporting its own include path to callers and then requiring callers to pass that include path back to it is necessarily better. For example when you use a toolchain's cc command, the cc command will allow you to override the toolchain include paths and require you to specify your own include paths, but will not force you to specify the toolchain include paths.

So I think the change in 1103f86cd96238903ae1c01c8af4bb672079d5c0 is reasonable and good simplification, but I'm also fine with reverting it or just updating the comment here. You can let me know what you prefer and of course a PR would be welcome.

PR #140, opened 2025-01-30
build: don't clobber user/superproject c++ version

theuni:
This allows a vendored libmultiprocess to inherit the c++ version from the main project

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,8 +5,10 @@
 cmake_minimum_required(VERSION 3.8)
 
 project("Libmultiprocess" CXX)
-set(CMAKE_CXX_STANDARD 17)
-set(CMAKE_CXX_STANDARD_REQUIRED YES)
+if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+  set(CMAKE_CXX_STANDARD 17)
+  set(CMAKE_CXX_STANDARD_REQUIRED YES)
+endif()
 
 include("cmake/compat_find.cmake")
 


ryanofsky (approved):
Sorry I missed this in #136. Code review ACK bf07b262ba43a000f7605d050de2b6e39fd350e7

TheCharlatan:
lgtm

CMakeLists.txt
```diff
@@ -5,7 +5,9 @@
 cmake_minimum_required(VERSION 3.8)
 
 project("Libmultiprocess" CXX)
-set(CMAKE_CXX_STANDARD 17)
+if(NOT DEFINED CMAKE_CXX_STANDARD)
```
ryanofsky, line 8: In commit "build: don't clobber user/superproject c++ version" (bf07b262ba43a000f7605d050de2b6e39fd350e7)

I'm wondering if you think it might be better to do

```
if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
  set(CMAKE_CXX_STANDARD 17
endif()
```

to handle the CMAKE_CXX_STANDARD and CMAKE_CXX_STANDARD_REQUIRED variables together instead of separately, to avoid clobbering CMAKE_CXX_STANDARD_REQUIRED, and to be more consistent with parent project detection code below.

ryanofsky (approved):
I'm wondering if fix suggested below might make more sense. Happy to use whichever fix you prefer.

CMakeLists.txt
theuni: Done.

hebasto: > Done.

Did you push the correct branch?

theuni: Heh, whoops. Thanks. Fixed.

hebasto (approved):
ACK df2153551e1bb05f9796fae75817051d52b86622.

ryanofsky (approved):
Code review ACK df2153551e1bb05f9796fae75817051d52b86622

Issue #138, opened 2025-01-28
Compiler warnings from capnproto when building via add_subdirectory


ryanofsky:
_Originally posted by @Sjors in https://github.com/bitcoin/bitcoin/pull/31741#pullrequestreview-2577588200_

On (M4) macOS 15.2 I get a flood of warnings, which means I can't use ` -DWERROR=ON`:

```
In file included from /opt/homebrew/include/capnp/common.h:31:
/opt/homebrew/include/kj/windows-sanity.h:39:6: warning: '_WIN32' is not defined, evaluates to 0 [-Wundef]
   39 | #if !_WIN32 && !__CYGWIN__
      |      ^
/opt/homebrew/include/kj/windows-sanity.h:39:17: warning: '__CYGWIN__' is not defined, evaluates to 0 [-Wundef]
   39 | #if !_WIN32 && !__CYGWIN__
      |     

/opt/homebrew/include/capnp/common.h:33:5: warning: 'CAPNP_DEBUG_TYPES' is not defined, evaluates to 0 [-Wundef]

/opt/homebrew/include/capnp/raw-schema.h:26:5: warning: '_MSC_VER' is not defined, evaluates to 0 [-Wundef]

/opt/homebrew/include/kj/async-prelude.h:34:6: warning: 'KJ_NO_EXCEPTIONS' is not defined, evaluates to 0 [-Wundef]
```

ryanofsky:
Using https://github.com/bitcoin/bitcoin/pull/31741 (https://github.com/ryanofsky/bitcoin/commits/pr/subtree.3) I don't see any warnings on my sstem, so help to provide more steps to reproduce.

If you can try building with `make -C build VERBOSE=1 ...` that would also help show the compiler comment lines so we can see if there's something wrong with `-I` `-isystem` `-Wno-system-headers` options that are being passed might be causing this issue.

Sjors:
On 93ba80b3d78fa1c4e7f614df6517b9cb492d05d7 (build directory deleted)

```
cmake -B build -DWERROR=OFF -DBUILD_MULTIPROCESS=ON
```

<details>
<summary>log</summary>

```
-- The CXX compiler identification is AppleClang 16.0.0.16000026
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
[...]

PR #141, opened 2025-01-30
build: Fix for vendored subdir in Bitcoin Core when using depends

theuni:
This is a really wonky issue (it took me quite a while to figure out why it was complaining). It works around a quirk path issue that presents itself when using libmultiprocess as a subdir of Core and using a depends-built capnproto within the same source tree:

```
CMake Error in src/ipc/libmultiprocess/CMakeLists.txt:
  Target "multiprocess" INTERFACE_INCLUDE_DIRECTORIES property contains path:

    "/home/cory/dev/bitcoin/depends/x86_64-pc-linux-gnu/include"

  which is prefixed in the source directory.

```

It doesn't present in https://github.com/bitcoin/bitcoin/pull/31741 because in that case, the lib is built in depends rather than as part of the source tree.

This is part of my work to always build from the source tree rather than from depends. I can push up the other changes (the `EXTERNAL_MPGEN` option )needed for that, but I wanted to break this one out in case it needed further discussion/explanation. It's safe to go in as-is though, it should just be a no-op.

While I was at it, because the capnp headers spew annoying and harmless warnings when built with Core's flags turned on, I marked them as `SYSTEM`.

See the commit description for more details.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,9 +97,18 @@ target_include_directories(multiprocess PUBLIC
 target_include_directories(multiprocess SYSTEM PUBLIC
   $<BUILD_INTERFACE:${CAPNP_INCLUDE_DIRECTORY}>
 )
-target_include_directories(multiprocess SYSTEM PUBLIC
-  $<INSTALL_INTERFACE:${CAPNP_INCLUDE_DIRECTORY}>
-)
+
+# Hack to avoid exporting a path "which is prefixed in the source directory".
+# If we're in another project's subdir and capnp is installed somewhere inside
+# that project (as it is with depends in Bicoin Core) CMake refuses to allow it
+# in the install interface. Use the "EXCLUDE_FROM_ALL" property as a proxy for
+# detecting this condition, as in that case install won't happen anyway.
+get_directory_property(mp_skip_install EXCLUDE_FROM_ALL)
+if(NOT "${mp_skip_install}")
+  target_include_directories(multiprocess SYSTEM PUBLIC
+    $<INSTALL_INTERFACE:${CAPNP_INCLUDE_DIRECTORY}>
+  )
+endif()
 
 target_link_libraries(multiprocess PRIVATE CapnProto::capnp)
 target_link_libraries(multiprocess PRIVATE CapnProto::capnp-rpc)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,7 +92,15 @@ target_include_directories(multiprocess PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
-  ${CAPNP_INCLUDE_DIRECTORY})
+)
+
+target_include_directories(multiprocess SYSTEM PUBLIC
+  $<BUILD_INTERFACE:${CAPNP_INCLUDE_DIRECTORY}>
+)
+target_include_directories(multiprocess SYSTEM PUBLIC
+  $<INSTALL_INTERFACE:${CAPNP_INCLUDE_DIRECTORY}>
+)
+
 target_link_libraries(multiprocess PRIVATE CapnProto::capnp)
 target_link_libraries(multiprocess PRIVATE CapnProto::capnp-rpc)
 target_link_libraries(multiprocess PRIVATE CapnProto::kj)


hebasto:
> While I was at it, because the capnp headers spew annoying and harmless warnings when built with Core's flags turned on, I marked them as `SYSTEM`.

A fix for https://github.com/chaincodelabs/libmultiprocess/issues/138?

CMakeLists.txt
```diff
@@ -92,7 +92,15 @@ target_include_directories(multiprocess PUBLIC
...
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
-  ${CAPNP_INCLUDE_DIRECTORY})
+)
+
+target_include_directories(multiprocess SYSTEM PUBLIC
+  $<BUILD_INTERFACE:${CAPNP_INCLUDE_DIRECTORY}>
+)
+target_include_directories(multiprocess SYSTEM PUBLIC
+  $<INSTALL_INTERFACE:${CAPNP_INCLUDE_DIRECTORY}>
+)
```
hebasto: style nit: Why not a single call?

hebasto:
Hmm, indeed a hacky fix... I'll try to look for alternatives.

theuni:
> Hmm, indeed a hacky fix... I'll try to look for alternatives.

If you'd like to reproduce the error, you can try my branch here where I've been hacking stuff together: https://github.com/theuni/bitcoin/commits/pr31741-split-native/

It's pretty much the same as what's here. You can remove the `if()` to hit the bug.

CMakeLists.txt
```diff
@@ -97,9 +97,18 @@ target_include_directories(multiprocess PUBLIC
...
+# If we're in another project's subdir and capnp is installed somewhere inside
+# that project (as it is with depends in Bicoin Core) CMake refuses to allow it
+# in the install interface. Use the "EXCLUDE_FROM_ALL" property as a proxy for
+# detecting this condition, as in that case install won't happen anyway.
+get_directory_property(mp_skip_install EXCLUDE_FROM_ALL)
+if(NOT "${mp_skip_install}")
+  target_include_directories(multiprocess SYSTEM PUBLIC
+    $<INSTALL_INTERFACE:${CAPNP_INCLUDE_DIRECTORY}>
+  )
+endif()
```
ryanofsky, line 111: In commit "build: work around CMake quirk when building as a subdir of Bitcoin Core" (7ac1237ea32aad4a174aaf1958fe2fe0ac129662)

This fix should work as you explained it, since when this build is a subdirectory of a bitcoin core build the libmultiprocess.a will be internal library that should not be installed and doesn't need to propagate any include paths after installation.

But I think this fix could be cleaner if it included the first half of this diff https://github.com/chaincodelabs/libmultiprocess/issues/138#issuecomment-2624919124, that way *both* uses of CAPNP_INCLUDE_DIRECTORY here could be deleted, because the necessary include paths would already be brought in by the target_link_libraries call, so specifying it again in target_include_directories should be redundant.

It seemed like even that diff alone was enough to fix the problem with header warnings according to what we saw in Sjors CI job: https://github.com/bitcoin/bitcoin/pull/30975#issuecomment-2625235833. The header warnings are reported in #138 but they don't happen locally for me locally (possibly because I use a nix shell which adds a bit of include/link magic for specified buildInputs)

theuni: Thanks @ryanofsky, I'll play with that. If it wasn't clear from my description, I acknowledge that this is a weird/ugly fix and I don't like it, so I'm up for any nicer alternative that fixes the problem.

As to the warnings, that's another good example imo of why we want a unified in-tree build, so that everyone's seeing the same thing. I'll get the rest of the pre-requisites for that pushed up today. This is the ugly one, the other should be more straightforward (though I expect you'll have some suggestions to tidy up the impl).

theuni: > But I think this fix could be cleaner if it included the first half of this diff https://github.com/chaincodelabs/libmultiprocess/issues/138#issuecomment-2624919124, that way both uses of CAPNP_INCLUDE_DIRECTORY here could be deleted, because the necessary include paths would already be brought in by the target_link_libraries call, so specifying it again in target_include_directories should be redundant.

I don't really understand why, because I thought `target_include_directories` was kinda an implied subset of `target_link_libraries`, so I would've expect this to cause the same problem... but yes, that does seem to work :)

hebasto:
From the CMake [docs](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#target-usage-requirements):
> Generally, a dependency should be specified in a use of [`target_link_libraries()`](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#command:target_link_libraries) with the PRIVATE keyword if it is used by only the implementation of a library, and not in the header files. If a dependency is additionally used in the header files of a library (e.g. for class inheritance), then it should be specified as a PUBLIC dependency.

I believe, that all we need is `target_link_libraries(multiprocess PUBLIC CapnProto::capnp)` without low-level manipulation with `CAPNP_INCLUDE_DIRECTORY`.

purpleKarrot:
> I believe, that all we need is target_link_libraries(multiprocess PUBLIC CapnProto::capnp) without low-level manipulation with CAPNP_INCLUDE_DIRECTORY.

Yes, this is the correct approach. The only valid usecase for `target_include_directories` is to set the target's own include directories. It should never be used for adding include directories originating from another target.

theuni:
Good, thanks for the proper cleanup. I hated this :)

PR #143, opened 2025-01-31
cleanup: initialize vars in the EventLoop constructor in the correct order

theuni:
This legitimate warning pops up when built with clang 20 and Core's flags:

`warning: field 'm_context' will be initialized after field 'm_task_set' [-Wreorder-ctor]`

It looks like there are no dependency issues here and the reordering is safe, but please let me know if I'm wrong.

### DIFF
--- a/src/mp/proxy.cpp
+++ b/src/mp/proxy.cpp
@@ -156,9 +156,9 @@ void Connection::addAsyncCleanup(std::function<void()> fn)
 EventLoop::EventLoop(const char* exe_name, LogFn log_fn, void* context)
     : m_exe_name(exe_name),
       m_io_context(kj::setupAsyncIo()),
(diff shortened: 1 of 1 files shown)


ryanofsky (approved):
Code review ACK f52d08c7f0fda95b25c984d39c5e53ebdc57323f. Looks good, thanks for the fix!

Again will probably merge this Monday in case anyone else wants to review, but can merge sooner if that would be helpful.

hebasto (approved):
ACK f52d08c7f0fda95b25c984d39c5e53ebdc57323f, tested on Ubuntu 24.04:
```
$ cmake -B build -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXX_FLAGS=-Wreorder-ctor
$ cmake --build build
```

PR #146, opened 2025-02-03
cmake: Suppress compiler warnings from capnproto headers

ryanofsky:
Tweak `target_link_libraries` calls to correctly declare dependencies on capnproto libraries as public, so capnproto include directories get correctly added to downstream targets. Without this, compiler warnings can be triggered from capnproto headers because they are not treated like external headers.

Fixes https://github.com/chaincodelabs/libmultiprocess/issues/138

This should be an alternative to #141


hebasto:
Concept ACK.

PR #147, opened 2025-02-03
cmake: EXTERNAL_MPGEN cleanups

ryanofsky:
Suggested by hebasto in https://github.com/chaincodelabs/libmultiprocess/pull/142#pullrequestreview-2588097796


CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ if(Libmultiprocess_ENABLE_CLANG_TIDY)
   set(CMAKE_CXX_CLANG_TIDY "${CLANG_TIDY_EXECUTABLE}")
 endif()
 
-set(EXTERNAL_MPGEN "" CACHE STRING "Use the supplied mpgen binary rather than the one built internally")
```
theuni, line 27: Hmm, I purposefully avoiding namespacing this because I was assuming that parent projects would always want to expose the same option and have it be automatically inherited here. Now they'll have to map it.

Not a big deal to me either way, I guess.

cmake/TargetCapnpSources.cmake
```diff
@@ -61,15 +61,15 @@ function(target_capnp_sources target include_prefix)
...
-  if(EXTERNAL_MPGEN)
-    set(MPGEN_BINARY "${EXTERNAL_MPGEN}")
+  if(Libmultiprocess_EXTERNAL_MPGEN)
+    set(MPGEN_BINARY "${Libmultiprocess_EXTERNAL_MPGEN}")
     if(NOT EXISTS "${MPGEN_BINARY}")
-      message(FATAL_ERROR "EXTERNAL_MPGEN: \"${MPGEN_BINARY}\" does not exist.")
+      message(FATAL_ERROR "Libmultiprocess_EXTERNAL_MPGEN: \"${MPGEN_BINARY}\" does not exist.")
     endif()
   elseif(TARGET Libmultiprocess::multiprocess)
-    set(MPGEN_BINARY $<TARGET_FILE:Libmultiprocess::mpgen>)
```
theuni, line 70: I attempted this a bunch of different ways, and with some approach this was necessary, but I forget why. I assume this is fine, thanks for simplifying.

CMakeLists.txt
ryanofsky: For reference, code which sets this in depends is implemented in https://github.com/bitcoin/bitcoin/commit/99b7b857d3b8b9ce9bd7501e2480583369740c55 from https://github.com/bitcoin/bitcoin/pull/31741 and seems to work reasonably well

theuni: For depends, sure, but when doing a non-cross build, ccmake shows:
```
CAPNPC_CXX_EXECUTABLE
CAPNP_EXECUTABLE
...
Libmultiprocess_EXTERNAL_MPGEN
```

which is rather inconsistent and non-obvious to builders imo. Seems to me Capnp has it right. For a Core, I wouldn't expect to see that namespace.

ryanofsky: I see. This suggestion just came from https://github.com/chaincodelabs/libmultiprocess/pull/142#discussion_r1938116078, and I'm happy to rename this to something else or make a different change. Given the other names maybe `MPGEN_EXECUTABLE` would be a good name for this.

ryanofsky: @hebasto might have more ideas too.

theuni: > I see. This suggestion just came from [#142 (comment)](https://github.com/chaincodelabs/libmultiprocess/pull/142#discussion_r1938116078), and I'm happy to rename this to something else or make a different change. Given the other names maybe `MPGEN_EXECUTABLE` would be a good name for this.

Fwiw, I originally picked `EXTERNAL_CAPNP` to mirror (what I thought was) the way Capnp handles it. But I see now that Capnp's is just a bool and `CAPNP_EXECUTABLE` is the actual path variable. So ACK to `MPGEN_EXECUTABLE` instead if we go down this path.

hebasto: > So ACK to `MPGEN_EXECUTABLE` instead if we go down this path.

So do I.

PR #151, opened 2025-02-04
Fix `-Wsign-compare` compiler warning

hebasto:
Fixes https://github.com/bitcoin/bitcoin/pull/30975#issuecomment-2634356140:
> ... then found another issue:
> 
> ```
> /home/runner/work/_temp/src/ipc/libmultiprocess/src/mp/gen.cpp:633:26: error: comparison of integers of different signs: 'size_t' (aka 'unsigned long') and 'int' [-Werror,-Wsign-compare]
>   633 |     for (size_t i = 4; i < argc; ++i) {
>       |                        ~ ^ ~~~~
> 1 error generated.
> ```

This is the only warning when all default Bitcoin Core [flags](https://github.com/bitcoin/bitcoin/blob/94ca99ac51dddbee79d6409ebcc43b1119b0aca9/CMakeLists.txt#L410-L440) are applied.

### DIFF
--- a/src/mp/gen.cpp
+++ b/src/mp/gen.cpp
@@ -630,7 +630,7 @@ int main(int argc, char** argv)
     } else {
         throw std::runtime_error(std::string("Failed to open src_prefix prefix directory: ") + argv[1]);
     }
-    for (size_t i = 4; i < argc; ++i) {
+    for (size_t i = 4; i < static_cast<size_t>(argc); ++i) {
         KJ_IF_MAYBE(dir, fs->getRoot().tryOpenSubdir(cwd.evalNative(argv[i]))) {
             import_paths.emplace_back(argv[i]);
             import_dirs.emplace_back(kj::mv(*dir));


src/mp/gen.cpp
```diff
@@ -630,7 +630,7 @@ int main(int argc, char** argv)
     } else {
         throw std::runtime_error(std::string("Failed to open src_prefix prefix directory: ") + argv[1]);
     }
-    for (size_t i = 4; i < argc; ++i) {
+    for (size_t i = 4; i < static_cast<size_t>(argc); ++i) {
         KJ_IF_MAYBE(dir, fs->getRoot().tryOpenSubdir(cwd.evalNative(argv[i]))) {
```
ryanofsky, line 634: In commit "Fix `-Wsign-compare` compiler warning" (7452ddd1ee11557de935af821782e9babc7036f5)

Is switching `size_t` on this line to `int` enough to fix the problem? It would seem better to avoid this error by using the same type consistently and avoiding the need to cast.

ryanofsky: re: https://github.com/chaincodelabs/libmultiprocess/pull/151#discussion_r1941803268

Went ahead and made this change in #152 but did not confirm it fixes the warning (since I wasn't seeing the warning). If not, can use the fix here instead.

hebasto: I can confirm that the [commit](https://github.com/chaincodelabs/libmultiprocess/pull/152/commits/0b8b7f92ca39d01f1422bfdbbbf55fb92857b7b9) from https://github.com/chaincodelabs/libmultiprocess/pull/152 fixed the warning.

PR #163, opened 2025-02-20
build: set cmake policy version to 3.31

purpleKarrot:
Please see [`cmake_minimum_required()`](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html) for details.

This PR does not change the miminum required version of CMake; it just changes the policy version, which would otherwise be identical to the minimum required version (3.8).

A policy version of less than 3.10 causes a warning when using CMake 3.31. The policy version should always be set to the latest version that a project has been tested with.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.12...4.1)
 
 project("Libmultiprocess" CXX)
 if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,10 @@
 
 cmake_minimum_required(VERSION 3.12...4.1)
 
+if(BSD AND POLICY CMP0155)
+  cmake_policy(SET CMP0155 OLD)
+endif()
+
 project("Libmultiprocess" CXX)
 if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
   set(CMAKE_CXX_STANDARD 20)


ryanofsky (approved):
Code review ACK 300278daa77e861564ae2f342dd995dfa8f8d171. Thanks for the fix. I've seen this warning before but didn't understand what caused it. This explanation and fix make sense.

hebasto:
> The policy version should always be set to the latest version that a project has been tested with.

Doing so will make the build system's behaviour dependent on the CMake version.

ryanofsky:
> Doing so will make the build system's behaviour dependent on the CMake version.

IIUC, that does seem like a possible downside to this change. Before 300278daa77e861564ae2f342dd995dfa8f8d171 with `cmake_minimum_required(VERSION 3.8)` specified everybody who is using this build will see cmake 3.8 behaviors no matter what version of cmake they are using. But after 300278daa77e861564ae2f342dd995dfa8f8d171 with `cmake_minimum_required(VERSION 3.8...3.31)` people who are using cmake 3.8 will get cmake 3.8 behaviors, people who are using cmake 3.12 will get 3.12 behaviors, 3.20 will see 3.20 behaviors, etc. Cmake will stop trying to be compatible with version 3.8 and only try to be compatible with version 3.31, so people using cmake versions after 3.31 will still see 3.31 behaviors.

On the other hand, it does seem better to use 3.31 behaviors instead of 3.8 behaviors as long as it doesn't cause bugs. Because if it does cause bugs we will need to fix them eventually anyway and it would be better to fix them sooner rather than later. Also in the worst case if any bugs are difficult to fix properly or difficult to fix while maintaining compatibility we can always set  `cmake_policy` `SET` to restore specific old behaviors that we want.

So overall this change seems good, even if it might expose some problems in the short run.

ryanofsky:
I guess another possible downside of this change is it increases the chances that we accidentally break compatibility with old versions of cmake without knowing about it.

For example right now maybe we can be reasonably assumed that we are not depending on cmake 3.20 features because our cmake policy version is 3.8 so those features aren't enabled for us. But if we increase our policy version to 3.31 then new features may become available and we may rely on them without knowing about it. I'm not sure if cmake can warn us when we are relying on a new feature without checking for it and need to increase our minimum version.

ryanofsky:
Would welcome more input on this. I think I'm still inclined to just test this a little bit and merge it and fix any possible bugs that are exposed. But we could also consider raising minimum version or reducing maximum version to try to ensure more predictability in the future. Open to opinions or more information if I'm not understanding this correctly.

purpleKarrot:
Whenever a new version of CMake comes out, you should test your project with that new version **without changing the policy version**.

For all new policies, cmake will execute two code paths: the one with the old behavior and the one with the new behavior. If the results mismatch, cmake will print a warning that your project is affected by the policy and then will use the **old result**. You can silence the warning by setting the policy to `OLD`, but this should be temporarily only, because the old behavior is deprecated by definition.

Once you fixed a policy warning by avoiding ambiguous cmake code, you should set the policy to `NEW`. Once you fixed all policy warnings, you may increase the policy version. You can then be sure that behaviour **does not** dependent on the CMake version.

Note that `cmake_minimum_required(VERSION 3.8)` does not prevent using features that are introduced later. The risk of accidental breakage is real, but it is not increased with this change. The only way to actually guarantee compatibility with version 3.8 is to use that version in CI.

hebasto:
As a part of the Bitcoin Core project, `libmultiprocess` should ensure a stable build environment.

Testing all available policies introduces unneeded burden for developers.

Also see https://github.com/hebasto/bitcoin/pull/143.

purpleKarrot:
> Testing all available policies introduces unneeded burden for developers.

You do that multiple times a day. Every time you run cmake, all available policies are tested. If no policy warnings are printed, it means that you are not affected by the breaking changes those policies are introduced for. You should trust cmake and increase the policy version in this case.

You don't turn cmake into a hermetic buildsystem by keeping policies unset. You should use them the way it is intended by their developers.

ryanofsky:
For now I merged #164 which raises minimum version to 3.12, since that seems like the most obvious fix for the warning.

I am still interested in setting in specifying a max version to use a higher default policy version though. It seems to me like there are different risks to whichever policy version you chose (though the risks are minor and theoretical in all cases). If you choose an old policy version, you are opting into old deprecated behaviors and relying on cmake's ability to emulate old versions of itself. If you set a new policy version, you put developers using older cmake versions at a disadvantage because you aren't testing the old policies they might be using.

IMO the best thing to do would be to disable application of old cmake policies and use non-deprecrated policies in our release build, which uses cmake [3.24.2](https://git.savannah.gnu.org/cgit/guix.git/tree/gnu/packages/cmake.scm?id=53396a22afc04536ddf75d8f82ad2eafa5082725#n169) (it is also 3.24.2 according to `guix time-machine --commit=53396a22afc04536ddf75d8f82ad2eafa5082725 -- shell cmake-minimal -- cmake --version`). So IMO it would be good to set 3.24 as max version, so developers are free to use older or new versions of cmake and their builds will always match the release build as closely as possible.

purpleKarrot:
> Sure, CI should be checking the minimum version, but it can't cover all possible user systems.

It does not have to cover all possible user systems; it has to cover all possible policies. This is archived by testing both `<min>` (where all policies in the `<min>...<max>` range are `OLD`) and `<max>` (where all policies in that range are `NEW`). With only two CI builds, you are covering both `OLD` and `NEW` for all possible policies. That is no burden.

ryanofsky (approved):
Code review ACK f507c50ee5fcf86c666b4046d4f9bac4dcd33b56. Seems like an improvement to me to opt into new, non-deprecated policies by default if we've done a little testing with them and believe the build should work. A theoretical downside of this change is that that it could put users of older cmake versions at a disadvantage, because we will no longer be emulating old versions of cmake in our own builds and might unintentionally break things for them by relying on newer policies. But I don't believe this will happen in practice and if it does happen I think it would be useful to have bug reports so we can know if we are relying on particular policies.

Would also be happy with a different change that sets max version to something less than 3.31 to be more conservative and not use the latest policies, but I don't think it makes sense to use 3.12 policies.

ryanofsky:
> If this PR is merged, what actions will be taken when a new CMake release comes out?

My plan would be to wait for someone to open another PR, wait for it to get acks, and merge it. If someone is interested in using newer cmake policies, they should have motivation to submit a PR. If no one is interested, it should also be ok to stick with oldier policies we are using. But we should not intentionally prefer old and deprecated policies.

ryanofsky:
I think I will probably merge this PR soon. It seems like a safe, practical change to opt the build into more widely used, modern recommended cmake default behaviors instead of enabling dozens of [backwards compatibility policies](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html) that cause modern cmake versions try to emulate a cmake release that is 7 years old.

I do think statement in the PR description "The policy version should always be set to the latest version that a project has been tested with" is probably a little too strong, because this change comes with some downside of making it possible for the build to unintentionally rely on new cmake policies without us realizing it (which is also possible without this change, but less likely). In the end it's probably best to set the max version to a recent version that we think some number of users and developers are using, without needing it to be the very latest version.

fanquake:
Note that the latest available CMake version is now also `4.0.2`.

hebasto:
Here are a couple of examples of policies that change the build system's behaviour depending on the CMake version in use:
- https://cmake.org/cmake/help/latest/policy/CMP0117.html
- https://cmake.org/cmake/help/latest/policy/CMP0141.html

That is the main reason I'm reluctant on use the `...<policy_max>` option.

ryanofsky:
> That is the main reason I'm reluctant on use the `...<policy_max>` option.

CMP0117 stops specifying a redundant rtti flag
CMP0141 stops specifying harmful debug information flags

You didn't say what your reasoning is, but it seems like from these examples you don't trust new cmake releases to set good default policies and want to opt into old policies cmake developers do not recommend... unless those policies are so old that they are below some minimum version of cmake we can't support or are choosing not to support.

This seems like precarious situation to opt into, where you want to allow running new versions of cmake with old policies from some arbitrary version instead the actual default policies they ship with and recommend.

I do understand your previously stated concern that enabling recent policies will make the "build system's behaviour dependent on the CMake version". But that concern exists regardless, not just with cmake but with all the tools we are using. It's a legitimate thing to be worried that the compilation behavior may be dependent on on the version of the compiler that is used. Maybe you are concerned GCC 14 will have some problems that GCC 13 doesn't have. But the ideal solution to that is to use GCC 13, not use GCC 14 with a bunch of flags to emulate old behaviors. That's a reason our official builds use a fixed set of dependencies, including a fixed cmake dependency. But letting version of cmake runtime and library vary up the the bleeding edge while statically setting a much older policy version seems like a dubious practice, IMO.

purpleKarrot:
> But with regard to (2), I am saying in that comment that it does not make sense to me to change minimum required version of cmake from 3.12 to something else when the cmake build here is minimal and only uses rudimentary cmake features.

The current CMake code in this project is a bit more complex than necessary. It kind of abuses the `PUBLIC_HEADER` property for something that it is not intended for and where there is a simper alternative in later versions. It uses complex generator expressions in `target_include_directories` that also have a simpler alternative in later versions. There is nothing we cannot do with CMake version 3.12, but we can greatly simplify the code if we increase the minimum version.

Note that "which version is required to make those simplifications" is the wrong question. The correct approach is to set the minimum according to outside requirements (like the version that is available in DebianTesting) and then make simplifications based on that.

hebasto:
Here is another example of how new policies might be controversial: https://github.com/bitcoin/bitcoin/pull/31504.

cc @theuni @fanquake

CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.12...3.31)
```
fanquake, line 5: Should we increase this to 4.1.1 now, if we are planning to make this change?

Sjors:
I'm generally fine with focussing support on modern environments and only adding support for older stuff if there's a need for it. Since Bitcoin Core requires 3.22 it would make little sense to support anything older.

If CI passes with 4.1 that's fine too.

purpleKarrot:
A new version of CMake is released. Project author tries it out and gets a warning that the project is affected by a new policy. Project author fixes the project so that it is no longer affected by the policy (eg by fixing an invalid regex, by properly quoting a string, by replacing a deprecated command invocation, by searching for a packages in config mode only, or similar). After making the change, CI verifies that the change did not raise the minimum requirement: Using an old versions of CMake still works. Project author increases the policy version to prevent regressions (like the re-introduction of invalid regexes, wrongly quoted strings, deprecated commands, deprecated find-modules, etc).

> ```cmake
> if(...)
>   cmake_minimum_required(...)
> endif()
> ```

Setting the policy version is a statement of the project author that the project is not affected by policy changes. Why should this be set from outside?

Sjors:
@purpleKarrot what you describe makes sense to me, more so than the documentation, maybe we should paste that as a comment :-)

ryanofsky:
> Setting the policy version is a statement of the project author that the project is not affected by policy changes. Why should this be set from outside?

@purpleKarrot it makes sense to me to use the policy version the way you describe, as an upgrade mechanism and way of documenting compatibility with new cmake versions as they are released.

However, the bitcoin core project is not using it this way, because the bitcoin core policy version is hardcoded to 3.22 even though most developers are using newer versions of cmake. Hebasto and marco also have expressed a preference for #175 instead of this PR, setting the libmultiprocess policy version to 3.22 to match bitcoin's policies.

As far as I can tell, hebasto and maybe marco are not looking at the cmake policy version setting as an upgrade mechanism and way to provide a smooth transition to new cmake versions as you describe. Instead they see the cmake policy version as a pinning mechanism: a way to guarantee old cmake behaviors are used regardless of the user's environment and version of cmake they have installed. This way for example, if developers are using new versions of cmake, and users are using old versions of cmake, everybody will be be using old cmake policies, and if there are any problems with those policies, developers will know about them and fix them.

If hebasto wants the bitcoin core policy to be set to 3.22 and for libmultiprocess policy to match this when it is compiled as part of bitcoin core, I don't have a problem with supporting that even though I feel the approach has more downsides than upsides. (Downsides being disabling cmake features and bugfixes, not respecting user preferences, using cmake in less common configuration against cmake developer recommendations. Upside being a form of pinning where there is less variance between builds in different environments.)

hebasto:
> Setting the policy version is a statement of the project author that the project is not affected by policy changes. Why should this be set from outside?

It’s not about setting the policy version from outside. The downstream project is simply asking the project author not to break its currently pinned policy behaviour. The project author chooses a pinned policy version for such a use case at their own discretion.

> If hebasto wants the bitcoin core policy to be set to 3.22 and for libmultiprocess policy to match this when it is compiled as part of bitcoin core...

The code snippet I [suggested](https://github.com/bitcoin-core/libmultiprocess/pull/163#issuecomment-3268184177) does indeed align the policies for Bitcoin Core and libmultiprocess, but that’s just a (reasonable) simplification. All I’m asking is that a pinned policy, chosen in some way, be used when building as a subproject of Bitcoin Core.

ryanofsky:
Thanks for clarification hebasto. That all makes sense to me and explains what you are looking for, except I am confused by the last sentence:

> All I’m asking is that a pinned policy, chosen in some way, be used when building as a subproject of Bitcoin Core.

This last sentence seems to imply that you don't care how the pinned policy is chosen. But you do have a clear preference for pinning the policy to an old version of cmake, not a current version of cmake like this PR.

hebasto:
> > All I’m asking is that a pinned policy, chosen in some way, be used when building as a subproject of Bitcoin Core.
> 
> This last sentence seems to imply that you don't care how the pinned policy is chosen.

That's correct. However, it should be a value within the [3.12..3.22] range.

> But you do have a clear preference for pinning the policy to an old version of cmake, not a current version of cmake like this PR.

How would you suggest implementing policy version pinning in the latter case?

ryanofsky:
Thanks I think I see the discrepancy. I am using the word "pinned" to mean "fixed." This PR is setting a fixed policy version, so I would say this PR is pinning the policy version.

But I think you are using "pinned" to not simply mean "fixed" but "fixed to the minimum version of cmake that does not trigger a build error." You want it to be an error to build with any version of cmake below the specified policy version.

ryanofsky:
I'm starting to work on a PR to implement your suggestion https://github.com/bitcoin-core/libmultiprocess/pull/163#issuecomment-3268184177. The PR will add a new "newdeps" CI job test cmake 4.1 since I don't think any current CI jobs are testing it. (It will also test a newer version of cap'n proto to mirror the "olddeps" CI job)

hebasto:
> @hebasto, I think you are asking for a feature that CMake does not provide. There is no way to tell CMake to behave as version 3.22. The policy mechanism may look like such a feature, but it is something different. CMake actually does provide feature flags for features that are currently under development. But the policies are not feature flags.

Hmm, I'm confused. The quote I mentioned [earlier](https://github.com/bitcoin-core/libmultiprocess/pull/163#issuecomment-3267890115) states that "the project will still build as it did before".

I'd like to avoid, for example, treating a compiler flag differently depending on the policy version.

@purpleKarrot 

Mind elaborating your point please?

ryanofsky:
6e29c02ec5afdb86874e83aa46c93b0d88c27968 is my attempt to implement hebasto's suggestion in https://github.com/bitcoin-core/libmultiprocess/pull/163#issuecomment-3268184177. It uses the purpleKarrot and the cmake documentation's recommended approach by default, but stops overriding the policy version set by parent projects in case they are using different practices. So in the bitcoin core build it update the libmultiprocess policy version from 3.12 to 3.22 like hebasto and marco seem to favor

hebasto:
> > The quote I mentioned [#163 (comment)](https://github.com/bitcoin-core/libmultiprocess/pull/163#issuecomment-3267890115) states that "the project will still build as it did before".
> 
> Yes, in the sense that it will not suddenly start failing or introduce other incompatible changes. It however makes no promise to get exactly the same behavior. If you need exactly the same behavior for reproducibility, you need to pin down the executable.
> 
> To demonstrate, I have installed CMake version 3.22.6 and configured bitcoin core. I have also configured it with CMake version 4.0.3 and compared the results. You can see that there are a number of changes both in the CMake cache and in the generated build system:
> 
> ```
> CMakeCache.txt              |  121 +-                          
> CMakeFiles/rules.ninja      | 1438 ++++++++--                  
> build.ninja                 | 3867 ++++++++++++++++++--------- 
> ```
> 
> I then changed the policy version to 4.0 and configured again. The following changes are not surprising; we expect different behavior when we set a different policy. But the output above should convince you that pinning a policy will not guarantee exact behavior across different versions of CMake.
> 
> ```
> build.ninja                 | 6466 ++++++++++++++++++++++++++++++++++++++---- 
> ```

Thanks for this demonstration.

So, a `CMakeLists.txt` with `cmake_minimum_required(VERSION 3.22)` generates different underlying build system code depending on the CMake binary version being used.

What about the resulting compiler, linker or archiver invocation strings?

ryanofsky:
> the current version of this PR breaks CI

I think an easy to fix for this would be to cherry pick e416d45528faa7d4a15dc6e46e265778e611a26c from #209

hebasto:
@purpleKarrot 
> > the current version of this PR breaks CI
> 
> @hebasto, can you reproduce the issue with a minimal example on freebsd?
> 
> ```cmake
> cmake_minimum_required(VERSION 3.12)
> project(Dummy CXX)
> find_package(Threads REQUIRED)
> ```
> 
> Does this start failing when `...4.1` is added?

Here is an MRE for FreeBSD 14.3 without the `llvm19` package installed (no `clang-scan-deps`):
```cmake
cmake_minimum_required(VERSION 3.12...4.1)
project(Dummy CXX)
set(CMAKE_CXX_STANDARD 20)
find_package(Threads REQUIRED)
```
Run as follows:
```
cmake -B build -G Ninja
```

purpleKarrot:
> invoking it without required is correct because it is not required.

Even if you just want to use it optionally, you want a reliable result whether it is available.

ryanofsky:
Marking as draft since changes still seems to be in progess with 5dfd890f25cbdde8f175b44b6be62b2f5fbb00a6

PR #164, opened 2025-02-22
Bump minimum required cmake to 3.12

maflcko:
This project requires C++20. However, according to https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html, the value `20` was only added in cmake 3.12.

Thus, reflect the reality also in the cmake required minimum version.

As a side-effect this also avoids a deprecation message when compiling with cmake 3.31, according to https://cmake.org/cmake/help/latest/release/3.31.html#deprecated-and-removed-features

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
-cmake_minimum_required(VERSION 3.8)
+cmake_minimum_required(VERSION 3.12)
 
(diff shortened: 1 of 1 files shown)


hebasto (approved):
ACK 729ff16d559c7727e883850c6ed245ec48d11954.

The minimum required CMake version was last bumped to support C++17 in https://github.com/chaincodelabs/libmultiprocess/pull/69.

hodlinator (approved):
ACK 729ff16d559c7727e883850c6ed245ec48d11954

Verified to remove the CMake deprecation warning when running `cmake -B build` with CMake 3.31.5.

PR #166, opened 2025-03-05
doc: rename from chaincodelabs to bitcoin-core

fanquake:


hebasto (approved):
ACK 2e119973cee95abc1923c3583bc25e639392411e.

ryanofsky:
Let me know if you need anything here @fanquake, hopefully you have right permissions to just use https://github.com/bitcoin-core/libmultiprocess/transfer and transfer it. According to https://docs.github.com/en/repositories/creating-and-managing-repositories/transferring-a-repository all accesses to the old location should automatically redirect so I don't think this should cause any major issues.

fanquake:
@ryanofsky I don't think anything else is required. The transfer has happened, so when this is merged, we can just update the subtree in our repo to remove any lingering links.

PR #171, opened 2025-04-14
cmake: Show error instead of invoking cross-compiled mpgen binary

ryanofsky:
Idea here is to avoid potentially confusing errors if cross compiling and forgetting to set the MPGEN_EXECUTABLE.

Idea came from theuni in https://github.com/bitcoin/bitcoin/pull/31741#discussion_r2040214663 with suggestion to show a warning whenever CMAKE_CROSSCOMPILING was set and MPGEN_EXECUTABLE wasn't set, but this would show a warning in the default build which excludes tests and examples and does not need to run the code generator. Also it could warn spuriously if cross-compiled binaries were actually executable.

Avoid these problems by using check_cxx_source_runs instead of CMAKE_CROSSCOMPILING to see if binaries can run and triggering an error at build time instead of configure time to avoid showing warnings when code generator is not actually needed.

This is currently a draft PR, because I haven't tested it on a cross-compiled system so not actually sure it is a significant improvement. Also not sure if this is the best approach.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,17 @@ include(GNUInstallDirs)
 # to avoid "error: Import failed: /mp/proxy.capnp" failures from capnproto.
 set_property(GLOBAL PROPERTY MP_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/include")
 
+# Set MP_NO_GENERATE as a global property if cross compiling and
+# MPGEN_EXECUTABLE not set and cross compiled executables can not run. This
+# should trigger a more readable attempting to run a cross compiled executable.
+if(CMAKE_CROSSCOMPILING AND NOT MPGEN_EXECUTABLE)
+  include(CheckCXXSourceRuns)
+  check_cxx_source_runs("int main() { return 0; }" EXECUTABLES_CAN_RUN)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -35,6 +35,17 @@ include(GNUInstallDirs)
 # to avoid "error: Import failed: /mp/proxy.capnp" failures from capnproto.
 set_property(GLOBAL PROPERTY MP_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/include")
 
+# Set MP_NO_GENERATE as a global property if cross compiling and
+# MPGEN_EXECUTABLE not set and cross compiled executables can not run. This
+# should trigger a more readable attempting to run a cross compiled executable.
+if(CMAKE_CROSSCOMPILING AND NOT MPGEN_EXECUTABLE)
+  include(CheckCXXSourceRuns)
+  check_cxx_source_runs("int main() { return 0; }" EXECUTABLES_CAN_RUN)
```
theuni, line 43: Nice. While looking for something like this, I stumbled upon `try_run` which takes `CMAKE_CROSSCOMPILING` into account, which is why I suggested just testing against the var itself. But it seems `check_cxx_source_runs` is more sane.

PR #175, opened 2025-05-15
Set cmake_minimum_required(VERSION 3.22)

maflcko:
This PR is doing two different things:

* It is switching on cmake policies CMP0076-CMP0128
* It is triggering a fatal error if versions of cmake before 3.22 are used.

This is a follow-up to the last bump in https://github.com/bitcoin-core/libmultiprocess/pull/164.

It is unclear why effort should be made to try to support earlier cmake versions than the ones Bitcoin Core is using (https://github.com/bitcoin/bitcoin/blob/e486597f9a57903600656fb5106858941885852f/CMakeLists.txt#L10), because libmultiprocess is currently only used there.

Moreover, old systems shipping with ancient cmake versions are likely already incompatible due to https://github.com/bitcoin-core/libmultiprocess/pull/205, if they lack the appropriate point release of capnproto.

Once there is a need or use-case to derive from Bitcoin Core, the policy could be changed then.

Bumping the minimum could also unlock simpler cmake code, see https://github.com/bitcoin-core/libmultiprocess/pull/163#issuecomment-2886577273

### DIFF
--- a/ci/scripts/ci.sh
+++ b/ci/scripts/ci.sh
@@ -18,20 +18,10 @@ fi
 [ -n "${CI_CLEAN-}" ] && rm -rf "${CI_DIR}"
 
 cmake --version
-cmake_ver=$(cmake --version | awk '/version/{print $3; exit}')
-ver_ge() { [ "$(printf '%s\n' "$2" "$1" | sort -V | head -n1)" = "$2" ]; }
 
 src_dir=$PWD
 mkdir -p "$CI_DIR"
 cd "$CI_DIR"
 cmake "$src_dir" "${CMAKE_ARGS[@]+"${CMAKE_ARGS[@]}"}"
-if ver_ge "$cmake_ver" "3.15"; then
-  cmake --build . -t "${BUILD_TARGETS[@]}" -- "${BUILD_ARGS[@]+"${BUILD_ARGS[@]}"}"
-else
-  # Older versions of cmake can only build one target at a time with --target,
-  # and do not support -t shortcut
-  for t in "${BUILD_TARGETS[@]}"; do
-    cmake --build . --target "$t" -- "${BUILD_ARGS[@]+"${BUILD_ARGS[@]}"}"
-  done
-fi
+cmake --build . -t "${BUILD_TARGETS[@]}" -- "${BUILD_ARGS[@]+"${BUILD_ARGS[@]}"}"
 ctest --output-on-failure
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.22)
 
 project("Libmultiprocess" CXX)
 if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
--- a/ci/configs/olddeps.bash
+++ b/ci/configs/olddeps.bash
@@ -4,5 +4,5 @@ CI_DIR=build-olddeps
 # requires an older GCC.
 NIXPKGS_CHANNEL=nixos-25.05
 export CXXFLAGS="-Werror -Wall -Wextra -Wpedantic -Wno-unused-parameter -Wno-error=array-bounds"
-NIX_ARGS=(--argstr capnprotoVersion "0.9.2" --argstr cmakeVersion "3.12.4" --argstr gccVersion "11")
+NIX_ARGS=(--argstr capnprotoVersion "0.9.2" --argstr cmakeVersion "3.22.6" --argstr gccVersion "11")
 BUILD_ARGS=(-k)
--- a/shell.nix
+++ b/shell.nix
@@ -54,7 +54,7 @@ let
   clang = if enableLibcxx then llvm.libcxxClang else llvm.clang;
   clang-tools = llvm.clang-tools.override { inherit enableLibcxx; };
   cmakeHashes = {
-    "3.12.4" = "sha256-UlVYS/0EPrcXViz/iULUcvHA5GecSUHYS6raqbKOMZQ=";
+    "3.22.6" = "sha256-c5MxY2cOpOqVwjFUkAewxyQygik1BqLPRENxSCatXsM=";
   };
   gcc = if gccVersion == null then null else builtins.getAttr ("gcc" + gccVersion) pkgs;
   cmakeBuild = if cmakeVersion == null then pkgs.cmake else (pkgs.cmake.overrideAttrs (old: {


hebasto (approved):
ACK 3f386cc3e31bd049eb0d069644a4d7188053b4bb.

ryanofsky:
Concept ~0 assuming #163 is merged first.

This PR is doing two different things (see https://github.com/bitcoin-core/libmultiprocess/pull/163#issuecomment-2884186860)

1. It is switching on cmake policies CMP0076-CMP0128
2. It is triggering a fatal error if versions of cmake before 3.22 are used.

I think the first change is good. The second change seems a like it introduces misleading documentation and an inappropriate error message, but it is not very damaging and seems ok if we want to be paranoid about old versions of cmake doing bad things.

I would like to see #163 merged first because it is only doing the first thing, not the second thing and I do think it is good to keep these things separate for clarity.

fanquake:
ACK 3f386cc3e31bd049eb0d069644a4d7188053b4bb - just because it would now match what Core is setting, which seems reasonable.

hebasto (approved):
ACK 34f48d1e8f85dcd471c16a14b0bae01c1f2d561d.

ryanofsky:
I'm not sure what to do about this PR because I feel like it is not describing itself or the code accurately.

From the PR description it sounds like it is only dropping support for old versions of cmake but this is not the case, since the changing poilcy version affects behavior of all versions of cmake including the newest ones. I don't think there is a good reason to tie changes to the policy version and minimum version together, because they are separate concepts and have different effects. 

Writing `cmake_minimum_required(VERSION 3.22)` also seems misleading because it implies the cmake build requires features present in 3.22, which it doesn't. This could be fixed by adding a comment, though.

maflcko:
The current status is that there are two reviews, and a suggestion to update the description, which I've done.

maflcko:
This has two acks and no open feedback, so I wonder if there is a blocker before merge?

in the meantime, the leveldb subtree also has this set:

```diff
diff --git a/src/leveldb/CMakeLists.txt b/src/leveldb/CMakeLists.txt
index cfd4faa325..78efde95de 100644
--- a/src/leveldb/CMakeLists.txt
+++ b/src/leveldb/CMakeLists.txt
@@ -2,19 +2,25 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file. See the AUTHORS file for names of contributors.
 
-cmake_minimum_required(VERSION 3.9)
+cmake_minimum_required(VERSION 3.22)

ryanofsky:
> This has two acks and no open feedback, so I wonder if there is a blocker before merge?

Thanks for following up. I haven't merged this PR or known how to give more constructive feedback about it, because I'm still confused about the motivation behind it. The PR description has improved since the last time I saw it, but it is only offering rationalizations about why these changes are not harmful, not explaining why these changes are actually good or desirable.

From my perspective, the PR is making two distinct changes that are confusing to bundle together, and one of them seems bad (breaks working builds and misleadingly documents build requirements), and other one might be good, but the thinking behind it is not explained (it chooses a newer policy version number but doesn't explain how the number was chosen, or how the number should be chosen in the future, why cmake developer guidelines are ignored, or what the tradeoffs are).

Since I don't understand the motivation behind this PR, I'm not sure I can make a suggestion that would accomplish whatever goals it is trying to accomplish. But I don't think I'd have any issue with this if were limited to just raising the policy version, since I think everybody agrees policy version is probably lower than it should be.

maflcko:
The motivation boils down to this in the pull description: "It is unclear why effort should be made to try to support earlier cmake versions".

That is, I think there is no value in writing extra code (even if it is "only" CI code) to try to support ancient cmake versions, where it is known that no user is running them.

I don't really see an alternative here. One could just commit the `ci/scripts/ci.sh` changes, and a ci-only and nix-only bump to 3.15, but then `CMakeLists.txt` would contain an untested `cmake_minimum_required`, so I think all the changes should go together. The only alternative I see here is using 3.15 instead of 3.22, to do the minimal bump needed to be able to drop the extra code.

Though, that seems backward. The correct approach is to pick the version, based on outside requirements and then apply the simplifications afterward, see the second section of https://github.com/bitcoin-core/libmultiprocess/pull/163#issuecomment-2886577273.

ryanofsky:
re: https://github.com/bitcoin-core/libmultiprocess/pull/175#issuecomment-4602675647

I do not mean to be uncharitable, but this response seems to be full of false assumptions.

There is no "correct approach" here, there are tradeoffs between different approaches.

There is no "extra code" here, there is just code that supports different use-cases.

The claim that "no user is running [the current minimum required version]" is not true because I am a user and I have a branch and PR testing that version. I also view code as documentation and prefer code that documents true things ("X is required") instead of false things ("X is not required but if we pretend it is, the simplest part of CI system can become slightly simpler".)

The sentence "I don't really see an alternative here" seems misguided because this PR is not solving any current problem.

I am guessing we have different viewpoints because we have different backgrounds. I think you are more used to a debian model where an upstream provider chooses decides which versions of packages should be used together and makes customization relatively difficult. My background is using configure/make/install, then gentoo, then nix, which all make it easy mix and match versions and apply customization of different kinds.

If the goal of changing the minimum version and policy version together is just to simplify the CI scripts slightly, when the most complicated part of the CI system by far is the sprawling YAML configuration, not the 20 line testing script, then I'd suggest closing this PR. It is not worth falsely stating requirements or changing the policy version (which affects builds in a myriad of ways) to do this. 

I'm completely happy with raising the minimum version when there is a real reason to raise it, but slightly simplifying a CI script doesn't seem like a real reason given that it changes which build policies are used in non-CI builds, breaks currently working builds, and misstates requirements.

maflcko:
> the most complicated part of the CI system by far is the sprawling YAML configuration

I am happy to review a pull request simplifying this. Feel free to use an LLM bot for this. I think most of this can be simplified, according to https://github.com/bitcoin-core/libmultiprocess/pull/253#issuecomment-4134235478. Would be interesting to see if there are any obstacles.

> I'd suggest closing this PR

Sure, if it is too subjective/controversial.  I'd presume the same change change will be picked up later when this is bumped to cmake 3.23 (or later) when https://cmake.org/cmake/help/latest/command/target_sources.html#file-sets FILE_SETs are made use of.

maflcko:
> The claim that "no user is running [the current minimum required version]" is not true because I am a user and I have a branch and PR testing that version.

I mean practically, a user does not exist that only has access to this version and would not be possible to compile the program if the minimum would be increased.

Simply weighing that against the extra cost (code, maintenance), e.g. see https://github.com/bitcoin-core/libmultiprocess/pull/296#issuecomment-4677757396, which I can't reproduce locally, I don't see why it is worth it to spend so much time on the ancient cmake version, when there is no real practical benefit.

hebasto (approved):
re-ACK faae6ebe303cb8d7a52bcc002dffb543c261804c.

maflcko:
Looks like the latest push to drop the old nix channel fails with:        error: gcc11 has been removed from Nixpkgs, as it is unmaintained and obsolete ...

ci/scripts/ci.sh
maflcko: .

PR #179, opened 2025-06-02
scripted-diff: Remove copyright year (ranges)

maflcko:
The years are problematic, because they can mean different things to different people, they are not needed, inconsistent and redundant.

So just remove them with a scripted-diff, which can be checked via a script from the Bitcoin Core repo: `../bitcoin-core/test/lint/commit-script-check.sh HEAD~..HEAD`

To give some more context:

* Anyone who wants to find the initial file creation date, or file history, can use `git log` or `git blame` to get more accurate results.
* Some files use stale ranges (`src/mp/util.cpp`), others do not, so it is unclear if ranges need to be added or updated, and when.

So just remove the years (or ranges) in this project with an easy scripted-diff.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2019 The Bitcoin Core developers
+# Copyright (c) The Bitcoin Core developers
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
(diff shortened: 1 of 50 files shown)


ryanofsky:
Code review ACK fa2ff9a66842c094060bd90ddf61232e078e4072. Makes sense to get rid of these instead of bumping them.

> they can mean different things to different people

I don't get how there are multiple ways to interpret the dates. Copyright lasts for a period of time and the dates tell you when the period begins. If there's a range or list of dates, the older dates tell when you older parts of the file will enter public domain, and the latest date tells you when the whole file will. Since the dates are only accurate if we bump them periodically, or whenever editing, and it would be annoying to do this, it's nice to drop them. But I don't think they were ambiguous, just wrong.

PR #181, opened 2025-06-03
 type-function.h: Fix CustomBuildField overload

ryanofsky:
Fix `std::function` `CustomBuildField` overload which is incompatible with a recent change in 3a96cdc18f2d1ca202fbc91551f27097fd7ec7f6 from https://github.com/bitcoin-core/libmultiprocess/pull/172 which changed generated IPC client code to pass it an rvalue `std::function` reference instead of an lvalue reference.

Motivation for this change is to avoid a build error in https://github.com/bitcoin/bitcoin/pull/29409, when rebased on top of https://github.com/bitcoin/bitcoin/pull/32641 which includes https://github.com/bitcoin-core/libmultiprocess/pull/172.


sedited (approved):
ACK 688140b1dffc78a97a5a16f62b540a55824f4091

PR #184, opened 2025-06-20
Add CI jobs and fix clang-tidy and iwyu errors

ryanofsky:
The main thing this change does is add a simple github actions CI job to build the project, run tests, and do checks with iwyu and clang-tidy.

It also fixes bitcoin CI warnings reported by fanquake in https://github.com/bitcoin/bitcoin/pull/31802#issuecomment-2988011933 that show up in the logs but do not cause errors, as well as a number of other warnings that do not seem to show up in bitcoin CI. Individual commit messages describe the fixes in detail.


ryanofsky:
Not sure why CI does not seem to run as part of this PR, but CI was successful in the branch behind the PR:

- https://github.com/ryanofsky/libmultiprocess/actions/runs/15780752720
- https://github.com/ryanofsky/libmultiprocess/commits/refs/heads/pr/ci
 
Maybe this PR has to be merged first before CI will run on any PRs.

ryanofsky:
I'll submit a patch to cap'n proto to avoid the `clang-analyzer-core.UndefinedBinaryOperatorResult` error that's disabled in the last commit of this PR 48ecb5f88989807fb7ce5672b60ca9b48f6b9454. Commit cb94a432971a0e3416da927e16776ddc85dcda15 tests a patch that [seems to work](https://github.com/ryanofsky/libmultiprocess/actions/runs/15782063369).

Ideally we'd never see this error because cap'n proto headers are included with `-isystem` and we disable warnings from system headers, but clang seems to have quirks which cause this error to be associated with the "main file" of our translation unit and shows this error anyway without a way of filtering it out: https://reviews.llvm.org/D26418

ryanofsky:
> See https://github.com/bitcoin-core/libmultiprocess/settings/actions

Thanks! That seems right. I can't change the setting there since I see "This setting has been disabled by organization administrators" but I assume if gets changed from "Disable actions" to one of the other settings this should start to work.

ryanofsky:
re: https://github.com/bitcoin-core/libmultiprocess/pull/184#issuecomment-2992034325

> I'll submit a patch to cap'n proto to avoid the `clang-analyzer-core.UndefinedBinaryOperatorResult` error that's disabled in the last commit of this PR

Opened https://github.com/capnproto/capnproto/pull/2334 with this

---

Updated 48ecb5f88989807fb7ce5672b60ca9b48f6b9454 -> 80f1c2ef32423f53f0dfff0aeb9ce8b02376a873 ([`pr/ci.1`](https://github.com/ryanofsky/libmultiprocess/commits/pr/ci.1) -> [`pr/ci.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/ci.2), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/ci.1..pr/ci.2)) just updating the last commit message to mention the patch

Sjors:
When I update the subtree to use this in https://github.com/bitcoin/bitcoin/pull/31802 my macOS build fails:

```
/Users/sjors/dev/bitcoin/src/ipc/capnp/protocol.cpp:45:42: error: no matching constructor for initialization of 'std::unique_lock<std::mutex>'
   45 |             std::unique_lock<std::mutex> lock(m_loop->m_mutex);
```

Is this expected?

ryanofsky (commented):
Updated 80f1c2ef32423f53f0dfff0aeb9ce8b02376a873 -> f15724a3ddc7ac7330c8aedb9f11c14361a9133d ([`pr/ci.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/ci.2) -> [`pr/ci.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/ci.3), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/ci.2..pr/ci.3), [CI](https://github.com/ryanofsky/libmultiprocess/actions/runs/15832327926)) expanding ci job to cover libc++ and gcc, rewriting a commit message, and making a few other tweaks.

---

re: https://github.com/bitcoin-core/libmultiprocess/pull/184#issuecomment-2995772352

> When I update the subtree to use this in [bitcoin/bitcoin#31802](https://github.com/bitcoin/bitcoin/pull/31802) my macOS build fails:

I misread this error and didn't realize it was coming from the bitcoin build not the libmultiprocess build, and assumed it might be caused a libc++ / libstdc++ difference (which is why I added libstdc++ ci job).

But I think this error is just caused by #160 and should be fixed by https://github.com/bitcoin/bitcoin/commit/d234efe92ec8b9f2bad63b6be46309b7cccccd24 from https://github.com/bitcoin/bitcoin/pull/32345

---

re: https://github.com/bitcoin-core/libmultiprocess/pull/184#issuecomment-2995545454

> Actions should be available in this repository now.

Thanks I saw the setting on https://github.com/bitcoin-core/libmultiprocess/settings/actions was changed from "Disable actions" to "Allow bitcoin-core actions and reusable workflows." Though after another push it still seems like PR actions don't run here, but not sure if that is because of settings, or because no actions are defined on the master branch right now. 

To experiment I did change "Allow bitcoin-core, and select non-bitcoin-core, actions and reusable workflows" and filled in "Allow specified actions and reusable workflows" list with:

- actions/checkout@*
- cachix/install-nix-action@*,

So maybe that will have an effect if there is another push.

hebasto:
> > Actions should be available in this repository now.
> 
> Thanks I saw the setting on https://github.com/bitcoin-core/libmultiprocess/settings/actions was changed from "Disable actions" to "Allow bitcoin-core actions and reusable workflows." Though after another push it still seems like PR actions don't run here, but not sure if that is because of settings, or because no actions are defined on the master branch right now.
> 
> To experiment I did change "Allow bitcoin-core, and select non-bitcoin-core, actions and reusable workflows" and filled in "Allow specified actions and reusable workflows" list with:
> 
>     * actions/checkout@*
> 
>     * cachix/install-nix-action@*,
> 
> 
> So maybe that will have an effect if there is another push.

Perhaps closing the PR and reopening it would help?

In any case, I suggest submitting a separate PR that only introduces the new workflow as a stub, without the `Run CI script` step, merging it promptly to ensure that all settings are correct.

ci/scripts/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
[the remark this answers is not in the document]
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/184#discussion_r2163585250

> I guess it is just me not knowing from reading this line if `-x` means "eXit on error" or "eXecution tracing" or something else (same for the other options). But I can look it up, and this is just a style nit.

Makes sense, latest push switches to long form

include/mp/proxy-types.h
```diff
@@ -92,7 +92,7 @@ struct StructField
 template <typename LocalType, typename EmplaceFn>
 struct ReadDestEmplace
 {
-    ReadDestEmplace(TypeList<LocalType>, EmplaceFn&& emplace_fn) : m_emplace_fn(emplace_fn) {}
+    ReadDestEmplace(TypeList<LocalType>, EmplaceFn emplace_fn) : m_emplace_fn(std::move(emplace_fn)) {}
 
     //! Simple case. If ReadField impementation calls this construct() method
```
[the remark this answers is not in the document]
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/184#discussion_r2163106467

> unrelated nit, found by the llm: ` impementation -> implementation [spelling error in comment]

Thanks fixed in latest push

ryanofsky (commented):
Updated f15724a3ddc7ac7330c8aedb9f11c14361a9133d -> 52f036f94b7f90b157533d771842ef7697073d75 ([`pr/ci.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/ci.3) -> [`pr/ci.4`](https://github.com/ryanofsky/libmultiprocess/commits/pr/ci.4), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/ci.3..pr/ci.4), [CI](https://github.com/ryanofsky/libmultiprocess/actions/runs/15853170216)) making suggested changes, adding readme, consolidating ci scripts.

---

re: https://github.com/bitcoin-core/libmultiprocess/pull/184#issuecomment-2999034667

> The linter trips up over the ci file names:

Thanks, renamed the files to fix this.

> I was unable to run the nix ci scripts on macOS:

Thanks for testing, it looks like the build is clashing with an existing "build" directory. I updated scripts to use separate build directories now to avoid this problem.

> It also fails on Ubuntu 25.04 with nix 2.14.1:

I had a difficult time trying to reproduce this. I tried single & multiuser nix installs and did not see this error. But then I noticed ubuntu seems to have a `nix-setup-systemd` package and saw the reported error ` error: file 'nixpkgs' was not found in the Nix search path (add it using $NIX_PATH or -I)` because it looks the ubuntu setup does not install nixpkgs. Adding a nix channel manually fixed the problem:

```bash
nix-channel --add https://nixos.org/channels/nixpkgs-unstable
nix-channel --update
```

---

re: https://github.com/bitcoin-core/libmultiprocess/pull/184#issuecomment-2997677389

> Should probably add documentation about this

Added ci/README.md file now

---

re: https://github.com/bitcoin-core/libmultiprocess/pull/184#issuecomment-3000434063

> Perhaps closing the PR and reopening it would help?

Thanks that is a good idea, will try that! I would like to avoid opening a separate PR so this can be reviewed in a single round. Since CI does run successfully in the branch pushes, and I'm not too concerned about CI status not showing up on the bottom of this page.

ci/configs/llvm.sh
```diff
@@ -0,0 +1,15 @@
+CI_DESC="CI job using LLVM-based libraries and tools (clang, libc++, clang-tidy, iwyu) and testing Ninja"
+CI_DIR=build-llvm
+NIX_ARGS=(
+  --arg libcxx true
+)
+CMAKE_ARGS=(
+  -G Ninja
+  -DCMAKE_CXX_COMPILER=clang++
+  -DCMAKE_CXX_STANDARD_LIBRARY=libc++
```
hebasto, line 9: I doubt that relying on the undocumented `CMAKE_CXX_STANDARD_LIBRARY` behaves as one might expect in this context. We definitely want:

```suggestion
  -DCMAKE_CXX_COMPILER="clang++;-stdlib=libc++"
```

CMakeLists.txt
```diff
@@ -51,6 +67,7 @@ configure_file(include/mp/config.h.in "${CMAKE_CURRENT_BINARY_DIR}/include/mp/co
 
 # Generated C++ Capn'Proto schema files
 capnp_generate_cpp(MP_PROXY_SRCS MP_PROXY_HDRS include/mp/proxy.capnp)
+set_source_files_properties("${MP_PROXY_SRCS}" PROPERTIES SKIP_LINTING TRUE)
```
hebasto, line 70: Same.

ci/configs/llvm.sh
hebasto: The suggested approach makes `-stdlib=libc++` a permanent part of the compiler invocation, whereas `CXXFLAGS` may be altered by the build system.

hebasto: > If you're saying CMake might randomly alter those flags...

I'm not blaming CMake. Just assuming that `CMakeListss.txt` can potentially process `CMAKE_CXX_FLAGS` in some way. I agree this sounds a bit vague. If it helps, here's a quote form the CMake [docs](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER.html):
> **Note:** Options that are required to make the compiler work correctly can be included as items in a list; they can not be changed.

I'm not opposed to using your suggestion.

ryanofsky: FWIW, neither of these suggested changes seem to work and both result in:

```
clang++: error: argument unused during compilation: '-stdlib=libc++' [-Werror,-Wunused-command-line-argument]
```

because the stdlib option is a linker option not a compile option. If I add the flag to CMAKE_EXE_LINKER_FLAGS it avoids that error, but then linking fails with lots of undefined symbol errors because libstdc++ headers are included but libc++ is linked.

Setting CMAKE_CXX_STANDARD_LIBRARY seems to be a clean way of choosing the library and avoiding this problem, and is only being set as part of this CI configuration, not part of the project itself. So I think I will just add a comment remarking that this option is experimental and may not work in the future.

There seem to be other ways of choosing the library (https://chatgpt.com/share/685bf4fc-b99c-800a-9342-d1d7458178b5) but this is simpler than any alternative and will probably be documented eventually and if it stops working the CI job can be changed without affecting the project itself.

ryanofsky (commented):
Thanks for the reviews!

Updated 52f036f94b7f90b157533d771842ef7697073d75 -> d219fec64a5b21762db0d07a1af12729fb6a3287 ([`pr/ci.4`](https://github.com/ryanofsky/libmultiprocess/commits/pr/ci.4) -> [`pr/ci.5`](https://github.com/ryanofsky/libmultiprocess/commits/pr/ci.5), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/ci.4..pr/ci.5), [CI](https://github.com/ryanofsky/libmultiprocess/actions/runs/15878535152)) squashing commits and applying suggestions

ci/configs/llvm.sh
ryanofsky: > Looking into the verbose build log, I still can't convince myself that using the `CMAKE_CXX_STANDARD_LIBRARY` variable is a valid approach.

Not sure if you saw the chatgpt summary, but it linked to https://gitlab.kitware.com/cmake/cmake/-/issues/20851 which implements this setting and https://cmake.org/cmake/help/latest/manual/cmake-variables.7.html which says internal variables "are subject to change, and not recommended for use in project code."

This PR is only using the setting in a CI config file that's not part of project code and nothing downstream will use or rely on. So I don't think there is a problem here, but if there is one I could try an alternate approach or just drop commit 403f63a891b1cf0a41e2771ef99261e34ae1f082 since it is not essential

include/mp/proxy.h
```diff
@@ -7,35 +7,31 @@
...
-#include <array>
 #include <cassert>
 #include <functional>
 #include <list>
+#include <memory>
 #include <stddef.h>
 #include <tuple>
 #include <type_traits>
 #include <utility>
+#include <variant>
```
hebasto, line 18: IWYU's suggestion here appears to be incorrect.

example/example.cpp
```diff
@@ -2,12 +2,17 @@
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
+#include <init.capnp.h>
+#include <init.capnp.proxy.h>
+
+#include <cstring>
```
hebasto, line 8: IWYU's suggestion here appears to be incorrect.

src/mp/gen.cpp
```diff
@@ -198,19 +201,40 @@ static void Generate(kj::StringPtr src_prefix,
 
     std::ofstream cpp_server(output_path + ".proxy-server.c++");
     cpp_server << "// Generated by " PROXY_BIN " from " << src_file << "\n\n";
-    cpp_server << "#include <" << include_path << ".proxy-types.h>\n";
+    cpp_server << "// IWYU pragma: no_include <kj/memory.h>\n";
+    cpp_server << "// IWYU pragma: no_include <memory>\n";
+    cpp_server << "#include <" << include_path << ".proxy.h>\n";
+    cpp_server << "#include <" << include_path << ".proxy-types.h> // IWYU pragma: keep\n";
+    cpp_server << "#include <capnp/generated-header-support.h>\n";
+    cpp_server << "#include <cstring>\n";
```
hebasto, line 209: Same here.

src/mp/gen.cpp
```diff
@@ -198,19 +201,40 @@ static void Generate(kj::StringPtr src_prefix,
...
     cpp_client << "// Generated by " PROXY_BIN " from " << src_file << "\n\n";
-    cpp_client << "#include <" << include_path << ".proxy-types.h>\n";
-    cpp_client << "#include <" << PROXY_TYPES << ">\n\n";
+    cpp_client << "// IWYU pragma: no_include <kj/memory.h>\n";
+    cpp_client << "// IWYU pragma: no_include <memory>\n";
+    cpp_client << "#include <" << include_path << ".h> // IWYU pragma: keep\n";
+    cpp_client << "#include <" << include_path << ".proxy.h>\n";
+    cpp_client << "#include <" << include_path << ".proxy-types.h> // IWYU pragma: keep\n";
+    cpp_client << "#include <capnp/generated-header-support.h>\n";
+    cpp_client << "#include <cstring>\n";
```
hebasto, line 226: Same here.

ci/configs/llvm.sh
ryanofsky: Oh nice find. So I guess the option doesn't even do anything here? The other changes in shell.nix do seem to work and Cap'n Proto binaries are built with libc++, but I can confirm ldd shows libmultiprocess binaries using a mix of libc++ and libstdc++ so CMAKE_CXX_STANDARD_LIBRARY is not having the desired effect.

```$ ldd build-llvm/test/mptest 
	linux-vdso.so.1 (0x00007f781796b000)
	libkj-test.so.1.1.0 => /nix/store/q4dhry7vwy2pyybgipnqmnrwnc3x62qn-capnproto-1.1.0/lib/libkj-test.so.1.1.0 (0x00007f781781a000)
	libcapnp-rpc.so.1.1.0 => /nix/store/q4dhry7vwy2pyybgipnqmnrwnc3x62qn-capnproto-1.1.0/lib/libcapnp-rpc.so.1.1.0 (0x00007f7817711000)
	libcapnp.so.1.1.0 => /nix/store/q4dhry7vwy2pyybgipnqmnrwnc3x62qn-capnproto-1.1.0/lib/libcapnp.so.1.1.0 (0x00007f7817652000)
	libkj-async.so.1.1.0 => /nix/store/q4dhry7vwy2pyybgipnqmnrwnc3x62qn-capnproto-1.1.0/lib/libkj-async.so.1.1.0 (0x00007f7817572000)
	libkj.so.1.1.0 => /nix/store/q4dhry7vwy2pyybgipnqmnrwnc3x62qn-capnproto-1.1.0/lib/libkj.so.1.1.0 (0x00007f78174de000)
	libpthread.so.0 => /nix/store/cg9s562sa33k78m63njfn1rw47dp9z0i-glibc-2.40-66/lib/libpthread.so.0 (0x00007f78174d9000)
	libstdc++.so.6 => /nix/store/7c0v0kbrrdc2cqgisi78jdqxn73n3401-gcc-14.2.1.20250322-lib/lib/libstdc++.so.6 (0x00007f7817200000)
	libm.so.6 => /nix/store/cg9s562sa33k78m63njfn1rw47dp9z0i-glibc-2.40-66/lib/libm.so.6 (0x00007f7817117000)
	libgcc_s.so.1 => /nix/store/7c0v0kbrrdc2cqgisi78jdqxn73n3401-gcc-14.2.1.20250322-lib/lib/libgcc_s.so.1 (0x00007f78174ab000)
	libc.so.6 => /nix/store/cg9s562sa33k78m63njfn1rw47dp9z0i-glibc-2.40-66/lib/libc.so.6 (0x00007f7816e00000)
	/nix/store/cg9s562sa33k78m63njfn1rw47dp9z0i-glibc-2.40-66/lib/ld-linux-x86-64.so.2 => /nix/store/cg9s562sa33k78m63njfn1rw47dp9z0i-glibc-2.40-66/lib64/ld-linux-x86-64.so.2 (0x00007f781796d000)
	libc++.so.1 => /nix/store/1gpni0qga8q9jkagxr9m4xzfzanq8avf-libcxx-20.1.5/lib/libc++.so.1 (0x00007f7816ceb000)
	libc++abi.so.1 => /nix/store/1gpni0qga8q9jkagxr9m4xzfzanq8avf-libcxx-20.1.5/lib/libc++abi.so.1 (0x00007f78170cc000)
	libatomic.so.1 => /nix/store/7c0v0kbrrdc2cqgisi78jdqxn73n3401-gcc-14.2.1.20250322-lib/lib/libatomic.so.1 (0x00007f781749d000)
```

I'm very surprised calling a c++ library which is compiled with libc++ from c++ code compiled with libstdc++ could even work, and had assumed if the build was working and tests were passing that would imply the same library was used everywhere. But I guess mixing libraries works because cap'n proto c++ api avoids types like std::string and std::vector, so the two runtime libraries don't actually need to interact with each other.

I'll see if there is some simple way to fix this, otherwise I'll just drop the commit 403f63a891b1cf0a41e2771ef99261e34ae1f082 from the PR. Thanks for catching the problem!

include/mp/proxy.h
hebasto: > IWYU's suggestion here appears to be incorrect.

I agree with @maflcko. I didn't mean the current code should be changed. Just pointed at potential IWYU bug.

ci/configs/llvm.sh
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/184#discussion_r2167002748

> The using of the libstdc++ is obvious.

This should be fixed now. Now it links with only libc++ not libc++ and libstdc++ when I run locally.

re: https://github.com/bitcoin-core/libmultiprocess/pull/184#discussion_r2167002748

> That means, if you set `CMAKE_CXX_FLAGS` (which you should never do), then `CXXFLAGS` may not be used (wich is the reason why you should not set `CMAKE_CXX_FLAGS`).

@purpleKarrot just saw this comment (sorry the thread appears all mixed up). It sound like you're asking to change the two config files to use CXXFLAGS environment variable instead of CMAKE_CXX_FLAGS command line option. I think that makes sense because even though CMAKE_CXX_FLAGS is empty by default, hypothetically if a toolchain or similar file might wanted to influence `CMAKE_CXX_FLAGS` with `CMAKE_CXX_FLAGS_INIT` the command line here would clobber that value while the environment variable would be combined with it.

I guess this logic also extends to CMAKE_CXX_COMPILER and CXX would be recommended instead?

I'll try switching to environment variables in the next push if that makes sense.

include/mp/proxy.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/184#discussion_r2166968873

Makes sense, added keep pragma.

example/example.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/184#discussion_r2166969762

Added keep pragma here too

src/mp/gen.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/184#discussion_r2166970927

Added keep pragma here too

src/mp/gen.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/184#discussion_r2166971405

Added keep pragma here too

ryanofsky (commented):
Thanks for the reviews!

Updated d219fec64a5b21762db0d07a1af12729fb6a3287 -> aaf1db6b2baff38f008b216f00998fd49dcf131c ([`pr/ci.5`](https://github.com/ryanofsky/libmultiprocess/commits/pr/ci.5) -> [`pr/ci.6`](https://github.com/ryanofsky/libmultiprocess/commits/pr/ci.6), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/ci.5..pr/ci.6), [CI](https://github.com/ryanofsky/libmultiprocess/actions/runs/15883905261)) fixing libc++ build and making IWYU work with both libc++ and libstdc++ now.

ryanofsky (commented):
Updated aaf1db6b2baff38f008b216f00998fd49dcf131c -> 1bfed25a14762b7b2f4d11a1faaa86679af0707c ([`pr/ci.6`](https://github.com/ryanofsky/libmultiprocess/commits/pr/ci.6) -> [`pr/ci.7`](https://github.com/ryanofsky/libmultiprocess/commits/pr/ci.7), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/ci.6..pr/ci.7), [CI](https://github.com/ryanofsky/libmultiprocess/actions/runs/15884579059)) implementing suggestion use environment variables instead of command line arguments for compiler settings and earlier suggestion to drop -DCMAKE_EXPORT_COMPILE_COMMANDS=ON

hebasto (approved):
ACK 1bfed25a14762b7b2f4d11a1faaa86679af0707c.

Sjors:
ACK 757e13a7554644d7ee1e3e85a9c56beeaffd5249

I ran the three ci scripts on Ubuntu.

We'll need to be careful when updating the subtree in https://github.com/bitcoin/bitcoin/pull/32345. But debugging that should be easier with the new CI coverage added in this PR.

ryanofsky:
Problem with github actions seems to be fixed. CI jobs appear at the bottom of a PR I just opened https://github.com/bitcoin-core/libmultiprocess/pull/186. I did updated the list of allowed actions in settings to `actions/checkout@*,cachix/install-nix-action@*` separated by a comma instead of a newline, and that might have helped.

PR #193, opened 2025-08-12
build: require CapnProto 0.7.0 or better

Sjors:
Although 1.0.1. is the oldest version currently covered by Bitcoin Core's extensive CI, Debian Bookwork ships 0.9.2 and #194 introduces test coverage for even older versions. 0.7 has been required since https://github.com/bitcoin-core/libmultiprocess/pull/88.

The CI run of https://github.com/Sjors/bitcoin/pull/100 @ [3d55222](https://github.com/Sjors/bitcoin/pull/100/checks?sha=3d552223712eed88d17e5ead1ef7d1ba6fd7e89e) previously checked Bitcoin Core CI against 1.0.1 as the minimum. Lowering the minimum further should not be a problem for that CI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@ endif()
 
 include("cmake/compat_find.cmake")
 
-find_package(CapnProto REQUIRED)
+find_package(CapnProto 0.7.0 REQUIRED)
 find_package(Threads REQUIRED)
 
 set(MPGEN_EXECUTABLE "" CACHE FILEPATH "If specified, should be full path to an external mpgen binary to use rather than the one built internally.")


Sjors:
Even if we want to go lower, older than v0.8.1 wouldn't make sense if it has a CVE: https://github.com/bitcoin/bitcoin/issues/33176#issuecomment-3180162395

Sjors:
Lowered to 0.9.2 to support Bookworm.

Unless there's another distro out there that has 0.8.2, I think we should hold off on lowering further until Ubuntu 24.04 LTS actually ships it.

ryanofsky (approved):
Code review ACK b8d438736928c92393bec8ac3123477714d0329b

re: some of the comments above, I don't think the CVE should be a factor in what minimum version is required because the fixes listed in https://github.com/capnproto/capnproto/security/advisories/GHSA-qqff-4vw4-f6hx go all the way back to 0.5.

IMO, it'd be nice with https://github.com/bitcoin-core/libmultiprocess/pull/194 to only require version 0.7 or later, so IPC just works without headaches on a wide variety of OS's including Ubuntu 22.04. Version 0.7 has been required since https://github.com/bitcoin-core/libmultiprocess/pull/88.

But this PR looks goods good as-is and unless there are any objections I'll merge it soon. Versions of 0.7 and 0.8 already don't work currently due to the bug in https://github.com/bitcoin/bitcoin/issues/33176, so this PR is just accurately documenting current requirements.

Sjors:
Ok, if you think the CVE can safely be worked around, then I'm fine with supporting older versions. And good point about it not being easy to rule out specific versions.

I would sleep a bit better if this repo had test coverage for these older versions, which #194 also introduces.

Sjors:
Lowered it to 0.7.0 in anticipation of the CI improvements in #194.

ryanofsky:
> Ok, if you think the CVE can safely be worked around, then I'm fine with supporting older versions.

I think I need to read more to know if we are affected. https://capnproto.org/news/2022-11-30-CVE-2022-46149-security-advisory.html says "The vulnerability is exploitable only if an application performs a certain unusual set of actions." and I haven't looked into what those are.

Regardless, I think the build should refuse to use any version affected by the CVE. It will just require a custom check that I can add in https://github.com/bitcoin-core/libmultiprocess/pull/194. I think I'll go ahead and merge this now to work on that. Thanks for the PR!

ryanofsky (approved):
Code review ACK 30930dff7b06be5ffe28fbef535676e73ee923e2. Planning to follow up in #194 to actually test minimum version and error if capnproto version detected is affected by CVE-2022-46149

ryanofsky:
This has been merged and #194 updated based on this, now triggering an error if any version affected by the CVE is detected.

Note: I edited PR description before merging this, adding "0.7 has been required since https://github.com/bitcoin-core/libmultiprocess/pull/88." just to give an explanation of where the 0.7 requirement comes from.

PR #194, opened 2025-08-12
mpgen: Work around c++20 / capnproto 0.8 incompatibility

ryanofsky:
Fix ubuntu 22.04 compatibility issue reported by fanquake in https://github.com/bitcoin/bitcoin/issues/33176 and add CI job to make sure libmultiprocess stays compatible with old versions of Cap'n Proto.

Changes are described in more detail in commit messages.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,8 +12,29 @@ endif()
 
 include("cmake/compat_find.cmake")
 
-find_package(CapnProto 0.7.0 REQUIRED)
 find_package(Threads REQUIRED)
+find_package(CapnProto 0.7 REQUIRED)
+
+# Check for list-of-pointers memory access bug from Nov 2022
+# https://nvd.nist.gov/vuln/detail/CVE-2022-46149
+# https://github.com/advisories/GHSA-qqff-4vw4-f6hx
+# https://github.com/capnproto/capnproto/security/advisories/GHSA-qqff-4vw4-f6hx
+# https://github.com/capnproto/capnproto/blob/master/security-advisories/2022-11-30-0-pointer-list-bounds.md
+# https://capnproto.org/news/2022-11-30-CVE-2022-46149-security-advisory.html
+# https://dwrensha.github.io/capnproto-rust/2022/11/30/out_of_bounds_memory_access_bug.html
+if(CapnProto_VERSION STREQUAL "0.7.0"
+   OR CapnProto_VERSION STREQUAL "0.8.0"
+   OR CapnProto_VERSION STREQUAL "0.9.0"
+   OR CapnProto_VERSION STREQUAL "0.9.1"
+   OR CapnProto_VERSION STREQUAL "0.10.0"
+   OR CapnProto_VERSION STREQUAL "0.10.1"
+   OR CapnProto_VERSION STREQUAL "0.10.2")
+  message(FATAL_ERROR
+    "Cap'n Proto ${CapnProto_VERSION} is affected by CVE-2022-46149.\n"
+    "Please install an updated package.\n"
+    "Details: https://github.com/advisories/GHSA-qqff-4vw4-f6hx
+  ")
+endif()
 
 set(MPGEN_EXECUTABLE "" CACHE FILEPATH "If specified, should be full path to an external mpgen binary to use rather than the one built internally.")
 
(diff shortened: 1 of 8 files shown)


shell.nix
```diff
@@ -2,12 +2,39 @@
...
+  };
+  capnprotoBase = if capnprotoVersion == null then crossPkgs.capnproto else crossPkgs.capnproto.overrideAttrs (old: {
+    version = capnprotoVersion;
+    src = crossPkgs.fetchFromGitHub {
+      owner = "capnproto";
+      repo  = "capnproto";
+      rev   = "v${capnprotoVersion}";
+      hash  = lib.attrByPath [capnprotoVersion] "" capnprotoHashes;
+    };
+    patches = lib.optionals (lib.versionAtLeast capnprotoVersion "0.9.0" && lib.versionOlder capnprotoVersion "0.10.4") [ ./ci/patches/spaceship.patch ];
```
Sjors, line 33: 88d95044a6cbf180f6c4cf6ed2e38e39c5bf4869: do I read this correctly as: apply patch to versions 0.9.0 <= ... < 0.10.4?

Sjors:
utACK 88d95044a6cbf180f6c4cf6ed2e38e39c5bf4869

ryanofsky (commented):
Rebased 88d95044a6cbf180f6c4cf6ed2e38e39c5bf4869 -> dc3ba2204606d8b5b681a6798407e68b3567e7a9 ([`pr/string.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/string.3) -> [`pr/string.4`](https://github.com/ryanofsky/libmultiprocess/commits/pr/string.4), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/string.3-rebase..pr/string.4)) on top of #193 and added explicit check for CVE-2022-46149

sedited (approved):
ACK dc3ba2204606d8b5b681a6798407e68b3567e7a9

CMakeLists.txt
```diff
@@ -12,8 +12,29 @@ endif()
...
+# https://capnproto.org/news/2022-11-30-CVE-2022-46149-security-advisory.html
+# https://dwrensha.github.io/capnproto-rust/2022/11/30/out_of_bounds_memory_access_bug.html
+if(CapnProto_VERSION STREQUAL "0.7.0"
+   OR CapnProto_VERSION STREQUAL "0.8.0"
+   OR CapnProto_VERSION STREQUAL "0.9.0"
+   OR CapnProto_VERSION STREQUAL "0.9.1"
+   OR CapnProto_VERSION STREQUAL "0.10.0"
+   OR CapnProto_VERSION STREQUAL "0.10.1"
+   OR CapnProto_VERSION STREQUAL "0.10.2")
+  message(FATAL_ERROR
```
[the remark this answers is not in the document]
ryanofsky: For ubuntu 22.04 if this will create confusion, I feel like a good solution might be to improve error message and documentation to specifically suggest disabling ENABLE_IPC, or installing a newer version of capnproto. Ideally it seems like ubuntu could update from 0.8.0 to 0.8.1 given 0.8.1 was released to fix the CVE and doesn't have any other changes.

On IPC being a trusted interface, it's definitely true with the current interface, a malicious client could do a lot of things to crash the node and maybe even take control of it. But this doesn't have to be the case. We could provide capnproto interfaces that do rigorously check all their inputs and enforce resource constraints. It would even be possible to adapt the Mining interface to do this, but it hasn't been a reason to make the extra effort. I also feel like showing an error after discovering a package with a CVE is kind of a public service, since we don't know if there may be other software using the package.

src/mp/gen.cpp
ryanofsky, line 1: In commit "mpgen: Work around c++20 / capnproto 0.8 incompatibility" (c4cb758eccb56f0d0c8bf26197156dfb5f6cb21c)

I think commit message is wrong in suggesting that with c++20 the `operator==` function is buggy and will call itself recursively. It seems the reason this happens is not really because of c++20 but because of a gcc bug https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53499. See also https://github.com/bitcoin/bitcoin/issues/33176#issuecomment-3194776823

ryanofsky:
re: @willcl-ark https://github.com/bitcoin-core/libmultiprocess/pull/194#issuecomment-3241833841

> I'm not sure, as you say this was fixed upstream in v0.9 , but I _think_ I might still be hitting this on debian bookworm with 0.9.2?:

If I'm interpreting correctly, debian bookworm uses clang 14 according to https://packages.debian.org/bookworm/clang so maybe doesn't have this issue by default, but I think you will run into this problem with clang-16 and newer. I made a new issue for this https://github.com/bitcoin-core/libmultiprocess/issues/199 so maybe we can figure out a solution there

PR #195, opened 2025-08-20
ci: Add openbsd

maflcko:
Requested in https://github.com/bitcoin/bitcoin/issues/33219#issuecomment-3201847735

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,34 @@ on:
   pull_request:
 
 jobs:
+  build-openbsd:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: openbsd {0}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Start OpenBSD VM
+        uses: vmactions/openbsd-vm@v1
+        with:
+          prepare: |
+            pkg_add -v cmake ninja git python bash
+          run: |
+            git clone --depth=1 https://codeberg.org/OpenBSD/ports.git /usr/ports
+          sync: 'rsync'
+          copyback: false
+
+      - name: Install capnproto
+        run: |
+          cd /usr/ports/devel/capnproto/
+          make install
+
+      - name: Run CI script
+        run: |
+          cd ${{ github.workspace }}
+          CI_CONFIG="ci/configs/openbsd.bash" bash ci/scripts/ci.sh
+
   build:
     runs-on: ubuntu-latest
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,11 +7,12 @@ on:
 jobs:
   build-openbsd:
     runs-on: ubuntu-latest
+    name: build • openbsd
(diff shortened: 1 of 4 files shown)


ryanofsky:
> It does fail. See the previous comment on why the CI is not run here.

This is great! Thanks for adding this. I added vmactions/openbsd-vm@* to https://github.com/bitcoin-core/libmultiprocess/settings/actions, so maybe CI could be triggered here now. Could consider adding the #include fix from https://github.com/bitcoin/bitcoin/issues/33219#issuecomment-3202305628 here too

ryanofsky:
I am also just now realizing the gnu32 job is very slow when run CI because it seems to be building a lot of things. It is fast when I run it locally, but that's because the nix outputs are cached. Should look into how to set up caching in CI.

.github/workflows/ci.yml
```diff
@@ -5,6 +5,34 @@ on:
   pull_request:
 
 jobs:
+  build-openbsd:
+    runs-on: ubuntu-latest
+    defaults:
```
hebasto, line 10: 98414e7d28672aea7a218c691d34295d8934ae09:

For consistency with the naming of other jobs:: 

```suggestion
    runs-on: ubuntu-latest
    name: build • openbsd
    defaults:
```

hebasto (approved):
ACK 486a510bbeff76a9cdb3498798217238631bb70e.

hebasto (approved):
ACK eed42f210d173f0049f238fe4ac13fba935aa917.

.github/workflows/ci.yml
```diff
@@ -5,6 +5,34 @@ on:
...
+
+      - name: Install capnproto
+        run: |
+          cd /usr/ports/devel/capnproto/
+          make install
+
+      - name: Run CI script
+        run: |
+          cd ${{ github.workspace }}
+          CI_CONFIG="ci/configs/openbsd.bash" bash ci/scripts/ci.sh
```
ryanofsky, line 35: In commit "ci: Add openbsd" (98414e7d28672aea7a218c691d34295d8934ae09)

Note: It seems to be necessary to set CI_CONFIG this way instead of with `env:` because only github environment variables and environment variables in `envs:` are shared with the vm according to https://github.com/vmactions/openbsd-vm/blob/main/README.md

ryanofsky (approved):
Code review ACK eed42f210d173f0049f238fe4ac13fba935aa917. Nice changes!

.github/workflows/ci.yml
```diff
@@ -5,6 +5,34 @@ on:
...
+          copyback: false
+
+      - name: Install capnproto
+        run: |
+          cd /usr/ports/devel/capnproto/
+          make install
+
+      - name: Run CI script
+        run: |
+          cd ${{ github.workspace }}
```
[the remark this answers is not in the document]
hebasto: When `defaults.run.shell` runs, the working directory is `/root`, so `cd ${{ github.workspace }}` is still necessary.

PR #197, opened 2025-08-26
ci: Add freebsd and macos build

maflcko:
Should be harmless to add and may be useful

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
         uses: vmactions/openbsd-vm@v1
         with:
           prepare: |
-            pkg_add -v cmake ninja git python bash
+            pkg_add -v cmake ninja git bash
           run: |
             git clone --depth=1 https://codeberg.org/OpenBSD/ports.git /usr/ports
           sync: 'rsync'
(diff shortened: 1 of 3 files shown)


Sjors:
Concept ACK

This may also help debug why (so far) I haven't been able to run the nix based tests on macOS.

Sjors:
> Nix seems more niche on macos and users are more likely to use brew

That's probably true.

It seems useful to test on macOS _without_ nix.

hebasto (approved):
ACK fa1ac300005530d4a9e816234df50abeaf6ddcd4.

ryanofsky (approved):
Code review ACK fa1ac300005530d4a9e816234df50abeaf6ddcd4. Seems good to cover these different package managers and platforms

.github/workflows/ci.yml
```diff
@@ -34,6 +34,45 @@ jobs:
...
+
+      - name: Install dependencies
+        env:
+          HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: 1
+        run: |
+          brew install --quiet ninja capnp
+
+      - name: Run CI script
+        run: |
+          CI_CONFIG="ci/configs/macos.bash" bash ci/scripts/ci.sh
```
Sjors, line 74: I tested that this also works locally on macOS.

Sjors:
ACK fa1ac300005530d4a9e816234df50abeaf6ddcd4

Issue #189, opened 2025-07-10
Races between `~ProxyClient<Thread>` and `~Connection` destructors


ryanofsky:
#186 fixed several race conditions that happen during the ["disconnecting and blocking"](https://github.com/bitcoin-core/libmultiprocess/blob/a11e6905c238dc35a8bbef995190296bc6329d49/test/mp/test/test.cpp#L251) unit test, but a new one was reported in https://github.com/Sjors/bitcoin/pull/90#issuecomment-3024942087 with details in https://github.com/Sjors/bitcoin/pull/90#issuecomment-3029059790 and I want to write more notes about it. There is also a [branch](https://github.com/ryanofsky/libmultiprocess/commits/pr/threadclean.2) with previously attempted fixes

The problem is not too hard to reproduce locally: if `mptest` binary is run in in a loop in bash, the tsan error happens in a few minutes. With [attempted fixes so far](https://github.com/ryanofsky/libmultiprocess/commits/pr/threadclean.2) running the mptest binary in a loop results in a deadlock after about ~20 minutes for me.

### Background
If a `Connection` object is destroyed while a `ProxyClient<Thread>` object is still using it (due to a sudden disconnect), the `ProxyClient<Thread>::m_disconnect_cb` callback will be called from `~Connection` destructor, deleting the `ProxyClient<Thread>` object from whichever `ConnThreads` map it is in.

If a `ProxyClient<Thread>` object is destroyed before its `Connection` object (the more normal case), the `ProxyClient<Thread>` destructor will unregister `m_disconnect_cb` so when the `Connection` is later destroyed it will not try to delete the `ProxyClient<Thread>` object.

If the two destructors are called in sequence in either order, everything is fine. But if the destructors are called from different threads in a race, numerous problems occur.

### Problem 1: invalid `m_disconnect_cb` iterator erase

If `~Connection` is destroyed first and is about to call the `m_disconnect_cb` callback, but then the `~ProxyClient` destructor is called in another thread, it will currently (as of a11e6905c238dc35a8bbef995190296bc6329d49) try to delete the `m_disconnect_cb` using an iterator which is no longer valid. 79c09e44420a79bb7ac56960fe94a16849f04bc3 fixes this by changing the `~ProxyClient<Thread>` destructor to avoid making deleting `m_disconnect_cb` once it is invalidated.

### Problem 2: `ProxyClient<Thread>` use-after-free

[...]

PR #201, opened 2025-09-02
bug: fix mptest hang, ProxyClient<Thread> deadlock in disconnect handler

ryanofsky:
This bug is currently causing mptest "disconnecting and blocking" test to occasionally hang as reported by maflcko in https://github.com/bitcoin/bitcoin/issues/33244.

The bug was actually first reported by Sjors in https://github.com/Sjors/bitcoin/pull/90#issuecomment-3024942087 and there are more details about it in https://github.com/bitcoin-core/libmultiprocess/issues/189.

The bug is caused by the "disconnecting and blocking" test triggering a disconnect right before a server IPC call returns. This results in a race between the IPC server thread and the `onDisconnect` handler in the event loop thread both trying to destroy the server's `request_threads` `ProxyClient<Thread>` object when the IPC call is done.

There was a lack of synchronization in this case, fixed here by adding `loop->sync()` various places. There were also lock order issues where `Waiter::m_mutex` could be incorrectly locked before `EventLoop::m_mutex` resulting in a deadlock.

### DIFF
--- a/ci/configs/sanitize.bash
+++ b/ci/configs/sanitize.bash
@@ -1,7 +1,7 @@
 CI_DESC="CI job running ThreadSanitizer"
 CI_DIR=build-sanitize
 export CXX=clang++
-export CXXFLAGS="-ggdb -Werror -Wall -Wextra -Wpedantic -Wthread-safety-analysis -Wno-unused-parameter -fsanitize=thread"
+export CXXFLAGS="-ggdb -Werror -Wall -Wextra -Wpedantic -Wthread-safety -Wno-unused-parameter -fsanitize=thread"
 CMAKE_ARGS=()
 BUILD_ARGS=(-k -j4)
 BUILD_TARGETS=(mptest)
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -537,8 +537,10 @@ void ProxyServerBase<Interface, Impl>::invokeDestroy()
 //! Map from Connection to local or remote thread handle which will be used over
 //! that connection. This map will typically only contain one entry, but can
 //! contain multiple if a single thread makes IPC calls over multiple
-//! connections.
-using ConnThreads = std::map<Connection*, ProxyClient<Thread>>;
+//! connections. A std::optional value type is used to avoid the map needing to
+//! be locked while ProxyClient<Thread> objects are constructed, see
+//! ThreadContext "Synchronization note" below.
+using ConnThreads = std::map<Connection*, std::optional<ProxyClient<Thread>>>;
 using ConnThread = ConnThreads::iterator;
 
 // Retrieve ProxyClient<Thread> object associated with this connection from a
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -534,6 +534,10 @@ void ProxyServerBase<Interface, Impl>::invokeDestroy()
     CleanupRun(m_context.cleanup_fns);
 }
 
+//! Map from Connection to local or remote thread handle which will be used over
+//! that connection. This map will typically only contain one entry, but can
+//! contain multiple if a single thread makes IPC calls over multiple
+//! connections.
 using ConnThreads = std::map<Connection*, ProxyClient<Thread>>;
 using ConnThread = ConnThreads::iterator;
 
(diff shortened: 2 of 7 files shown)


ryanofsky:
Updated fe1cbedc78c8eadcd668fb6cac244d63747f1503 -> 1d6d854047ab2c6c06f36e5a21fa381969abb483 ([`pr/context.1`](https://github.com/ryanofsky/libmultiprocess/commits/pr/context.1) -> [`pr/context.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/context.2), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/context.1..pr/context.2)) fixing iwyu error and making minor cleanups

Eunovo (approved):
TestedACK https://github.com/bitcoin-core/libmultiprocess/pull/201/commits/1d6d854047ab2c6c06f36e5a21fa381969abb483

As I understand it, there are two problems that are solved by this PR:
- a kind of use-after-free bug where the disconnect_cb is deleted in one thread while it is being executed in another
- a deadlock caused by two threads competing for the same waiter mutex

I left comments with more details under this review.

TheCharlatan:
Can confirm that this fixes the observed test hang.

include/mp/proxy-io.h
```diff
@@ -66,8 +66,6 @@ struct ProxyClient<Thread> : public ProxyClientBase<Thread, ::capnp::Void>
     ProxyClient(const ProxyClient&) = delete;
     ~ProxyClient();
 
-    void setDisconnectCallback(const std::function<void()>& fn);
```
Sjors, line 69: In 7a8a99c50feca10be4e42a1cdb135864ce376f71 _bug: fix ProxyClient<Thread> deadlock if disconnected as IPC call is returning_: can you dedicate a sentence in the commit message as to why this function was dropped?

include/mp/proxy-io.h
```diff
@@ -537,8 +535,10 @@ void ProxyServerBase<Interface, Impl>::invokeDestroy()
 //! Map from Connection to local or remote thread handle which will be used over
 //! that connection. This map will typically only contain one entry, but can
 //! contain multiple if a single thread makes IPC calls over multiple
-//! connections.
-using ConnThreads = std::map<Connection*, ProxyClient<Thread>>;
+//! connections. A std::optional value type is used to avoid the map needing to
+//! be locked while ProxyClient<Thread> objects are constructed, see
+//! ThreadContext "Synchronization note" below.
+using ConnThreads = std::map<Connection*, std::optional<ProxyClient<Thread>>>;
```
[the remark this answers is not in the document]
Sjors: I split off an additional commit 94fe2251f6bc629f98d7ab71b09a8c0b5459d150 that switches the early return to `try_emplace`. Now the remaining changes are nice and short: 5b65dae936a53a2c6e43ed49f1f382f919979514

Sjors:
utACK f219f1b51c85e5c4a7157c78d278f832eb0bd975

I left some suggestions to make the intermediate commits more clear, which might benefit other reviewers.

include/mp/proxy-io.h
Sjors: I agree it's nicer to drop it.

include/mp/proxy-io.h
```diff
@@ -307,7 +305,7 @@ struct Waiter
     //! mutexes than necessary. This mutex can be held at the same time as
     //! EventLoop::m_mutex as long as Waiter::mutex is locked first and
     //! EventLoop::m_mutex is locked second.
-    std::mutex m_mutex;
+    Mutex m_mutex;
     std::condition_variable m_cv;
     std::optional<kj::Function<void()>> m_fn;
```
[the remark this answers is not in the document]
hodlinator: (I guess that doesn't work because of `std::condition_variable` <-> `MP_GUARDED_BY` interaction. Please resolve).

ryanofsky (commented):
Thanks for the reviews! I took Sjors commits to break the change up into smaller parts and addressed hodlinators suggestions to add more thread safety annotations and make them more strict. I think both changes should make this PR easier to review.


Updated f219f1b51c85e5c4a7157c78d278f832eb0bd975 -> b6321bb99c8a79400460899eaf3b5e9b15e41be7 ([`pr/context.4`](https://github.com/ryanofsky/libmultiprocess/commits/pr/context.4) -> [`pr/context.5`](https://github.com/ryanofsky/libmultiprocess/commits/pr/context.5), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/context.4..pr/context.5)) with suggested changes to split commits and improve thread safety annotations. This actual fix is not changed.


Updated b6321bb99c8a79400460899eaf3b5e9b15e41be7 -> d86823eed705056e430a6c1c3cd676c360f6a0ae ([`pr/context.5`](https://github.com/ryanofsky/libmultiprocess/commits/pr/context.5) -> [`pr/context.6`](https://github.com/ryanofsky/libmultiprocess/commits/pr/context.6), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/context.5..pr/context.6)) fixing typos

Sjors:
@ryanofsky that commit also touches non-test code, so I think it would be better as a standalone PR.

ryanofsky:
> @ryanofsky that commit also touches non-test code, so I think it would be better as a standalone PR.

That's reasonable. But to be fair the only changes to non-test code are adding 4 lines like:

```c++
if (connection->m_loop->m_signal) connection->m_loop->m_signal(SignalCallSetup{});
```

so the test is able to disconnect at specific points. `m_signal` is always null unless the test sets it, so it should be pretty clear that these lines only affect tests. In practice I'd also split up the change into nontest and test-only commits. And I'd add hasSignal()/sendSignal() helpers so signal mechanism wouldn't be exposed and this would look more like:

```c++
if (connection->m_loop->hasSignal<CallSetup>()) connection->m_loop->sendSignal<CallSetup>());
```

ci/configs/sanitize.bash
```diff
@@ -1,7 +1,7 @@
 CI_DESC="CI job running ThreadSanitizer"
 CI_DIR=build-sanitize
 export CXX=clang++
-export CXXFLAGS="-ggdb -Werror -Wall -Wextra -Wpedantic -Wthread-safety-analysis -Wno-unused-parameter -fsanitize=thread"
+export CXXFLAGS="-ggdb -Werror -Wall -Wextra -Wpedantic -Wthread-safety -Wno-unused-parameter -fsanitize=thread"
```
Eunovo, line 4: https://github.com/bitcoin-core/libmultiprocess/pull/201/commits/4e365b019a9feb3150ac8de089014e719f08c9c4:
This is unrelated to the PR, but why generate debugging information here? I'm referring to the use of the `-ggdb` flag

Eunovo (commented):
ACK https://github.com/bitcoin-core/libmultiprocess/pull/201/commits/4a269b21b8c8c2c7138df82c8dc4103a388e339b

PR #204, opened 2025-09-03
cmake: support default build and test workflow

purpleKarrot:
Make sure that the default workflow of `cmake ..; make; make test` does not fail with an error message that no tests are found.

Remove any "convenience" targets that add maintenance burden and require tribal knowledge.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -216,4 +216,4 @@ add_dependencies(install-lib multiprocess)
 
 # Example and test subdirectories
 add_subdirectory(example EXCLUDE_FROM_ALL)
-add_subdirectory(test EXCLUDE_FROM_ALL)
+add_subdirectory(test)
--- a/ci/scripts/ci.sh
+++ b/ci/scripts/ci.sh
@@ -12,7 +12,7 @@ set -o errexit -o nounset -o pipefail -o xtrace
 
 : "${CI_DIR:=build}"
 if ! [ -v BUILD_TARGETS ]; then
-  BUILD_TARGETS=(all tests mpexamples)
+  BUILD_TARGETS=(all mpexamples)
 fi
 
 [ -n "${CI_CLEAN-}" ] && rm -rf "${CI_DIR}"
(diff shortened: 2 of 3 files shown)


ryanofsky:
Code review 47deaef36c2b720cece2a471d23cca4c79dcf617.

Thanks for the PR. These changes seem worth considering, and it does seem pretty indefensible that "make test" doesn't build the tests it depends on. This behavior seems to be a [longstanding bug or quirk](https://stackoverflow.com/questions/733475/cmake-ctest-make-test-doesnt-build-tests) of cmake, and it's especially surprising given how cmake's "make test" target works differently than its "make install" target when it comes to building dependencies.

I think it could be good to implement the "fixtures" idea in the linked stackoverflow thread to address this, but not good to build targets by default that aren't needed for installation. I think it's best for the 'make; make install` workflow to be fast and reliable and not do unnecessary work.

I'm also not sure about the other part of this change removing convenience targets. `make check` seems like a good thing to support given cmake's `make test` limitations, but maybe it could go away if `make test` is improved, and maybe the other targets are overkill. The convenience targets are useful for CI as well as development, though, which is why several CI jobs are currently broken with this change.

hebasto:
CI fails:
```
 make: *** No rule to make target 'tests'.
 ```

purpleKarrot:
Thanks for the review, @ryanofsky.

> longstanding bug or quirk (link to stackoverflow)

Stackoverflow is not a good source of information when it comes to CMake (neither is chatgpt, which seems to be trained on content from stackoverflow as well as projects that followed the advices found there). Too often, the accepted answers are outdated or plain wrong. In this particular case, what you are looking for is [`CMAKE_SKIP_TEST_ALL_DEPENDENCY`](https://cmake.org/cmake/help/latest/variable/CMAKE_SKIP_TEST_ALL_DEPENDENCY.html).

> The convenience targets are useful for CI

This is exactly my concern. The project configuration is intertwined with CI scripts. What I aim for, is that projects and CI scripts are developed independently against a defined interface so that CI scripts can be reused for multiple projects.
The defined interface for CMake projects is that tests can be executed after the default target is built.

#### About `make test`:

The `test` target just invokes `ctest` without any options. When you execute `make test -j20`, make will try to spawn up to 20 processes in parallel, but it only has one process to spawn: `ctest`. And it does not pass any options to that command, so the tests are invoked without parallelization. Is that what you wanted? Maybe. But maybe not.

Developers really should learn what command line options are supported when ctest is invoked directly, like `ctest -j20`.

Issue #199, opened 2025-09-01
Cap'n Proto 0.9 & 0.10 capnp/blob.h error: invalid operands to binary expression ('const char *' and 'kj::StringPtr')


ryanofsky:
Cap'n Proto versions 0.9.x and 0.10.x are incompatible with clang 16 and later when `-std=c++20` is used, causing error:

```c++
/usr/include/capnp/blob.h:176:74: error: invalid operands to binary expression ('const char *' and 'kj::StringPtr')
```

This was reported by willcl-ark in https://github.com/bitcoin-core/libmultiprocess/pull/194#issuecomment-3241833841 and maflcko in https://github.com/bitcoin-core/libmultiprocess/pull/194#issuecomment-3241965569

This error can be avoided by any of the following:

- Upgrading to Cap'n Proto version 1.0 or newer which fixed this in [capnproto/capnproto#1618](https://github.com/capnproto/capnproto/pull/1618), or applying the [patch](https://github.com/capnproto/capnproto/commit/4b35480e18af80b9f0f0c3b9013c2e1ab6d5d7c2.diff) as done in [shell.nix](https://github.com/bitcoin-core/libmultiprocess/blob/1b8d4a6f1e54b92708bd2ad627ec6d440a1daf3d/shell.nix#L33).
- Downgrading to Cap'n Proto version 0.8 or earlier, before [capnproto/capnproto#1170](https://github.com/capnproto/capnproto/pull/1170) which triggers this issue starting in version 0.9.
- Downgrading to clang-15 or earlier, before [P2468R2](https://wg21.link/p2468r2) which triggers this issue starting in [version 16](https://releases.llvm.org/16.0.0/tools/clang/docs/ReleaseNotes.html#c-language-changes).
- Maybe using GCC (not sure what the status of this is with GCC)

### Background

- Libmultiprocess only uses basic Cap'n Proto features, and for the most part doesn't rely on its c++20 behavior, though it did recently work around a c++20 GCC bug https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53499 in #194
- Cap'n Proto started to improve it's C++20 support in [capnproto/capnproto#1170](https://github.com/capnproto/capnproto/pull/1170) and [capnproto/capnproto#1235](https://github.com/capnproto/capnproto/pull/1235) which were both included in version 0.9.0. But unfortunately these changes stopped working well when clang-16 was released because it implemented P2468R2.
[...]

PR #205, opened 2025-09-03
cmake: check for Cap'n Proto / Clang / C++20 incompatibility

willcl-ark:
Cap'n Proto 0.9.x and 0.10.x are incompatible with Clang 16+ when using C++20 due to P2468R2 implementation. This was fixed in Cap'n Proto 1.0+.

Generate a helpful error when these combinations are detected.

See: https://github.com/bitcoin-core/libmultiprocess/issues/199 for more context

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,15 @@ endif()
 include("cmake/compat_find.cmake")
 
 find_package(Threads REQUIRED)
-find_package(CapnProto 0.7 REQUIRED NO_MODULE)
+find_package(CapnProto 0.7 QUIET NO_MODULE)
+ if(NOT CapnProto_FOUND)
+   message(FATAL_ERROR
+     "Cap'n Proto is required but was not found.\n"
+     "To resolve, choose one of the following:\n"
+     "  - Install Cap'n Proto (version 1.0+ recommended)\n"
+     "  - For Bitcoin Core compilation build with -DENABLE_IPC=OFF to disable multiprocess support\n"
+   )
+ endif()
 
 # Cap'n Proto compatibility checks
 set(CAPNPROTO_ISSUES "")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,11 @@ include("cmake/compat_find.cmake")
 find_package(Threads REQUIRED)
 find_package(CapnProto 0.7 REQUIRED NO_MODULE)
 
+# Cap'n Proto compatibility checks
+set(CAPNPROTO_ISSUES "")
+set(CAPNPROTO_CVE_AFFECTED FALSE)
+set(CAPNPROTO_CLANG_INCOMPATIBLE FALSE)
+
 # Check for list-of-pointers memory access bug from Nov 2022
 # https://nvd.nist.gov/vuln/detail/CVE-2022-46149
 # https://github.com/advisories/GHSA-qqff-4vw4-f6hx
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -36,6 +36,25 @@ if(CapnProto_VERSION STREQUAL "0.7.0"
...
+   (CMAKE_CXX_COMPILER_ID STREQUAL "Clang") AND
+   (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "16") AND
+   (CMAKE_CXX_STANDARD EQUAL 20))
+  message(FATAL_ERROR
+    "Cap'n Proto ${CapnProto_VERSION} is incompatible with Clang ${CMAKE_CXX_COMPILER_VERSION} when using C++20.\n"
+    "To resolve this issue, choose one of the following:\n"
+    "  1. Upgrade to Cap'n Proto version 1.0 or newer (recommended)\n"
+    "  2. Downgrade to Cap'n Proto version 0.8 or earlier\n"
+    "  3. Use GCC instead of Clang\n"
+    "  4. For Bitcoin Core compilation build with -DENABLE_IPC=OFF to disable multiprocess support\n"
```
Sjors, line 54: This wording seems fine.

ryanofsky (approved):
Code review ACK ee5e284cca563eb43ec3587d114586865db70b43. This seems like it should be a helpful change. 

The approach suggested https://github.com/bitcoin/bitcoin/pull/33290#issuecomment-3249470528 could potentially be used to avoid needing to mention the bitcoin core option here, and could be a followup

ryanofsky:
LLM linter also suggested some reasonable edits https://github.com/bitcoin-core/libmultiprocess/pull/205#issuecomment-3248612296

CMakeLists.txt
```diff
@@ -36,6 +36,25 @@ if(CapnProto_VERSION STREQUAL "0.7.0"
   ")
```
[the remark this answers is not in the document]
willcl-ark: Do you mean this part?

```
/usr/include/capnp/blob.h:176:74: error: invalid operands to binary expression ('const char *' and 'kj::StringPtr')
```

CMakeLists.txt
```diff
@@ -36,6 +36,25 @@ if(CapnProto_VERSION STREQUAL "0.7.0"
...
+if((CapnProto_VERSION VERSION_GREATER_EQUAL "0.9.0") AND
+   (CapnProto_VERSION VERSION_LESS "1.0.0") AND
+   (CMAKE_CXX_COMPILER_ID STREQUAL "Clang") AND
+   (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "16") AND
+   (CMAKE_CXX_STANDARD EQUAL 20))
+  message(FATAL_ERROR
+    "Cap'n Proto ${CapnProto_VERSION} is incompatible with Clang ${CMAKE_CXX_COMPILER_VERSION} when using C++20.\n"
+    "To resolve this issue, choose one of the following:\n"
+    "  1. Upgrade to Cap'n Proto version 1.0 or newer (recommended)\n"
+    "  2. Downgrade to Cap'n Proto version 0.8 or earlier\n"
```
fanquake, line 52: > Downgrade to Cap'n Proto version 0.8 or earlier

I think this should say `0.8.1`, so it's clear we aren't suggesting to downgrade to 0.8.0, which wont work, given the check for CVEs above. Although I think ideally, we wouldn't suggest downgrading at all.

willcl-ark:
OK, I am a bit of a cmake language noob, so apologies in advance if I've butchered this, but I've tried to share the common error message in 197aef2e2b4a4dd35ffa2bf5783ba6fe636ea48d to along with @fanquake's suggestion of not recommending downgrading (which seems sensible).

To avoid also the situation where we have two _seperate_ checks which a user could fail sequentially (but only after installing a second version of capnproto!), I have tried to merge them into a single check, with logic for each check type.

This way a user should only ever fail on this version check once, and should be shown appropriate options.

ryanofsky:
@willcl-ark I started to look at 197aef2e2b4a4dd35ffa2bf5783ba6fe636ea48d which seems like a nice approach. 

Any interested in extending it to replace https://github.com/bitcoin/bitcoin/pull/33290? It would be great to show a better error message when `find_package(CapnProto)` fails here and it looks like the new code you've written could be easily extended to do it.

I think you would just need to replace

```cmake
find_package(CapnProto 0.7 REQUIRED NO_MODULE)
```

with

```cmake
find_package(CapnProto 0.7 QUIET NO_MODULE)
if (NOT CapnProto_FOUND)
   # ... custom handling here ...
endif()
```

ryanofsky (approved):
Code review ACK ee5e284cca563eb43ec3587d114586865db70b43. New code is clean and pretty straightforward and I really like the approach because it clearly spells out the solutions instead of requiring users and developers to look for them.

Sjors:
I tested 657d80622f81ba3258780b2157a2771195ee4988 by updating the subtree in Bitcoin Core, and remove capnp:

```
CMake Error at src/ipc/libmultiprocess/CMakeLists.txt:18 (message):
  Cap'n Proto is required but was not found.

  To resolve, choose one of the following:

    - Install Cap'n Proto (version 1.0+ recommended)
    - For Bitcoin Core compilation build with -DENABLE_IPC=OFF to disable multiprocess support
```

That's quite nice.

ryanofsky (approved):
Code review ACK 657d80622f81ba3258780b2157a2771195ee4988. Thanks for these changes! I think they should very helpful and prevent unnecessary confusion.

CMakeLists.txt
```diff
@@ -13,7 +13,20 @@ endif()
...
-find_package(CapnProto 0.7 REQUIRED NO_MODULE)
+find_package(CapnProto 0.7 QUIET NO_MODULE)
+ if(NOT CapnProto_FOUND)
+   message(FATAL_ERROR
+     "Cap'n Proto is required but was not found.\n"
+     "To resolve, choose one of the following:\n"
+     "  - Install Cap'n Proto (version 1.0+ recommended)\n"
+     "  - For Bitcoin Core compilation build with -DENABLE_IPC=OFF to disable multiprocess support\n"
+   )
+ endif()
```
hebasto, line 24: mis-indentation?

hebasto (commented):
Concept ACK.

As a follow-up, all code related to handling the CapnProto package could be encapsulated in a `cmake/FindCapnProto.cmake` module.

ryanofsky:
I think I'll go ahead and merge this now, and we can iterate and improve in future PRs. There are good ideas above for fixing up an indent, changing wording, and moving code to a module. There is also the MP_SUBPROJECT_ERROR idea from https://github.com/bitcoin/bitcoin/pull/33290#issuecomment-3249470528 and possibility of extending this to detect an incompatible netbsd package from https://github.com/bitcoin-core/libmultiprocess/pull/196#issuecomment-3258581807

PR #209, opened 2025-09-09
cmake: Increase cmake policy version

ryanofsky:
Increase cmake policy version from 3.12 to 4.1 to stop using old and deprecated CMake policies in standalone builds.

Also stop overriding policy version if a cmake parent project has already set one, so parent projects are able to control which policies are enabled by default based on their own needs. Specifically, in the Bitcoin Core subtree, this change causes the libmultiprocess cmake policy version to increase from 3.12 to 3.22, which is the version Bitcoin Core uses.

This PR also adds a new `newdeps` CI job to test CMake 4.1 and the master branch of Cap'n Proto. This PR doesn't change the minimum version of cmake required to build the project, which is still 3.12.

---

This PR is an alternative to #163 which increases the policy version to 3.31 but doesn't include fixes for CI jobs, and doesn't allow the bitcoin core build to choose a lower policy version. This PR is also an alternative to #175 which sets the policy version to 3.22 like the bitcoin build, but also causes builds with earlier versions of cmake to fail.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,33 @@
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
-cmake_minimum_required(VERSION 3.12)
+# Call cmake_minimum_required() only if it was not already called, so parent
+# projects set the policy version when this project is included via
+# add_subdirectory().
+#
+# Rationale: Different projects have different practices for choosing policy
+# versions. For example, the Bitcoin Core project sets an old policy version,
+# causing CMake to use deprecated behaviors instead of new behaviors and
+# maximize compatibility with a single old version of CMake, reducing variance
+# between builds with newer CMake versions. By contrast CMake documentation
+# recommends setting the policy version to the latest supported version of
+# CMake, as an upgrading mechanism rather than a pinning mechanism, to let
+# project authors fix problems before enabling newer policies, while not opting
+# into deprecated policies on a longer term basis.
+if(NOT DEFINED CMAKE_MINIMUM_REQUIRED_VERSION
+   OR CMAKE_MINIMUM_REQUIRED_VERSION VERSION_LESS 3.12)
+  # The left number in the range below is the minimum CMake version required to
+  # run this project. The right number is the CMake policy version.
+  #
+  # The purpose of the minimum version is to trigger a helpful error if a version
+  # of CMake is being used that is too old to work. If this number is changed,
+  # the version in ci/configs/olddeps.bash should be changed to match.
+  #
+  # The purpose of the policy version is to opt out of policies introduced in
+  # newer versions of CMake until they have been tested. If this number is
+  # changed, the version in ci/configs/newdeps.bash should be changed to match.
+  cmake_minimum_required(VERSION 3.12...4.1 FATAL_ERROR)
+endif()
 
 project("Libmultiprocess" CXX)
 if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,14 @@ endif()
 
 include("cmake/compat_find.cmake")
 
-find_package(Threads REQUIRED)
+add_library(mpdeps INTERFACE)
+
+find_package(Threads)
+
+if(Threads_FOUND)
+  target_link_libraries(mpdeps INTERFACE Threads::Threads)
+endif()
+
 find_package(CapnProto 0.7 QUIET NO_MODULE)
  if(NOT CapnProto_FOUND)
    message(FATAL_ERROR
@@ -202,7 +209,7 @@ target_link_libraries(mpgen PRIVATE CapnProto::capnp)
 target_link_libraries(mpgen PRIVATE CapnProto::capnp-rpc)
 target_link_libraries(mpgen PRIVATE CapnProto::capnpc)
 target_link_libraries(mpgen PRIVATE CapnProto::kj)
-target_link_libraries(mpgen PRIVATE Threads::Threads)
+target_link_libraries(mpgen PRIVATE mpdeps)
 set_target_properties(mpgen PROPERTIES
     INSTALL_RPATH_USE_LINK_PATH TRUE)
 set_target_properties(mpgen PROPERTIES
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,11 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
   set(CMAKE_CXX_STANDARD_REQUIRED YES)
 endif()
 
+# Disable automatic C++20 module dependency scanning.
+# CMake >=3.28 tries to use `clang-scan-deps` by default, which may not
+# be installed on all platforms. We don't use named modules, so turn this off.
+set(CMAKE_CXX_SCAN_FOR_MODULES OFF)
+
 include("cmake/compat_find.cmake")
 
 add_library(mpdeps INTERFACE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,11 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
   set(CMAKE_CXX_STANDARD_REQUIRED YES)
 endif()
 
+# Disable automatic C++20 module dependency scanning.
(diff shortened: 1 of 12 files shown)


ryanofsky:
Updated 6e29c02ec5afdb86874e83aa46c93b0d88c27968 -> 233b394123fcb96c5164dc1f4b1876b8d0ac6514 ([`pr/policy.1`](https://github.com/ryanofsky/libmultiprocess/commits/pr/policy.1) -> [`pr/policy.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/policy.2), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/policy.1..pr/policy.2)) with a few fixes for CI errors

CMakeLists.txt
```diff
@@ -2,7 +2,33 @@
...
+  # run this project. The right number is the CMake policy version.
+  #
+  # The purpose of the minimum version is to trigger a helpful error if a version
+  # of CMake is being used that is too old to work. If this number is changed,
+  # the version in ci/configs/olddeps.bash should be changed to match.
+  #
+  # The purpose of the policy version is to opt out of policies introduced in
+  # newer versions of CMake until they have been tested. If this number is
+  # changed, the version in ci/configs/newdeps.bash should be changed to match.
+  cmake_minimum_required(VERSION 3.12...4.1 FATAL_ERROR)
```
hebasto, line 30: `FATAL_ERROR` will probably be [ignored](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html).

ci/scripts/ci.sh
```diff
@@ -21,17 +21,43 @@ cmake --version
...
+fi
 if ver_ge "$cmake_ver" "3.15"; then
-  cmake --build . -t "${BUILD_TARGETS[@]}" -- "${BUILD_ARGS[@]+"${BUILD_ARGS[@]}"}"
+  cmake --build . --parallel -t "${BUILD_TARGETS[@]}" -- "${BUILD_ARGS[@]+"${BUILD_ARGS[@]}"}"
 else
   # Older versions of cmake can only build one target at a time with --target,
   # and do not support -t shortcut
   for t in "${BUILD_TARGETS[@]}"; do
-    cmake --build . --target "$t" -- "${BUILD_ARGS[@]+"${BUILD_ARGS[@]}"}"
+    cmake --build . --parallel --target "$t" -- "${BUILD_ARGS[@]+"${BUILD_ARGS[@]}"}"
```
hebasto, line 60: 6ba1050bef0b5ea5df35d3623c9e101cceb82677

If the [`--parallel` option](https://cmake.org/cmake/help/latest/manual/cmake.1.html#build-a-project) is used without an argument:
> the native build tool's default number is used.

This means that the build will be efficiently parallelized with the "Ninja" generator, but not with the "Unix Makefiles" generator. However, not all scripts in `ci/configs/` pass `-G Ninja`.

ci/scripts/ci.sh
```diff
@@ -21,17 +21,43 @@ cmake --version
...
 cd "$CI_DIR"
-cmake "$src_dir" "${CMAKE_ARGS[@]+"${CMAKE_ARGS[@]}"}"
+git --no-pager log -1 || true
+cmake_args=("${CMAKE_ARGS[@]+"${CMAKE_ARGS[@]}"}")
+if ! cmake "$src_dir" "${cmake_args[@]}"; then
+  # If cmake failed, try it again with debug options.
+  # Could add --trace / --trace-expand here too but they are very verbose.
+  cmake_args+=(--debug-find --debug-output --debug-trycompile --log-level=DEBUG)
+  cmake "$src_dir" "${cmake_args[@]}" || : "cmake exited with $?"
+  cat CMakeFiles/CMakeConfigureLog.yaml || true
```
hebasto, line 50: 6ba1050bef0b5ea5df35d3623c9e101cceb82677

This log file name has been used only since CMake 3.26.

ryanofsky:
re: https://github.com/bitcoin-core/libmultiprocess/pull/209#issuecomment-3292167145

> I am not happy with this. The title of the topic says "Increase cmake policy version", which I would expect to be a one line change. Instead, it has +95/-17 changes.

Happy to change the title if you have a suggestion. This PR is only increasing the policy version and improving the CI, and I did not feel the CI changes were were mentioning since they are supporting changes, not the goal of the PR.

>     * the  length of the bash buildscript is doubled. Why isn't this written in a portable cmake script? Why is CMake preferred over handwritten Makefiles when the actual CI logic is written in bash?

I think the language question should be raised new issue if you want to discuss it. (But just to provide a little context here if it can avoid an argument about the language of a ~70 line script, I just personally find bash a lot nicer and more useful than cmake and am much familiar with it. More generally, bash is commonly used in CI, there many more developers who know bash than cmake.)

The script does increase from 37 to 67 lines here to do two things:

1. To support cloning and building unreleased versions of cap'n proto, so we can fix or report compatibility issues upstream cap'n proto earlier, and not be blindsided by new changes.
2. To provide better output when find_package fails, because by default the information cmake provides is not useful for debugging.

IMO both changes are worth making and do not have any real downsides.

> If you have to explain what you are doing, you are doing it wrong.

Interesting point. It might be helpful if you could explain practical downsides of the change. Unless you also believe that if you have to explain why a change is wrong, then it must be right?

> Putting Threads::Threads behind a non-namespaced abstraction mpdeps makes the build less robust.

Can you explain this a little more as well? Is there a practical downside here or a different approach you would recommend?

ryanofsky:
re: https://github.com/bitcoin-core/libmultiprocess/pull/209#issuecomment-3292177333

> I'm struggling to create a MRE for the issue with the `FindThreads` module on FreeBSD. The following works just fine:

The error happens when you don't have `clang-scan-deps` installed on the build machine, as seems to be the case currently with our freebsd and openbs vms. This issue is not specifically related to the `find_package(Threads REQUIRED)` call but the `find_package(Threads REQUIRED)` call makes it harder to understand what is wrong because it obscures the real error. Making the threads package optional produces a clearer error later, and I think it's just more correct anyway since modern platforms don't use separate threading libraries.

If it helps there is a failing CI run with more complete output:

https://github.com/ryanofsky/libmultiprocess/actions/runs/17622518177/job/50071319558

And the relevant cmake policy is https://cmake.org/cmake/help/latest/policy/CMP0155.html

ryanofsky:
> Make sure the `CMakeLists.txt` files are declarative and follow best practices instead of inventing new approaches and deviations.

To be sure, I agree with you it would be better if Bitcoin Core followed cmake's recommended best practices and used new policies instead of setting an really old policy version.

But bitcoin core isn't doing that, and other bitcoin core developers have given reasons why it shouldn't do that, and while I don't agree with these reasons, I think it is a reasonable judgement call, and they are just choosing a different tradeoffs than I would. This PR makes libmultiprocess compatible with either approach for choosing a policy version and stops using really old policies (older than the ones bitcoin core chooses) in any case.

ryanofsky:
> Here is a [build log](https://github.com/hebasto/github-actions/actions/runs/17734446515/job/50392701817) from GHA using FreeBSD:

Is that link right? It seems to be a 404 for me. I also don't think I understand what it indicates if find_package(Threads REQUIRED) works in another build. We already know it works in many builds. You can see some builds where it is failing
in https://github.com/bitcoin-core/libmultiprocess/pull/209#issuecomment-3271994909 and https://github.com/ryanofsky/libmultiprocess/actions/runs/17622518177/job/50071319558. Making the threads library optional improves error output, works well and doesn't have other downsides, and makes sense conceptually on modern platforms that do not use separate libraries for threading.

hebasto:
> > Here is a [build log](https://github.com/hebasto/github-actions/actions/runs/17734446515/job/50392701817) from GHA using FreeBSD:
> 
> Is that link right? It seems to be a 404 for me. I also don't think I understand what it indicates if find_package(Threads REQUIRED) works in another build.

Sorry. I've made that repo public. The link should work now.

> We already know it works in many builds. You can see some builds where it is failing in [#209 (comment)](https://github.com/bitcoin-core/libmultiprocess/pull/209#issuecomment-3271994909) and https://github.com/ryanofsky/libmultiprocess/actions/runs/17622518177/job/50071319558. Making the threads library optional improves error output, works well and doesn't have other downsides, and makes sense conceptually on modern platforms that do not use separate libraries for threading.

I want to report the issue upstream and I need an MRE for it.

hebasto:
@ryanofsky 
> Some things I might suggest:
> 
>    * Starting from my failing commit and deleting unnecessary steps to find a minimal failing case

Thank you for the suggestion.

The culprit turned out to be the "Ninja" generator, which was quite surprising to me.

ryanofsky (commented):
Thanks for the review! Note: first two commits are now part of #212

PR #210, opened 2025-09-10
Replace KJ_DEFER with kj::defer

Sjors:
My understanding is that this doesn't really matter much, e.g. tools like tidy and TSan won't work better. They might produce slightly more readable output.

It should however work for the all the versions of capnp that we support.

Based on #201


Sjors:
> It should however work for the all the versions of capnp that we support.

I mainly opened this PR to test this assumption.

Feel free to close if it otherwise seems useless.

Sjors:
CI passed, but closing this for now.

PR #212, opened 2025-09-17
ci: add newdeps job testing newer versions of cmake and capnproto

ryanofsky:
Add CI job testing latest released version of cmake, and latest unreleased version of cap'n proto to uncover any problems there may be with new releases. Use git version of cap'n proto because we've submitted patches to that project to fix issues in the past, so it would be good to know about any new issues as soon as possible.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
         uses: vmactions/freebsd-vm@v1
         with:
           prepare: |
-            pkg install -y cmake ninja bash capnproto
+            pkg install -y cmake ninja bash capnproto git
           sync: 'rsync'
           copyback: false
 
@@ -79,7 +79,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        config: [default, llvm, gnu32, sanitize, olddeps]
+        config: [default, llvm, gnu32, sanitize, olddeps, newdeps]
 
     name: build • ${{ matrix.config }}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,7 +106,7 @@ jobs:
         uses: vmactions/freebsd-vm@v1
         with:
           prepare: |
-            pkg install -y cmake ninja bash capnproto
+            pkg install -y cmake ninja bash capnproto git
           sync: 'rsync'
           copyback: false
 
@@ -148,7 +148,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        config: [default, llvm, gnu32, sanitize, olddeps]
+        config: [default, llvm, gnu32, sanitize, olddeps, newdeps]
 
     name: build • ${{ matrix.config }}
 
--- a/ci/configs/newdeps.bash
+++ b/ci/configs/newdeps.bash
@@ -1,0 +1,6 @@
+CI_DESC="CI job using newest Cap'n Proto and cmake versions"
+CI_DIR=build-newdeps
+export CXXFLAGS="-Werror -Wall -Wextra -Wpedantic -Wno-unused-parameter -Wno-error=array-bounds"
+CAPNP_CHECKOUT=master
+NIX_ARGS=(--argstr capnprotoVersion "none" --argstr cmakeVersion "4.1.1")
+BUILD_ARGS=(-k)
--- a/ci/scripts/ci.sh
+++ b/ci/scripts/ci.sh
@@ -21,9 +21,26 @@ cmake --version
 cmake_ver=$(cmake --version | awk '/version/{print $3; exit}')
 ver_ge() { [ "$(printf '%s\n' "$2" "$1" | sort -V | head -n1)" = "$2" ]; }
 
+# If CAPNP_CHECKOUT was requested, clone and install requested Cap'n Proto branch or tag
+capnp_prefix=
+if [ -n "${CAPNP_CHECKOUT-}" ]; then
+  capnp_prefix="$PWD/capnp-install"
+  [ -e "capnp" ] || git clone -b "${CAPNP_CHECKOUT}" "https://github.com/capnproto/capnproto" capnp
+  mkdir -p capnp/build
+  (
+    cd capnp/build
+    git --no-pager log -1 || true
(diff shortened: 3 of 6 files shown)


ryanofsky:
These changes were originally part of #209 but moved them to a separate PR since #209 might be blocked or require more time to discuss

ryanofsky (commented):
Appreciate your review, even though I am having trouble understanding what your practical concerns are and what specific changes I could make that would address them.

Hopefully we both accept the premise that it is good to have a new CI job testing newer versions of cmake and capnproto, and it is good to show more debug output when cmake configuration fails.

ci/scripts/ci.sh
```diff
@@ -21,9 +21,26 @@ cmake --version
...
+    cmake --build .
+    cmake --install .
+  )
+  export CMAKE_PREFIX_PATH="${capnp_prefix}:${CMAKE_PREFIX_PATH-}"
+fi
+
 src_dir=$PWD
 mkdir -p "$CI_DIR"
 cd "$CI_DIR"
+git --no-pager log -1 || true
```
maflcko, line 44: is there a ci config that does not have `git`? I can't see one and it seems easy enough to just require it, so that logs always have the commit embedded?

fdtwd8vv45-sketch (approved):

ryanofsky (commented):
Rebased 6ba1050bef0b5ea5df35d3623c9e101cceb82677 -> 03b147e5cbb6b8b4d362f57304a29a14031002e5 ([`pr/newdeps.1`](https://github.com/ryanofsky/libmultiprocess/commits/pr/newdeps.1) -> [`pr/newdeps.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/newdeps.2), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/newdeps.1-rebase..pr/newdeps.2)) due to conflict with #334

Updated 03b147e5cbb6b8b4d362f57304a29a14031002e5 -> b3fc922ee4a663811e0048556b788b4ca473d781 ([`pr/newdeps.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/newdeps.2) -> [`pr/newdeps.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/newdeps.3), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/newdeps.2..pr/newdeps.3)) dropping debug commit 03b147e5cbb6b8b4d362f57304a29a14031002e5

ci/configs/newdeps.bash
```diff
@@ -0,0 +1,6 @@
+CI_DESC="CI job using newest Cap'n Proto and cmake versions"
+CI_DIR=build-newdeps
+export CXXFLAGS="-Werror -Wall -Wextra -Wpedantic -Wno-unused-parameter -Wno-error=array-bounds"
+CAPNP_CHECKOUT=master
```
maflcko, line 4:
```suggestion
CAPNP_CHECKOUT=master  # This is the v1.x release branch
```

The default branch is v2, maybe clarify what master means?

maflcko (commented):
lgtm ACK b3fc922ee4a663811e0048556b788b4ca473d781

hebasto (approved):
ACK b3fc922ee4a663811e0048556b788b4ca473d781.

> Add CI job testing latest released version of cmake...

Could be updated as follows:
```diff
--- a/ci/configs/newdeps.bash
+++ b/ci/configs/newdeps.bash
@@ -2,5 +2,5 @@ CI_DESC="CI job using newest Cap'n Proto and cmake versions"
 CI_DIR=build-newdeps
 export CXXFLAGS="-Werror -Wall -Wextra -Wpedantic -Wno-unused-parameter -Wno-error=array-bounds"
 CAPNP_CHECKOUT=master
-NIX_ARGS=(--argstr capnprotoVersion "none" --argstr cmakeVersion "4.1.1")
+NIX_ARGS=(--argstr capnprotoVersion "none" --argstr cmakeVersion "4.3.4")
 BUILD_ARGS=(-k)
--- a/shell.nix
+++ b/shell.nix
@@ -55,7 +55,7 @@ let
   clang-tools = llvm.clang-tools.override { inherit enableLibcxx; };
   cmakeHashes = {
     "3.12.4" = "sha256-UlVYS/0EPrcXViz/iULUcvHA5GecSUHYS6raqbKOMZQ=";
-    "4.1.1" = "sha256-sp9vGXM6oiS3djUHoQikJ+1Ixojh+vIrKcROHDBUkoI=";
+    "4.3.4" = "sha256-/e/4l7nrSddkU58rHtxut+FEDfMlZ4qXwZeEmekxrdo=";
   };
   gcc = if gccVersion == null then null else builtins.getAttr ("gcc" + gccVersion) pkgs;
   cmakeBuild = if cmakeVersion == null then pkgs.cmake else (pkgs.cmake.overrideAttrs (old: {
```

.github/workflows/ci.yml
```diff
@@ -148,7 +148,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        config: [default, llvm, gnu32, sanitize, olddeps]
+        config: [default, llvm, gnu32, sanitize, olddeps, newdeps]
```
ryanofsky: >

ci/scripts/ci.sh
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/212#discussion_r2367210874

> is there a ci config that does not have `git`? I can't see one and it seems easy enough to just require it, so that logs always have the commit embedded?

Sounds like the request here is to remove `|| true`?

I'd slightly prefer to keep it just because these CI scripts are useful for testing locally even outside context of a real CI system, and there are reasons it can be useful to copy files around and not have valid git checkouts during testing. But it is true that a drawback of `|| true` is that it could allow someone to add a CI job that does not have print the commit hash here. So I'd be ok with a change dropping `|| true` in my planned followup, just let me know.

ci/configs/newdeps.bash
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/212#discussion_r2367210874

> The default branch is v2, maybe clarify what master means?

Good suggestion, have added in followup

ryanofsky (commented):
Thanks for the reviews! Am planning to merge as-is since this is reviewed and passes ci, but will make a followup bumping the cmake version and implementing other suggestions

PR #213, opened 2025-09-18
util+doc: Clearer errors when attempting to run examples + polished docs

hodlinator:
Found while reviewing #201.


ryanofsky (approved):
Code review ACK 91606b232b4f62efcad4795f7b48cbfea60e2276. Thanks for the PR! Left some suggestions below, but overall I think this is an improvement.

> Let me know if you see any merit to the Drahtbot recommendation ([#213 (comment)](https://github.com/bitcoin-core/libmultiprocess/pull/213#issuecomment-3307125359))

Not too important, I think it'd make sense to replace "Its return values" with "The return values." Currently it seems like "its" refers to `capnp::RpcSystem` which doesn't really make sense.

theuni (approved):
utACK commits/ec03a9639ab55de85df4b688c08b1d7faec01c67

Issue #206, opened 2025-09-05
type-context.h should return error instead of asserting on simultaneous IPC calls


ryanofsky:
Currently, when a client makes simultaneous IPC calls on a single server thread (assigning the same [Context.thread](https://github.com/bitcoin-core/libmultiprocess/blob/878e84dc3030fb9a545c3f2ec6b4aacfdac834c8/include/mp/proxy.capnp#L59-L60) value in the request), [`assert(!m_fn)`](https://github.com/bitcoin-core/libmultiprocess/blob/878e84dc3030fb9a545c3f2ec6b4aacfdac834c8/include/mp/proxy-io.h#L279) triggers in `Waiter::post`.

Since this error is pretty easy to trigger in python and rust cilents not using libmultiprocess, it would be better for the server to refuse these requests and provide a clear error message instead of crashing. This should just require a small change to the `Waiter` class and the code calling it in [`type-context.h`](https://github.com/bitcoin-core/libmultiprocess/blob/878e84dc3030fb9a545c3f2ec6b4aacfdac834c8/include/mp/type-context.h#L145). Context: https://github.com/bitcoin/bitcoin/pull/33201#issuecomment-3258111873 and https://gnusha.org/bitcoin-core-dev/2025-09-05.log:

```
04:50 < sipa> ryanofsky: any idea about this error in https://github.com/bitcoin/bitcoin/actions/runs/17491932432/job/49683489249?pr=33201 :
04:51 < sipa>  node0 stderr bitcoin-node: ipc/libmultiprocess/include/mp/proxy-io.h:279: void mp::Waiter::post(Fn &&) [Fn = (lambda at ./ipc/libmultiprocess/include/mp/type-context.h:67:19)]: Assertion `!m_fn' failed. 
04:55 < kevkevin> looks like the test got hung up on this line in the ipc test: template4 = File "/home/admin/actions-runner/_work/_temp/build/test/functional/interface_ipc.py", line 156, in async_routine: await template2.result.waitNext(ctx, waitoptions)
04:56 < sipa> yeah, bitcoin-node crashed during that invocation
05:01 < sipa> i can reproduce it locally, though it's pretty rare (once in 60 runs, so far)
05:04 < kevkevin> hmm interesting, I'll try to reproduce it locally myself
05:05 < ryanofsky> that is interesting, i've never seen that assert hit before. i think it could happen if two ipc clients try to execute a different IPC calls on the same thread at the same time
05:06 < sipa> huh!
[...]

PR #214, opened 2025-09-23
Fix crash on simultaneous IPC calls using the same thread

Eunovo:
Relevant Issue: https://github.com/bitcoin-core/libmultiprocess/issues/206

This error occurs when non-C++ clients make simultaneous IPC calls to the server using the same Server thread.  
The libmultiprocess C++ client ensures a 1-to-1 mapping of client-to-server threads, so simultaneous calls using the same thread cannot be made.

---

### **Testing**
I have added a unit test in the last commit.
You can cherry-pick this commit on master to reproduce the crash.

### DIFF
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -281,12 +281,13 @@ struct Waiter
     Waiter() = default;
 
     template <typename Fn>
-    void post(Fn&& fn)
+    bool post(Fn&& fn)
     {
         const Lock lock(m_mutex);
-        assert(!m_fn);
+        if (m_fn) return false;
         m_fn = std::forward<Fn>(fn);
         m_cv.notify_all();
+        return true;
     }
 
     template <class Predicate>
--- a/include/mp/type-context.h
+++ b/include/mp/type-context.h
@@ -152,7 +152,12 @@ auto PassField(Priority<1>, TypeList<>, ServerContext& server_context, const Fn&
                 const auto& thread = static_cast<ProxyServer<Thread>&>(*thread_server);
                 server.m_context.loop->log()
                     << "IPC server post request  #" << req << " {" << thread.m_thread_context.thread_name << "}";
-                thread.m_thread_context.waiter->post(std::move(invoke));
+                if (!thread.m_thread_context.waiter->post(std::move(invoke))) {
+                    server.m_context.loop->log()
+                        << "IPC server error request #" << req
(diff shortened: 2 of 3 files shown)


Eunovo:
> Nice fix! And the python test looks like it could be a nice addition: [master...Eunovo:libmultiprocess:test-ipc-with-py](https://github.com/bitcoin-core/libmultiprocess/compare/master...Eunovo:libmultiprocess:test-ipc-with-py)

I wasn't sure it was worth it to add the Python test. It could be useful for reproducing scenarios experienced by non-libmultiprocess clients in general, but I'm not convinced.

> I do think it should be pretty straightforward to write a c++ unit test too, not sure if you were planning on doing that or if something was in the way.

I think it might be possible if I use the Capnp library directly instead of the libmultiprocess client.

ryanofsky:
Code review 21886f2bbdf7238282909f04c8c0d57ec4f75726. Thanks for the new test!

The approach in the test seems pretty clever. In order to trigger the "thread busy" error, you need the server IPC thread to be busy waiting for something when a new request comes in. By having the server run the "callbackSaved" method, it guarantees the thread will be busy because it will be stuck trying to call the client while the event loop is occupied.

I think a more direct approach is probably possible, maybe by adding new FooInterface `wait` and `stopWaiting` methods where `wait` just blocks indefinitely until `stopWaiting` is called, and the "thread busy" error can be called by just making 2 `wait` requests simultaneously. Or instead of adding any new methods, the existing `callFnAsync` method can be called instead of `wait()` and `m_fn` could just be set in the beginning of the test to wait indefinitely until the end of the test.

I do like current approach though and there's no need to change it.

Eunovo:
@ryanofsky I was inspired by your comment to try a different approach in https://github.com/bitcoin-core/libmultiprocess/pull/214/commits/1238170f68e8fa7ae41c79465df5cdae34d568e9 to make the test more reliable by setting `m_fn` to wait on a `promise`. It worked, but it still needs 3 simultaneous calls. I expected this approach to work with just 2 calls, but for some reason, `m_fn` is unset when the second call posts to the waiter.

ryanofsky:
That's interesting. 1238170f68e8fa7ae41c79465df5cdae34d568e9 looks like a straightforward test for the problem, and I would also expect only 2 simultaneous calls to necessary not 3.

Code review ACK 1238170f68e8fa7ae41c79465df5cdae34d568e9, since the code changes here all make sense except the initial `running = 3` value, and there is a comment about that, so at least it's called out and may be something that can be improved later.

PR #216, opened 2025-09-23
Add a verbose logging option and put expensive debug output behind it

### LINKED ISSUE #215: Substantial overhead caused by encoding messages for logging
I'm working on a multiprocess branch that passes a lot of data back and forth between server and client.

When looking at a flamegraph, it's clear that >50% of my client's cpu time is spent on logging (encoding) sends. Likewise for the server with responses.

The culprit is `toString()` from the `params`/`results`/`request`/`response`, which ends up in `kj::_::encodeCEscapeImpl()`, which is _horribly_ slow.

Current master:
Image
Fixed:
Image

As you can see, before removing the logging, 52% of my thread's runtime is spent in `serverInvoke` (which calls `foo.toString()`, which calls `structString()`, which calls `kj::_::encodeCEscapeImpl()`. With the debug lines removed, `serverInvoke` is no longer visible in the flamegraph at all.

I would imagine the sv2 implementation suffers from this as well, though I'm not sure how chatty it is, so I can't speak to the potential impact.

~I'll open a RFC PR which fixes the issue rather naively. A more robust solution would introduce logging categories similar to Bitcoin Core's.~

Edit: See https://github.com/bitcoin-core/libmultiprocess/pull/216.
Also, toned down the original title :)

theuni:
Fixes #215. It's rather naive, but demonstrates a fix for the problem.

Logging every send/receive/request/response is very expensive for structs due to the way they're stringified by capnproto.

Introduce the concept of verbose logging, and expose it to clients by adding a new constructor for EventLoop which accepts LogOptions.

For backwards compatibility, the existing constructor is unchanged. Additionally for back compat, a new callback function is added rather than adding a parameter to the existing one.

The usage is somewhat awkward due to the fact that the data should only be stringified if verbose logging has been requested. This could be less awkward by changing the logger to take a callback which is only called if verbosity is enabled instead, that way the slow serialization only happens in that case. Additionally, a more complete solution would introduce actual log levels rather than a binary option

But those suggestions would be rather invasive and presumably out of scope for simply fixing the linked issue.

### DIFF
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -136,6 +136,9 @@ struct LogOptions {
     //! Maximum number of characters to use when representing
     //! request and response structs as strings.
     size_t max_chars{200};
+
+    //! Enable verbose logging which may affect performance
+    bool verbose{false};
 };
 
 std::string LongThreadName(const char* exe_name);
@@ -168,8 +171,13 @@ std::string LongThreadName(const char* exe_name);
 class EventLoop
 {
 public:
-    //! Construct event loop object.
-    EventLoop(const char* exe_name, LogFn log_fn, void* context = nullptr);
+    //! Construct event loop object with default logging options.
+    EventLoop(const char* exe_name, LogFn log_fn, void* context = nullptr)
+        : EventLoop(exe_name, LogOptions{std::move(log_fn)}, context){}
+
+    //! Construct event loop object with specified logging options.
+    EventLoop(const char* exe_name, LogOptions log_opts, void* context = nullptr);
+
     ~EventLoop();
 
     //! Run event loop. Does not return until shutdown. This should only be
--- a/src/mp/proxy.cpp
+++ b/src/mp/proxy.cpp
@@ -191,13 +191,13 @@ void EventLoop::addAsyncCleanup(std::function<void()> fn)
     startAsyncThread();
 }
 
-EventLoop::EventLoop(const char* exe_name, LogFn log_fn, void* context)
+EventLoop::EventLoop(const char* exe_name, LogOptions log_opts, void* context)
     : m_exe_name(exe_name),
       m_io_context(kj::setupAsyncIo()),
       m_task_set(new kj::TaskSet(m_error_handler)),
+      m_log_opts(std::move(log_opts)),
       m_context(context)
 {
-    m_log_opts.log_fn = log_fn;
     int fds[2];
     KJ_SYSCALL(socketpair(AF_UNIX, SOCK_STREAM, 0, fds));
     m_wait_fd = fds[0];
--- a/include/mp/proxy-types.h
+++ b/include/mp/proxy-types.h
@@ -629,15 +629,19 @@ void clientInvoke(ProxyClient& proxy_client, const GetRequest& get_request, Fiel
         using FieldList = typename ProxyClientMethodTraits<typename Request::Params>::Fields;
         invoke_context.emplace(*proxy_client.m_context.connection, thread_context);
         IterateFields().handleChain(*invoke_context, request, FieldList(), typename FieldObjs::BuildParams{&fields}...);
-        proxy_client.m_context.loop->logPlain()
-            << "{" << thread_context.thread_name << "} IPC client send "
-            << TypeName<typename Request::Params>() << " " << LogEscape(request.toString(), proxy_client.m_context.loop->m_log_opts.max_chars);
+        if (proxy_client.m_context.loop->m_log_opts.verbose) {
+                proxy_client.m_context.loop->logPlain()
+                    << "{" << thread_context.thread_name << "} IPC client send "
+                    << TypeName<typename Request::Params>() << " " << LogEscape(request.toString(), proxy_client.m_context.loop->m_log_opts.max_chars);
+        }
 
         proxy_client.m_context.loop->m_task_set->add(request.send().then(
             [&](::capnp::Response<typename Request::Results>&& response) {
-                proxy_client.m_context.loop->logPlain()
-                    << "{" << thread_context.thread_name << "} IPC client recv "
-                    << TypeName<typename Request::Results>() << " " << LogEscape(response.toString(), proxy_client.m_context.loop->m_log_opts.max_chars);
+                if (proxy_client.m_context.loop->m_log_opts.verbose) {
+                    proxy_client.m_context.loop->logPlain()
+                        << "{" << thread_context.thread_name << "} IPC client recv "
+                        << TypeName<typename Request::Results>() << " " << LogEscape(response.toString(), proxy_client.m_context.loop->m_log_opts.max_chars);
+                }
                 try {
                     IterateFields().handleChain(
                         *invoke_context, response, FieldList(), typename FieldObjs::ReadResults{&fields}...);
@@ -700,8 +704,10 @@ kj::Promise<void> serverInvoke(Server& server, CallContext& call_context, Fn fn)
     using Results = typename decltype(call_context.getResults())::Builds;
 
     int req = ++server_reqs;
-    server.m_context.loop->log() << "IPC server recv request  #" << req << " "
-                                     << TypeName<typename Params::Reads>() << " " << LogEscape(params.toString(), server.m_context.loop->m_log_opts.max_chars);
+    if (server.m_context.loop->m_log_opts.verbose) {
+        server.m_context.loop->log() << "IPC server recv request  #" << req << " "
+                                         << TypeName<typename Params::Reads>() << " " << LogEscape(params.toString(), server.m_context.loop->m_log_opts.max_chars);
+    }
 
     try {
         using ServerContext = ServerInvokeContext<Server, CallContext>;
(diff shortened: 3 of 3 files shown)


ryanofsky:
Code review ACK e50802f31e3cfc12f9447ea9dcb951e0123788d0. I think this is a good fix, and the backwards compatibility is helpful for avoiding version mismatch errors during development, and for avoiding the need to modify bitcoin core code in a PR bumping the ilbmultiprocess subtree (which should make it a little easier to review).

I think this PR could be simplified a little by dropping new `LogFn verbose_log_fn` struct member and just using the existing log callback for logging everything. This would also allow dropping new `bool verbose=false` logging arguments.

Longer run, it might make sense to introduce a macro to make logging less cumbersome like:

```c++
MP_LOG(loop, level) << "IPC client recv" << request;
```

where `MP_LOG` macro just uses an `mp::Logger` class:

```c++
#define MP_LOG(...) if (mp::Logger logger{__VA_ARGS__}; logger) logger
```

and `mp::Logger` class can:

- Decide whether the rest of the log statement is evaluated by providing `operator bool`
- Control log formatting by providing `operator <<`
- Accept log level levels or other context by providing different constructors

theuni:
> I think this PR could be simplified a little by dropping new `LogFn verbose_log_fn` struct member and just using the existing log callback for logging everything. This would also allow dropping new `bool verbose=false` logging arguments.

This would be my preference as well, but that leaves the log function unable to tell whether this is a "verbose" message or not. In Core's case, we'd want treat that differently (as shown here: https://github.com/bitcoin-core/libmultiprocess/pull/216#issuecomment-3325553341). In order to tell them apart, we'd need to add a `bool verbose` param to the log function, which I was hesitant to do for back compat reasons.

> Longer run, it might make sense to introduce a macro to make logging less cumbersome like:
> 
> ```c++
> MP_LOG(loop, level) << "IPC client recv" << request;
> ```

I think we'd want to take that one step further and do what Core does: encapsulating the arguments in the macro as well to [avoid their evaluation at unrequested log levels](https://github.com/bitcoin/bitcoin/blob/master/src/logging.h#L363).

But since the codebase is still rather young and flexible, my suggestion would be a callback. Something like:

```c++

enum class LogLevel {
    Trace = 0,
    Debug,
    Info,
    Warning,
    Error,
};

...

template <typename Callable>
void log(LogLevel log_level, Callable&& log_callback)
{
    logPlain(log_level, [this, &log_callback](auto& logger) {
        logger << "{" << LongThreadName(m_exe_name) << "} ";
        std::invoke(log_callback, logger);
    });
}

template <typename Callable>
void logPlain(LogLevel log_level, Callable&& log_callback)
{
    if (log_level >= m_log_opts.log_level) {
        Logger logger(false, m_log_opts.log_fn);
        std::invoke(log_callback, logger);
    }
}
```

With macros like:
```c++
#define MP_LOGINFO(loop, logger, code) loop.log(LogLevel::Info, [&](auto& logger) { code; })
#define MP_LOGDEBUG(loop, logger, code) loop.log(LogLevel::Debug, [&](auto& logger) { code; })
```

and usage like:
```c++
MP_LOGDEBUG(loop, logger, logger << "foo" << 3; logger << bar.makeExpensiveString());
```

That way, in the example above, `makeExpensiveString()` won't be called unless debugging is enabled.

theuni:
Hmm, maybe I read too quickly.

Looking again, your approach is clever and I believe keeps the arguments unevaluated as necessary as well?

ismaelsadeeq (commented):
nice Concept ACK

theuni:
> Hmm, maybe I read too quickly.
> 
> Looking again, your approach is clever and I believe keeps the arguments unevaluated as necessary as well?

Ok, yeah, neat. Let's do that :)

But I think that's probably out of scope for fixing #215 for v30.

Another quick fix would be to simply remove the expensive logging calls for now (or hide them behind an ifdef) and re-add them along with a more comprehensive logging refactor. Since they're only useful for developing/debugging anyway, that would be perfectly reasonable imo.

ryanofsky:
Thanks I do like this PR as a simple fix and I think it would be good to merge. I'd just suggest simplifying it a little with:

<p>

```diff
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -133,9 +133,6 @@ struct LogOptions {
     //! External logging callback.
     LogFn log_fn;
 
-    //! External verbose logging callback.
-    LogFn verbose_log_fn{nullptr};
-
     //! Maximum number of characters to use when representing
     //! request and response structs as strings.
     size_t max_chars{200};
@@ -220,13 +217,13 @@ public:
     //! Check if loop should exit.
     bool done() const MP_REQUIRES(m_mutex);
 
-    Logger log(bool verbose = false)
+    Logger log()
     {
-        Logger logger(false, verbose ? m_log_opts.verbose_log_fn : m_log_opts.log_fn);
+        Logger logger(false, m_log_opts.log_fn);
         logger << "{" << LongThreadName(m_exe_name) << "} ";
         return logger;
     }
-    Logger logPlain(bool verbose = false) { return {false, verbose ? m_log_opts.verbose_log_fn : m_log_opts.log_fn}; }
+    Logger logPlain() { return {false, m_log_opts.log_fn}; }
     Logger raise() { return {true, m_log_opts.log_fn}; }
 
     //! Process name included in thread names so combined debug output from
```
</p>
</details>

But not very important, so feel free to keep `verbose_log_fn` if useful.

theuni:
Ok. That means that Core can't make the distinction between verbose and non-verbose messages.

For the sake of this issue, that's fine. And assuming libmp eventually gains log levels, we'll need an API break anyway.

Will push that change. But imo it makes sense to revisit logging and change the api asap before there are too many downstreams.

include/mp/proxy-io.h
```diff
@@ -136,6 +136,9 @@ struct LogOptions {
     //! Maximum number of characters to use when representing
     //! request and response structs as strings.
     size_t max_chars{200};
+
+    //! Enable verbose logging which may affect performance
+    bool verbose{false};
```
ryanofsky, line 141: In commit "Add a verbose logging option and put expensive debug output behind it" (8f9baf7):

Setting default to false is ok, but would be good to update commit message to make clear this PR turns off most IPC logging by default. E.g. when merged into bitcoin core this will make the -debug=ipc option less useful. Could leave the default to true to be conservative. But I'm also fine with keeping the default false since probably -debug=ipc should be less spammy and verbose logs should require -loglevel=ipc:trace

src/mp/proxy.cpp
```diff
@@ -191,13 +191,13 @@ void EventLoop::addAsyncCleanup(std::function<void()> fn)
     startAsyncThread();
 }
 
-EventLoop::EventLoop(const char* exe_name, LogFn log_fn, void* context)
+EventLoop::EventLoop(const char* exe_name, LogOptions log_opts, void* context)
     : m_exe_name(exe_name),
       m_io_context(kj::setupAsyncIo()),
       m_task_set(new kj::TaskSet(m_error_handler)),
+      m_log_opts{log_opts},
```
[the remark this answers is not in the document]
theuni: Hah, I had them originally but got rid of them since they're currently trivial. Will do.

theuni:
> Maybe it can be as simple as declaring a LogLevel enum to replace `bool verbose` option, and passing the LogLevel value as an argument to the logging callback function.

That's exactly what I had in mind as well.

include/mp/proxy-io.h
theuni: I'd rather just change the default if you're ok with that. It's so expensive (for my application, anyway) that it's harmful to the default experience. When that's the case, I think it's reasonable to require the user to change the default.

For Core, I can PR a change to add the logging either to debug or trace, whichever is your preference. The thing that matters most to me is that the super-slow stringification does not happen when logging is not even enabled, as is the case currently.

theuni:
Addressed @ryanofsky's feedback.

include/mp/proxy-io.h
theuni: Sorry to keep repeating myself, I just want to be clear about my concern and make sure we're on the same page.

Whether or not Core or other users of the library show the individual function calls in their debug/trace logs is not a big issue imo -- if someone has logging on, they would expect performance to suffer for it.

What's unexpected, and my primary concern here, is that the user pays the _very_ heavy cost of the stringification even if logging is disabled (either via a null callback, or in Core's case, by not setting -debug=ipc). That majorly violates the principle of least surprise.

Indeed this is a stopgap fix. A more robust change to add the log levels as you proposed above would also fix my concern as a side-effect, as with that you'd only pay for the logging you use. I think that would also address your onboarding concern because trace logging could be the default, and setting a quieter log level would remove the overhead.

ismaelsadeeq (commented):
It seems to me this does not effectively solve the expensive debug logs but rather hides the log at all.
I try running the examples binary and can't see any logs at all.
For anyone who uses this, they will effectively have empty IPC logs.

If I understand correccly verbose debug is supposed to have more information, which should perhaps increase the `max_structs` to some big value rather than turning off logs.

For verbose users (In this context users that want to see logs which is almost all the projects experimenting with the IPC interface the log might help in detecting issues e.g https://github.com/bitcoin/bitcoin/issues/33463#issuecomment-3328278574), wont see this if this is pulled to bitcoin core and we did not make -debug option or some log level add the verbose option they also won't see logs at all, and if a log level is added the slow down is their anyway.

I would prefer if we had something like the approach suggested at https://github.com/bitcoin-core/libmultiprocess/pull/216#issuecomment-3328200613

theuni:
> It seems to me this does not effectively solve the expensive debug logs but rather hides the log at all. I try running the examples binary and can't see any logs at all. ~For anyone who uses this, they will effectively have empty IPC logs~ They can turn it on by passing verbose as true.

I have a local branch which implements @ryanofsky's suggested changes above. It's more invasive and would require a change to Core to update the logging callback, but it is more robust. I will PR it on Monday.

TheCharlatan:
Concept ACK

By the reported amount of latency eliminated by this change, I think this should be backported if possible. Or at least for a mining use case, that magnitude of latency for just logging seems unacceptable. Maybe I missed something in the conversation here, but surely it should be possible to hookup whatever severity variable is introduced here to the severity levels in core.

PR #218, opened 2025-09-24
Better error and log messages

ryanofsky:
This PR contains some logging and error message improvements I made locally in recent months that seem useful to upstream. Details are in commit messages


ismaelsadeeq:
Concept ACK

include/mp/util.h
```diff
@@ -238,6 +244,24 @@ inline char* CharCast(unsigned char* c) { return (char*)c; }
 inline const char* CharCast(const char* c) { return c; }
 inline const char* CharCast(const unsigned char* c) { return (const char*)c; }
 
+#if defined(__GNUG__)          // GCC & Clang ─ use <cxxabi.h> to demangle
```
theuni, line 247: I don't love this without an escape-hatch for more esoteric compilers. Something like `#if defined(__GNUG__) && !defined(MP_NO_DEMANGLE)` would make it more bulletproof.

That can always be added if it ever becomes a real problem though.

theuni (commented):
code review ACK 0ce75e3162bcc605757cd36cb3b90541d7a13a88.

The logging additions look helpful, though they'll need to be updated to use the new macros from #220.

I'd rather hold off on that second commit until after v30 has settled though, since it introduces some platform-specific functionality that has the (admittedly small) potential for causing build problems.

ryanofsky:
Updated a5b8a2a3293a9650feab1101afa35a1d47981049 -> 0ce75e3162bcc605757cd36cb3b90541d7a13a88 ([`pr/errs.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/errs.2) -> [`pr/errs.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/errs.3), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/errs.2..pr/errs.3)) rebasing to fix conflict with #220 and applying suggestions. Also renamed `TypeName` to `CxxTypename` since there is actually already another `TypeName` function for cap'nproto types

ryanofsky:
gnu32 CI job is failing here but I think it is not directly related to changes in this PR. The new thread busy test introduced in #214 is segfaulting here but it has a known race condition fixed in #222 . The segfault is consistent with the race and might just be happening now because new log prints in this PR change the timing of the test, and fact that bugs tend to trigger more segfaults in the gnu32 job because of the smaller memory space.

Will cherry-pick the race fix from #222 as a workaround to see if it fixes the segfault. Assuming this works, if reviewing this PR, would be nice to just ACK first two commits (2af6a9bc3036895ddddbd80d1f7789bec92a6516) and ignore the workaround so it can be dropped later without needing to rebase or re-review.

include/mp/proxy-types.h
```diff
@@ -726,6 +726,13 @@ kj::Promise<void> serverInvoke(Server& server, CallContext& call_context, Fn fn)
                 MP_LOG(*server.m_context.loop, Log::Debug) << "IPC server send response #" << req << " " << TypeName<Results>();
                 MP_LOG(*server.m_context.loop, Log::Trace) << "response data: "
                     << LogEscape(call_context.getResults().toString(), server.m_context.loop->m_log_opts.max_chars);
+            }, [&server, req](::kj::Exception&& e) {
```
theuni, line 729: isn't the local `server_capture` being captured by reference problematic if `invoke` is being called async? Possible that's smashing the stack here?

(Maybe I'm way off, these tangled callbacks a breaking my brain).

But I think I'd expect this to be `return ReplaceVoid([server_context = std::move(server_context)]()...`

ryanofsky:
> Since you mentioned this in [bitcoin/bitcoin#34568](https://github.com/bitcoin/bitcoin/pull/34568), I should probably review this. Can you give it a rebase?

I think this PR effectively depends on #240, because there is a problem with the "thread busy" exception test in gnu32 CI job that I can't figure out or reliably reproduce, and #240 rewrites thread busy behavior to avoid the exceptions that seem to trigger the problem.

I could rebase this on #240, but if you have time to review #240 it would be very helpful and I am happy to explain anything there.

ryanofsky:
Note: There's probably not much urgency to reviewing this PR since #240 fixed previous issues with uncaught exceptions and this PR doesn't actually affect them (https://github.com/bitcoin/bitcoin/pull/34568#discussion_r2844118243). The main benefit of this PR is clearer object lifetime logging so it easier to figure out when objects are being created and destroyed. And there are also better log messages in two more narrow error cases.

sedited (commented):
The first commit makes sense to me, but I left a question about the second one.

ryanofsky (commented):
Rebased e62b8f681cc386f6403353044443e1768b062437 -> 4b07ec29e0d1567f1de31aedc559fe3dc23d2a1d ([`pr/errs.9`](https://github.com/ryanofsky/libmultiprocess/commits/pr/errs.9) -> [`pr/errs.10`](https://github.com/ryanofsky/libmultiprocess/commits/pr/errs.10), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/errs.9-rebase..pr/errs.10)) due to conflicts with #258 and #249

Updated 4b07ec29e0d1567f1de31aedc559fe3dc23d2a1d -> 9fe94e999e95c05f75f9349cf49b85aefa0f4b5c ([`pr/errs.10`](https://github.com/ryanofsky/libmultiprocess/commits/pr/errs.10) -> [`pr/errs.11`](https://github.com/ryanofsky/libmultiprocess/commits/pr/errs.11), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/errs.10..pr/errs.11)) simplifying the first commit

Updated 9fe94e999e95c05f75f9349cf49b85aefa0f4b5c -> 4eae445d6d8a17e844bde85ddc96ec389c275197 ([`pr/errs.11`](https://github.com/ryanofsky/libmultiprocess/commits/pr/errs.11) -> [`pr/errs.12`](https://github.com/ryanofsky/libmultiprocess/commits/pr/errs.12), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/errs.11..pr/errs.12)) fixing CI warning about unused capture https://github.com/bitcoin-core/libmultiprocess/actions/runs/23914061887/job/69743242246?pr=218

sedited (approved):
ACK 4eae445d6d8a17e844bde85ddc96ec389c275197

Issue #198, opened 2025-08-28
How to generate Test Coverage


Eunovo:
I wanted to see where testing could be improved, so I thought it would be a good idea to see the test coverage for the project. I tried my hand at setting up test coverage in https://github.com/Eunovo/libmultiprocess/tree/make-test-coverage

The process to generate the test coverage is:
```bash
cmake -B build -DCMAKE_BUILD_TYPE=Coverage
cd build
make tests
cmake -DLCOV_OPTS="--ignore-errors mismatch"  -P Coverage.cmake
```

The `.gcno` files are generated, but the `.gcda` files are not generated, so the process fails to complete.
I'm not sure why the `gcda` files are not being generated. I can see gcov symbols when I do `nm ./test/mptest | grep gcov`, but it doesn't work.

PR #221, opened 2025-10-01
test default PassField impl handles output parameters

Eunovo:
This PR adds two test cases to ensure that output parameter results from a server call are correctly captured and included in the response.
The new tests also serve as practical examples of how to work with output parameters and `in|out` parameters in the codebase.

### Testing
Automated coverage is not yet available, see https://github.com/bitcoin-core/libmultiprocess/issues/198.

As a temporary verification method:

- Comment out [these lines](https://github.com/bitcoin-core/libmultiprocess/blob/47d79db8a5528097b408e18f7b0bae11a6702d26/include/mp/proxy-types.h#L308-L310)
 and run make check. The new test should fail, confirming it is effective.

- If you then also comment out the newly added tests and rerun make check, it should pass, demonstrating that the lines above were previously uncovered.

### DIFF
--- a/test/mp/test/foo.capnp
+++ b/test/mp/test/foo.capnp
@@ -13,6 +13,7 @@ $Proxy.includeTypes("mp/test/foo-types.h");
 
 interface FooInterface $Proxy.wrap("mp::test::FooImplementation") {
     add @0 (a :Int32, b :Int32) -> (result :Int32);
+    addOut @19 (a :Int32, b :Int32) -> (ret :Int32);
     mapSize @1 (map :List(Pair(Text, Text))) -> (result :Int32);
     pass @2 (arg :FooStruct) -> (result :FooStruct);
     raise @3 (arg :FooStruct) -> (error :FooStruct $Proxy.exception("mp::test::FooStruct"));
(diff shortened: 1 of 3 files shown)


test/mp/test/foo.capnp
```diff
@@ -13,6 +13,7 @@ $Proxy.includeTypes("mp/test/foo-types.h");
 
 interface FooInterface $Proxy.wrap("mp::test::FooImplementation") {
     add @0 (a :Int32, b :Int32) -> (result :Int32);
+    addOut @19 (a :Int32, b :Int32, ret :Int32) -> (ret :Int32);
```
ryanofsky, line 16: In commit "test default PassField impl handles output parameters" (f3688fdaed1fddd52755857f2b068927f1a428a7)

Since the test is supposed to demonstrate using output parameters and is not using the input value of `ret` it would be good to make the declaration reflect that by simplifying:

```diff
-    addOut @19 (a :Int32, b :Int32, ret :Int32) -> (ret :Int32);
+    addOut @19 (a :Int32, b :Int32) -> (ret :Int32);
```

If you wanted to extend the test to cover input/output parameters, too, you could add another function like:

```c++
void addInOut(int x, int& sum) { sum += x; }
```

```capnp
addInOut @20 (x :Int32, sum :Int32) -> (sum :Int32);
```

ryanofsky (commented):
Code review f3688fdaed1fddd52755857f2b068927f1a428a7. Nice test! I'm surprised there wasn't already a test for output parameters. Would suggest making the simplification below but otherwise this looks good.

ryanofsky (approved):
Code review ACK 6db669628387b481b0d5726f011eb45025f31ccf. Thanks, both changes look good!

PR #223, opened 2025-10-02
ci: Replace nix-shell with equivalent nix develop command

ryanofsky:
Functionality is the same, this just shows nicer status information when building locally. (For even nicer status information you can install nix-output-monitor and replace "nix develop" with "nom develop")


TheCharlatan (approved):
ACK 18a2237a8ef4129c477301ad567baa524aaa231e

PR #225, opened 2025-10-07
Improve and document act support

Sjors:
Facilitate using [act](https://github.com/nektos/act) with Docker / Podman as an alternative for using nix directly.

I found this very useful for testing the sanitizer job against individual commits in #222. That is, once I finally got it to work...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,11 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
+    env:
+      NIX_EXTRA_CONFIG_ACT: |
+        sandbox = false
+        filter-syscalls = false
+
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


ryanofsky:
Nice changes. They seem very useful. It seems like they could be especially useful for testing non-nix jobs like openbsd locally, though not sure if that requires KVM support.

Am a little curious why you needed this to run the sanitize job. Was it a macos issue? I regularly use the steps in the readme to run CI scripts locally on my machine without needing any containers (e.g. `CI_CONFIG=ci/configs/sanitize.bash ci/scripts/run.sh`) and that has seemed like the most fast and efficient way possible to run them.

ryanofsky (approved):
Code review ACK 4e3f8fa0d2cf32198cf2b1587fd14d014d26dd18. Nice changes, and sorry for not reviewing this earlier. This seems like a useful feature and I plan to use it myself to test githuba ctions changes. Setting up the `/nix` subvolume in the instructions is also really nice since it should enable builds to run really quickly.

PR #236, opened 2025-12-30
ci: Install binary package `capnproto` on OpenBSD instead of building it

hebasto:
OpenBSD 7.8 now provides a binary package for [`capnproto`](https://www.ports.to/path/devel/capnproto.html).


ryanofsky (approved):
Code review ACK 1ee909393f4f4b3e0198d6c4b734247e46398339. Nice simplification!

PR #241, opened 2026-01-21
doc: Bump version number v7 -> v8

ryanofsky:
Increase version number after previous version was added to bitcoin core. Also update version history and documentation.

### DIFF
--- a/doc/versions.md
+++ b/doc/versions.md
@@ -7,40 +7,55 @@ Library versions are tracked with simple
 Versioning policy is described in the [version.h](../include/mp/version.h)
 include.
 
-## v7
+## v8
 - Current unstable version in master branch.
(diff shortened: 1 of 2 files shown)


ryanofsky:
Thanks for the review!

> Since the version is encoded in `MP_MAJOR_VERSION` we can't retroactively adjust the tags.

Exactly yes. I added the `v5.0-pre1` tag when I wanted to tag a version between `v4.0` and `v5.0` and I added a `v7.0-pre1` tag when I wanted to tag a version between `v6.0` and `v7.0`. This seemed less awkward than renumbering due to `MP_MAJOR_VERSION` being present. In the future, it shouldn't be necessary to have any more `-pre` tags as long as the version number is bumped consistently. When I initially created the `v0`..`v7` tags, I just didn't create enough of them and didn't leave space between them.

PR #242, opened 2026-02-20
proxy-types: add CustomHasField hook to map Cap'n Proto values to null C++ values

Sjors:
Taken from: https://github.com/bitcoin/bitcoin/pull/34020#issuecomment-3624001654

Let applications override `CustomHasField` so they can decide to treat certain capnproto values as being unset. For example, when converting `List(Data)` to `vector<shared_ptr<CTransaction>>`, mapping empty `Data` fields to null pointers.

This safe to do in special cases, like in this example because serialized `CTransaction` representations are never empty. It is also useful to do in this case because Cap'n Proto doesn't currently provide any API for distinguishing between unset and empty data values in a list (although they can be distinguished on the wire).

### DIFF
--- a/include/mp/proxy-types.h
+++ b/include/mp/proxy-types.h
@@ -23,9 +23,10 @@ public:
     ValueField(Value&& value) : m_value(value) {}
     Value& m_value;
 
+    const Value& get() const { return m_value; }
     Value& get() { return m_value; }
     Value& init() { return m_value; }
-    bool has() { return true; }
+    bool has() const { return true; }
 };
 
 template <typename Accessor, typename Struct>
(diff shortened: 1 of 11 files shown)


Sjors:
Added missing `#include <vector>` to `mp/gen.cpp`.

ryanofsky (approved):
Code review ACK 413f91597932921176904492ca09673549750fd1. Code here looks good. I think PR title & description need to be updated (also title & description of main commit) because they are referencing `CTransactionRef` which is not in this repository and also not affected by this change (it requires the `CustomHasField(TypeList<CTransaction>...)` overload in the original commit).

A good summary of changes here might be "Add CustomHasValue overload to allow mapping non-null Cap'n Proto values to null C++ values" and probably the LLM can turn that into a good title & description not referring to CTransactionRef

Also wondering if you'd want this added to https://github.com/bitcoin/bitcoin/pull/34422 if merged (seems reasonable)

ryanofsky:
Code review ACK 2d3cc775e655965b6e70c6c22b1e14c5fd527eb7, with commit message rewrite and comment fix since last review.

FWIW I think new commit message is clear, but new PR description is pretty random and doesn't say what the change does. Not important, but something like this might be better:

> Let applications override `CustomHasField` so they can decide to treat certain capnproto values as being unset. For example, when converting `List(Data)` to `vector<shared_ptr<CTransaction>>`, mapping empty `Data` fields to null pointers.
> This safe to do in special cases, like in this example because serialized CTransaction representations are never empty. It is also useful to do in this case because Cap'n Proto doesn't currently provide any API for distinguishing between unset and empty data values in a list (although they can be distinguished on the wire).

PR #243, opened 2026-02-20
mpgen: support primitive std::optional struct fields

ryanofsky:
Currently C++ structs with primitive `std::optional` members (ints, bools, floats) cannot easily by mapped to Cap'n Proto structs because Cap'n Proto does not provide a way to leave primitive fields unset, so there isn't a natural way to represent `std::nullopt` values. This PR makes it possible to map C++ structs with fields like:

```c++
    std::optional<int> foo;
```

to Cap'n Proto structs by using extra `Bool` fields prefixed with "has" for primitive optional members:

```capnp
    foo @3 :Int32;
    hasFoo @4 :Bool;
```

Boolean "has" fields were already supported by the code generator and used to pass primitive `std::optional`  parameters and return values, so this PR just extends it work with all struct fields, not just fields in params and result structs.

Note: Motivation for this change is dealing with the [`CreatedTransactionResult::change_pos`](https://github.com/furszy/bitcoin-gui/blob/4c0d4f6f93f371a8ad097735945d32510a7e83bb/src/wallet/types.h#L39) field introduced to the wallet interface in bitcoin-core/gui#807. This also could have been useful in https://github.com/bitcoin/bitcoin/pull/33965#issuecomment-3937522269

### DIFF
--- a/src/mp/gen.cpp
+++ b/src/mp/gen.cpp
@@ -125,6 +125,87 @@ static bool BoxedType(const ::capnp::Type& type)
              type.isFloat64() || type.isEnum());
 }
 
+struct Field
+{
+    ::capnp::StructSchema::Field param;
+    bool param_is_set = false;
+    ::capnp::StructSchema::Field result;
+    bool result_is_set = false;
+    int args = 0;
+    bool retval = false;
+    bool optional = false;
+    bool requested = false;
+    bool skip = false;
+    kj::StringPtr exception;
+};
+
+struct FieldList
+{
+    std::vector<Field> fields;
+    std::map<kj::StringPtr, int> field_idx; // name -> args index
+    bool has_result = false;
+
+    void addField(const ::capnp::StructSchema::Field& schema_field, bool param)
+    {
+        if (AnnotationExists(schema_field.getProto(), SKIP_ANNOTATION_ID)) {
+            return;
+        }
+
+        auto field_name = schema_field.getProto().getName();
+        auto inserted = field_idx.emplace(field_name, fields.size());
+        if (inserted.second) {
+            fields.emplace_back();
+        }
+        auto& field = fields[inserted.first->second];
+        if (param) {
+            field.param = schema_field;
+            field.param_is_set = true;
+        } else {
+            field.result = schema_field;
+            field.result_is_set = true;
+        }
+
+        if (!param && field_name == kj::StringPtr{"result"}) {
+            field.retval = true;
+            has_result = true;
+        }
+
+        GetAnnotationText(schema_field.getProto(), EXCEPTION_ANNOTATION_ID, &field.exception);
+
+        int32_t count = 1;
+        if (!GetAnnotationInt32(schema_field.getProto(), COUNT_ANNOTATION_ID, &count)) {
+            if (schema_field.getType().isStruct()) {
+                GetAnnotationInt32(schema_field.getType().asStruct().getProto(),
+                        COUNT_ANNOTATION_ID, &count);
+            } else if (schema_field.getType().isInterface()) {
+                GetAnnotationInt32(schema_field.getType().asInterface().getProto(),
+                        COUNT_ANNOTATION_ID, &count);
+            }
+        }
+
+
+        if (inserted.second && !field.retval && !field.exception.size()) {
+            field.args = count;
+        }
+    }
+
+    void mergeFields()
+    {
+        for (auto& field : field_idx) {
+            auto has_field = field_idx.find("has" + Cap(field.first));
+            if (has_field != field_idx.end()) {
+                fields[has_field->second].skip = true;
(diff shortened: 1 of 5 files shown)


Sjors:
It would be useful to demonstrate that other implementations can straightforwardly handle this, e.g. with a functional tests in a Bitcoin Core branch (or PR).

ryanofsky:
> It would be useful to demonstrate that other implementations can straightforwardly handle this, e.g. with a functional tests in a Bitcoin Core branch (or PR).

My next rebase of #10102 will use this for the `change_pos` field as mentioned in the PR description, but that's probably a less obvious application than line a mining interface change accepting optional BlockCreateOptions. Technically https://github.com/bitcoin/bitcoin/pull/33966 could do that but I think because of complexity of that PR it's best if that PR just only updates the C++ mining interface and leaves extending the Cap'n Proto to as a followup.

In the meantime I updated the PR description with a more specific example of how this can be used and I think the unit test should provide a good test of functionality.


Updated 9f99c7d8e0f5bbdbf5262094b6b9fe515c19bf05 -> 4de44299ea6d981f79c41fbbd76b61d902dd3580 ([`pr/optint.1`](https://github.com/ryanofsky/libmultiprocess/commits/pr/optint.1) -> [`pr/optint.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/optint.2), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/optint.1..pr/optint.2)) adding a refactoring commit deduplicating more code between parameter & struct field mapping implementations and reducing size of the last commit

enirox001:
Concept ACK

src/mp/gen.cpp
```diff
@@ -484,13 +487,11 @@ static void Generate(kj::StringPtr src_prefix,
                 inl << "public:\n";
                 inl << "    using Struct = " << message_namespace << "::" << node_name << ";\n";
                 size_t i = 0;
-                for (const auto field : struc.getFields()) {
-                    if (AnnotationExists(field.getProto(), SKIP_ANNOTATION_ID)) {
-                        continue;
-                    }
-                    auto field_name = field.getProto().getName();
+                for (const auto& field : fields.fields) {
```
enirox001, line 490: In commit "mpgen: support primitive std::optional struct fields" (https://github.com/bitcoin-core/libmultiprocess/pull/243/changes/6dbfa56a04012772e900214fe11d9f238f7c3449):

I think the `hasX` field here is intended as internal serialization plumbing for the logical optional field, so I would have expected it to stay hidden from the generated `ProxyStruct` surface rather than also getting its own public accessor alias.

I believe a minimal fix would be to skip alias generation for `field.skip` entries in the `ProxyStruct` loop, similar to how they are already skipped when building the Accessors tuple `below`. Something along these lines:

```cpp
for (const auto& field : fields.fields) {
      if (field.skip) continue;
      auto field_name = field.param.getProto().getName();
      add_accessor(field_name);
      dec << "    using " << Cap(field_name) << "Accessor =
  "
          << AccessorType(base_name, field) << ";\n";
  
```

enirox001 (commented):
I think this is a useful extension of the existing `hasX` convention to wrapped struct fields, and I believe it would be a net positive for downstream libmultiprocess users.

My understanding is that this teaches `mpgen` to represent primitive `std::optional` struct members the same way optional method params/results are already represented today: a `value` field plus a `hasValue` boolean. I think that is a sensible direction because it reuses an existing convention instead of introducing a separate encoding rule for struct fields.

src/mp/gen.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/243#discussion_r3187887614

> I think the `hasX` field here is intended as internal serialization plumbing for the logical optional field, so I would have expected it to stay hidden from the generated `ProxyStruct` surface rather than also getting its own public accessor alias.

Yes originally, the PR took an approach like you suggested, but it was changed in a later push https://github.com/ryanofsky/libmultiprocess/compare/pr/optint.2..pr/optint.3 because it broke https://github.com/bitcoin/bitcoin/pull/10102. Specifically, I think it broke this code trying to use the accessor of a skipped field:

https://github.com/ryanofsky/bitcoin/blob/pr/ipc.232/src/ipc/capnp/node.capnp#L172
https://github.com/ryanofsky/bitcoin/blob/pr/ipc.232/src/ipc/capnp/node-types.h#L87

Currently, the `Field::skip` struct member is a little overloaded. It is true for has & want fields but it is also true for fields where `$Proxy.skip` was explicitly used. And it is useful to have accessors for `$Proxy.skip` fields to write custom serialization code, even it if it is probably not useful to have accessors for has & want fields.

Since outputting accessors does not cost much, though, I think it should not be a problem to have a few extra one even if they are less likely to be used.

PR #244, opened 2026-02-24
ci: suppress two tidy lint issues

Sjors:
Needed for clang-22 tidy job, see:
https://github.com/bitcoin/bitcoin/pull/34662


ryanofsky:
Code review ACK bc2c12ecfe88a9c79633e2ce46f638993007633a. I do think it'd be good to drop `// NOLINT(bugprone-unused-return-value)` and use `(void)` instead, and am pretty sure that should be sufficient for https://github.com/bitcoin/bitcoin/pull/34662#discussion_r2848564166 but we can see. Planning to hold off merging for now until there's an update.

PR #250, opened 2026-03-11
Fix race conditions on disconnects

Sjors:
The PR fixes 3 race conditions on disconnects that were detected in Bitcoin core CI runs and by antithesis:

* [src/ipc/libmultiprocess/src/mp/proxy.cpp:416:44: runtime error: member access within null pointer of type 'mp::Waiter' bitcoin/bitcoin#34711 (comment)](https://github.com/bitcoin/bitcoin/issues/34711#issuecomment-3986380070)
* [ipc: AddressSanitizer: heap-use-after-free in `capnp::CallContext<ipc::capnp::messages::BlockTemplate::GetBlockParams, ipc::capnp::messages::BlockTemplate::GetBlockResults>::getParams()` bitcoin/bitcoin#34777 (comment)](https://github.com/bitcoin/bitcoin/issues/34777#issuecomment-4024285314)
* [mptest: mp::CancelMonitor: AddressSanitizer: stack-use-after-return bitcoin/bitcoin#34782 (comment)](https://github.com/bitcoin/bitcoin/issues/34782#issuecomment-4033169085)

### DIFF
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -461,6 +461,11 @@ public:
     //! ThreadMap.makeThread) used to service requests to clients.
     ::capnp::CapabilityServerSet<Thread> m_threads;
 
+    //! Hook called on the worker thread inside makeThread, right after
+    //! set_value. Used by tests to verify the waiter mutex is held when
+    //! the thread context is published.
+    std::function<void()> testing_hook_makethread;
+
     //! Canceler for canceling promises that we want to discard when the
     //! connection is destroyed. This is used to interrupt method calls that are
     //! still executing at time of disconnection.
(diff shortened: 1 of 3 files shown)


ryanofsky:
Thanks for updating this! I opened https://github.com/bitcoin/bitcoin/pull/34804 to see how this interacts with bitcoin core CI.

(I wonder if there is a way we automatically run bitcoin core CI on every libmultiprocess PR, or run it more easily. Might be something to look into).

> The LLM came up with a third test, but beyond reproducing the original issue, it doesn't seem very useful. So I omitted it in this PR. See [Sjors@7c59dac](https://github.com/Sjors/libmultiprocess/commit/7c59dacbddd0779240c14f554c18605b2d39cbdd)

Didn't look very closely but the test seems clean and I'd be inclined to add it since I already know I want to follow up this PR by duplicating the disconnect tests and maybe reviving my earlier [attempt](https://github.com/bitcoin-core/libmultiprocess/pull/201#issuecomment-3302131507). If the test turns out not to be meaningful or is a maintenance burden we can remove it.

Anyway I plan to review this soon

PR #253, opened 2026-03-11
ci: run some Bitcoin Core CI jobs 

Sjors:
Slightly more elaborate version of #252 with ccache and depends caching.

### DIFF
--- a/.github/workflows/bitcoin-core-ci.yml
+++ b/.github/workflows/bitcoin-core-ci.yml
@@ -32,9 +32,9 @@ jobs:
       matrix:
         include:
           - name: 'ASan + UBSan'
-            unit_test_runs: 1
-            functional_test_runs: 1
-            nproc_multiplier: 1
+            unit_test_runs: 15
+            functional_test_runs: 300
+            nproc_multiplier: 10
             functional_timeout_factor: 4
             runner: ubuntu-24.04
             apt-llvm: true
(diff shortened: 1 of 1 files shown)


.github/workflows/bitcoin-core-ci.yml
```diff
@@ -0,0 +1,344 @@
...
+          echo "deb http://apt.llvm.org/${VERSION_CODENAME}/ llvm-toolchain-${VERSION_CODENAME}-${LLVM_VERSION} main" | sudo tee "/etc/apt/sources.list.d/llvm.list"
+          sudo apt-get update
+
+      - name: Install APT packages
+        if: matrix.packages
+        run: |
+          sudo apt-get install --no-install-recommends -y ${{ matrix.packages }}
+          sudo update-alternatives --install /usr/bin/clang++ clang++ "/usr/bin/clang++-${LLVM_VERSION}" 100
+          sudo update-alternatives --install /usr/bin/clang clang "/usr/bin/clang-${LLVM_VERSION}" 100
+          sudo update-alternatives --install /usr/bin/llvm-symbolizer llvm-symbolizer "/usr/bin/llvm-symbolizer-${LLVM_VERSION}" 100
```
maflcko: I wonder if we can split the upstream install script into smaller re-useable bash or python scripts, to be used here?

Sjors:
@maflcko this doesn't include #250 so might not catch those?

fanquake:
> this doesn't include #250 so might not catch those?

If it doesn't include #250, then shouldn't it catch all the bugs being fixed by #250? (or any of the others that exist with #250 included, https://github.com/bitcoin/bitcoin/pull/34804#issuecomment-4047157116, assuming they don't require #250 to be merged).

Sjors:
Test run times:

- macOS: unit 4s, functional ?s
- TSAN: unit 15s, functional ?s
- ASan / UBSan: unit 10s, functional ?s

Unit tests are still running, from previous experience the functional tests took about 1 second each.

I'd like to keep the job runtimes under 15 minutes:
- olddeps is the slowest, takes about 10 minutes
- gnu32 currently takes 1+ hour, but hopefully #256 speeds it up

Assuming 3 minutes for setup, 4 minutes for unit tests and 4 minutes for functional tests, I'll tweak the numbers.

(also forgot to install ccache on macOS)

maflcko:
ok, none failed :(

PR #271, opened 2026-04-13
test: improve TimingResistantEqual test coverage

mangoostaa:
This improves the test for `TimingResistantEqual` by adding more relevant edge cases, including:

- Empty string vs non-empty string
- Differences at the beginning of the string
- Additional prefix/suffix length mismatch tests

These cases help ensure the function properly resists timing attacks when comparing secrets of different lengths

### DIFF
--- a/include/mp/util.h
+++ b/include/mp/util.h
@@ -14,6 +14,7 @@
 #include <kj/string-tree.h>
 #include <mutex>
 #include <string>
+#include <string_view>
 #include <tuple>
 #include <type_traits>
 #include <utility>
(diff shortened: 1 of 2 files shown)


ViniciusCestarii:
Is there a planned call site for `TimingResistantEqual`? Is there any place libmultiprocess would need a constant-time comparison?

mangoostaa:
> Is there a planned call site for `TimingResistantEqual`? Is there any place libmultiprocess would need a constant-time comparison?

You're right, currently there is no clear call site in libmultiprocess that requires constant-time comparison.
This test only improves coverage for the existing function, but I agree that without a concrete use case it might not be justified here.

ryanofsky:
Will close this PR. Change seems ok but would make more sense to review & merge when there is something else that can use this.

Note that bitcoin core already has a similar [function](https://github.com/bitcoin/bitcoin/blob/61d1c78ed4168c1033f850d1f7f628bf4cf7af83/src/util/strencodings.h#L196-L209) and [tests](https://github.com/bitcoin/bitcoin/blob/61d1c78ed4168c1033f850d1f7f628bf4cf7af83/src/test/util_tests.cpp#L550-L559).

PR #231, opened 2025-10-22
Add windows support

ryanofsky:
Add support for running on windows. These changes make the libmultiprocess API more generic, using stream types instead of file descriptors. All features are supported, including spawning processes with socket connections to the parent process. These changes were originally made in https://github.com/bitcoin/bitcoin/pull/32387

PR #272, opened 2026-04-20
ci: add Windows job

Sjors:
Based on #231

CI commits are entirely vibe coded, so will need some polishing if it actually works,


Sjors:
This will need whitelisting of `msys2/setup-msys2`.

But I'm thinking of switching to cross-compilation, since that's how most users will use this library anyway.

Sjors:
Ok, similar to Bitcoin Core we now cross-compile and then run the result on a Windows machine.

The agent made a few other adjustments to the code that I haven't looked at yet.

This needs the following whitelist:

- actions/upload-artifact
- actions/download-artifact

Sjors:
@ryanofsky can you whitelist the upload / download actions I mentioned above?

ryanofsky:
I think we will need some CI coverage for windows, and this approach looks ok, but it's also compilicated.

I wonder if it would be possible to implement a more minimal approach using nix's mingw compiler and wine (`nativeBuildInputs = [ pkgs.pkgsCross.mingwW64.buildPackages.gcc pkgs.wine ];`) to just build the tests and examples normally and run them with wine.

This would seem helpful for local development and debugging. If CI job fails just run it locally with `CI_CONFIG=ci/configs/windows.bash ci/scripts/run.sh` or similar.

Sjors:
I'll first try to get the current attempt to run and then will look at some different approaches. Building with nix makes sense. Running in wine could also make sense, though I suspect a native run will catch other things, so maybe we should do both.

Sjors:
I switched cross-compilation to nix. The (cold cache) build takes _very_ long on CI, so I'll check again tomorrow and then see if I can clean things up.

Sjors:
Alright, CI green!

@ryanofsky I'll wait for you to get #231 green again before rebasing and then I'll look into cleanup. You can cherry-pick things out of here if you think they're correct, but otherwise feel free to ignore. I need to study that PR first to better understand what does and doesn't make sense here.

ryanofsky:
Thanks! I looked more at the nix/wine changes here than the native windows stuff but it looks like everything makes sense. My plan is to split #231 up and open a new PR that only makes non-windows changes. Then #231 can depend on that and this can depend on #231.

Alternately, this PR could be implemented independently of the other PRs by just having the windows &wine jobs build libmultiprocess and run mptest, adding a dummy test that can run on windows like

```c++
KJ_TEST("Hello world")
{
    KJ_EXPECT(true);
}
```

and skipping the other tests that don't compile on windows yet. But this is useful for testing in its current form too. Just suggesting in case you are impatient to get this merged.

Sjors:
I'm happy to stay based on #231 so there's some useful coverage.

Sjors:
Rebased. The agent ended up dropping some commits so that's a good sign.

@ryanofsky can you check if 07a7f7131e1555053dec160a6a4eb4c0bc23cdde has something useful for #231?

I bumped the capnproto version for Windows builds to 1.4.0. That gets rid of one patch: https://github.com/bitcoin-core/libmultiprocess/pull/272/changes/67e5833039419883da5d2591ee08783eeeed879f#diff-022254a27cb514a71c153b75204b316472ba0a5bdf9e20041d2fc17cc2007806

Although it's a small patch, I don't see a good reason to support older versions when we're adding a new platform like this. Especially because I expect most mining users to use our binaries.

The other patch I just upstreamed: https://github.com/capnproto/capnproto/pull/2633

ryanofsky:
> @ryanofsky can you check if [07a7f71](https://github.com/bitcoin-core/libmultiprocess/commit/07a7f7131e1555053dec160a6a4eb4c0bc23cdde) has something useful for #231?

Yes it looks like this fixes the same issue @ViniciusCestarii reported https://github.com/bitcoin-core/libmultiprocess/pull/231#issuecomment-4424748490. The fix there seems simpler, though, and I think a better fix might be possible by dropping the winsock2.h include header to avoid windows headers being pulled in so many places.

The main commit here d97c586b6ef63cd73585826dcca5ef744670cce7 also looks good, though the commit message looks out of date, it says it is testing an msvc build but only actually testing a cross-compiled one

Sjors:
I probably going to hold off rebasing this until after #274 lands, or at least I've had some time to review the windows changes myself.

Big621 (commented):
K

PR #274, opened 2026-04-22
Add nonunix platform support

ryanofsky:
This PR implements API changes and fixes needed to allow libmultiprocess to work on nonunix platforms.

These changes were originally part of #231, which adds windows support, but were split out to allow windows and nonwindows changes to be reviewed separately.

### DIFF
--- a/ci/configs/olddeps.bash
+++ b/ci/configs/olddeps.bash
@@ -1,5 +1,5 @@
 CI_DESC="CI job using old Cap'n Proto and cmake versions"
 CI_DIR=build-olddeps
 export CXXFLAGS="-Werror -Wall -Wextra -Wpedantic -Wno-unused-parameter -Wno-error=array-bounds"
-NIX_ARGS=(--argstr capnprotoVersion "0.7.1" --argstr cmakeVersion "3.12.4")
+NIX_ARGS=(--argstr capnprotoVersion "0.9.2" --argstr cmakeVersion "3.12.4")
 BUILD_ARGS=(-k)
--- a/doc/versions.md
+++ b/doc/versions.md
@@ -7,8 +7,14 @@ Library versions are tracked with simple
 Versioning policy is described in the [version.h](../include/mp/version.h)
 include.
 
-## v13
+## v14
+- Adds support for nonunix platforms, making API changes that are not backwards compatible ([#274](https://github.com/bitcoin-core/libmultiprocess/pull/274)).
 - Current unstable version.
+
+## [v13.0](https://github.com/bitcoin-core/libmultiprocess/commits/v13.0)
+- Fixes a `memcpy` ubsan warning by using `std::ranges::copy` ([#305](https://github.com/bitcoin-core/libmultiprocess/pull/305)).
+- Misc: map serialization round-trip test coverage ([#297](https://github.com/bitcoin-core/libmultiprocess/pull/297)) and version bump ([#306](https://github.com/bitcoin-core/libmultiprocess/pull/306)).
+- Used in Bitcoin Core 32.x and newer releases with [#35720](https://github.com/bitcoin/bitcoin/pull/35720).
 
 ## [v12.0](https://github.com/bitcoin-core/libmultiprocess/commits/v12.0)
 - Adds an optional `max_connections` parameter to `ListenConnections` ([#269](https://github.com/bitcoin-core/libmultiprocess/pull/269)).
@@ -24,7 +30,7 @@ include.
 - Fixes a rare mptest hang on musl builds caused by a lost wakeup bug in `Waiter` ([#295](https://github.com/bitcoin-core/libmultiprocess/pull/295)).
 - Fixes a race condition in a log print detected by TSan ([#286](https://github.com/bitcoin-core/libmultiprocess/pull/286)).
 - Build improvements: makes `target_capnp_sources` work correctly when libmultiprocess is used as a CMake subproject ([#289](https://github.com/bitcoin-core/libmultiprocess/pull/289)), adds `mp_headers` target for better lint tool support ([#291](https://github.com/bitcoin-core/libmultiprocess/pull/291)), and fixes compatibility with recent Nix and CMake 4.0 ([#238](https://github.com/bitcoin-core/libmultiprocess/pull/238)).
-- Test, CI, documentation, and minor code improvements: design document corrections ([#278](https://github.com/bitcoin-core/libmultiprocess/pull/278)), field constant comments ([#279](https://github.com/bitcoin-core/libmultiprocess/pull/279)), clang-tidy fix ([#292](https://github.com/bitcoin-core/libmultiprocess/pull/292)), new smoke test for double-precision float values ([#294](https://github.com/bitcoin-core/libmultiprocess/pull/294)), new test for recursive async IPC calls ([#301](https://github.com/bitcoin-core/libmultiprocess/pull/301)), removal of libevent from Core CI builds ([#299](https://github.com/bitcoin-core/libmultiprocess/pull/299)), and rename of `EventLoop::m_num_clients` to `m_num_refs` ([#302](https://github.com/bitcoin-core/libmultiprocess/pull/302)).
+- Test, CI, documentation, and minor code improvements: design document corrections ([#278](https://github.com/bitcoin-core/libmultiprocess/pull/278)), field constant comments ([#279](https://github.com/bitcoin-core/libmultiprocess/pull/279)), clang-tidy fix ([#292](https://github.com/bitcoin-core/libmultiprocess/pull/292)), new smoke test for double-precision float values ([#294](https://github.com/bitcoin-core/libmultiprocess/pull/294)), new test for recursive async IPC calls ([#301](https://github.com/bitcoin-core/libmultiprocess/pull/301)), removal of libevent from Core CI builds ([#299](https://github.com/bitcoin-core/libmultiprocess/pull/299)), rename of `EventLoop::m_num_clients` to `m_num_refs` ([#302](https://github.com/bitcoin-core/libmultiprocess/pull/302)), and version bump ([#270](https://github.com/bitcoin-core/libmultiprocess/pull/270)).
 - Used in Bitcoin Core 32.x and newer releases with [#35661](https://github.com/bitcoin/bitcoin/pull/35661).
 
 ## [v10.0](https://github.com/bitcoin-core/libmultiprocess/commits/v10.0)
--- a/example/calculator.cpp
+++ b/example/calculator.cpp
@@ -6,8 +6,7 @@
 #include <init.capnp.h>
 #include <init.capnp.proxy.h> // NOLINT(misc-include-cleaner) // IWYU pragma: keep
 
-#include <charconv>
-#include <cstring>
+#include <cstring> // IWYU pragma: keep
 #include <fstream>
 #include <functional>
 #include <iostream>
@@ -16,9 +15,9 @@
 #include <kj/memory.h>
 #include <memory>
 #include <mp/proxy-io.h>
+#include <mp/util.h>
 #include <stdexcept>
 #include <string>
-#include <system_error>
 #include <utility>
 
 class CalculatorImpl : public Calculator
@@ -51,14 +50,10 @@ int main(int argc, char** argv)
         std::cout << "Usage: mpcalculator <fd>\n";
         return 1;
     }
-    mp::SocketId fd;
-    if (std::from_chars(argv[1], argv[1] + strlen(argv[1]), fd).ec != std::errc{}) {
-        std::cerr << argv[1] << " is not a number or is larger than an int\n";
-        return 1;
-    }
+    mp::SocketId socket{mp::StartSpawned(argv[1])};
     mp::EventLoop loop("mpcalculator", LogPrint);
     std::unique_ptr<Init> init = std::make_unique<InitImpl>();
-    mp::ServeStream<InitInterface>(loop, fd, *init);
+    mp::ServeStream<InitInterface>(loop, socket, *init);
     loop.loop();
     return 0;
 }
--- a/example/calculator.cpp
+++ b/example/calculator.cpp
@@ -51,7 +51,7 @@ int main(int argc, char** argv)
         std::cout << "Usage: mpcalculator <fd>\n";
         return 1;
     }
-    int fd;
+    mp::SocketId fd;
     if (std::from_chars(argv[1], argv[1] + strlen(argv[1]), fd).ec != std::errc{}) {
         std::cerr << argv[1] << " is not a number or is larger than an int\n";
         return 1;
--- a/example/calculator.cpp
+++ b/example/calculator.cpp
@@ -53,7 +53,7 @@ int main(int argc, char** argv)
     mp::SocketId socket{mp::StartSpawned(argv[1])};
     mp::EventLoop loop("mpcalculator", LogPrint);
     std::unique_ptr<Init> init = std::make_unique<InitImpl>();
-    mp::ServeStream<InitInterface>(loop, socket, *init);
+    mp::ServeStream<InitInterface>(loop, mp::MakeStream(loop, socket), *init);
     loop.loop();
     return 0;
 }
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -456,7 +456,7 @@ public:
     //! Capability::Client handles owned by ProxyClient objects), then schedules
     //! asynchronous cleanup functions to run in a worker thread (to run
     //! destructors of m_impl instances owned by ProxyServer objects).
-    ~Connection();
+    ~Connection() noexcept(false);
 
     //! Register synchronous cleanup function to run on event loop thread (with
     //! access to capnp thread local variables) when disconnect() is called.
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -18,6 +18,7 @@
 #include <cstdlib>
 #include <functional>
 #include <kj/function.h>
+#include <kj/io.h>
 #include <map>
 #include <memory>
 #include <optional>
@@ -311,11 +312,12 @@ public:
     //! Callback functions to run on async thread.
     std::optional<CleanupList> m_async_fns MP_GUARDED_BY(m_mutex);
 
-    //! Pipe read handle used to wake up the event loop thread.
-    SocketId m_wait_fd = SocketError;
+    //! Socket pair used to post and wait for wakeups to the event loop thread.
+    kj::Own<kj::AsyncIoStream> m_wait_stream;
+    kj::Own<kj::AsyncIoStream> m_post_stream;
 
-    //! Pipe write handle used to wake up the event loop thread.
-    SocketId m_post_fd = SocketError;
+    //! Synchronous writer used to write to m_post_stream.
+    kj::Own<kj::OutputStream> m_post_writer;
 
     //! Number of EventLoopRef instances referencing this event loop. This is a
     //! sum of the number of client and server objects (Connection, ProxyClient,
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -312,10 +312,10 @@ public:
     std::optional<CleanupList> m_async_fns MP_GUARDED_BY(m_mutex);
 
     //! Pipe read handle used to wake up the event loop thread.
-    int m_wait_fd = -1;
+    SocketId m_wait_fd = SocketError;
 
     //! Pipe write handle used to wake up the event loop thread.
-    int m_post_fd = -1;
+    SocketId m_post_fd = SocketError;
 
     //! Number of EventLoopRef instances referencing this event loop. This is a
     //! sum of the number of client and server objects (Connection, ProxyClient,
@@ -917,10 +917,10 @@ void ServeStream(EventLoop& loop, int fd, InitImpl& init)
         [] {});
 }
 
-//! Given listening socket file descriptor and an init object, handle incoming
+//! Given listening socket identifier and an init object, handle incoming
 //! connections and requests by calling methods on the Init object.
 template <typename InitInterface, typename InitImpl>
-void ListenConnections(EventLoop& loop, int fd, InitImpl& init, std::optional<size_t> max_connections = std::nullopt)
+void ListenConnections(EventLoop& loop, SocketId fd, InitImpl& init, std::optional<size_t> max_connections = std::nullopt)
 {
     loop.sync([&]() {
         auto listener{std::make_shared<Listener>(
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -214,6 +214,9 @@ private:
 
 std::string LongThreadName(const char* exe_name);
 
+//! Wrap a socket file descriptor as an async stream, taking ownership of the fd.
+Stream MakeStream(EventLoop&loop, SocketId socket);
+
 //! Event loop implementation.
 //!
 //! Cap'n Proto threading model is very simple: all I/O operations are
@@ -826,17 +829,15 @@ kj::Promise<T> ProxyServer<Thread>::post(Fn&& fn)
     return ret;
 }
 
-//! Given stream file descriptor, make a new ProxyClient object to send requests
-//! over the stream. Also create a new Connection object embedded in the
-//! client that is freed when the client is closed.
+//! Given a stream, make a new ProxyClient object to send requests over it.
+//! Also create a new Connection object embedded in the client that is freed
+//! when the client is closed.
 template <typename InitInterface>
-std::unique_ptr<ProxyClient<InitInterface>> ConnectStream(EventLoop& loop, int fd)
+std::unique_ptr<ProxyClient<InitInterface>> ConnectStream(EventLoop& loop, Stream stream)
 {
     typename InitInterface::Client init_client(nullptr);
     std::unique_ptr<Connection> connection;
     loop.sync([&] {
-        auto stream =
-            loop.m_io_context.lowLevelProvider->wrapSocketFd(fd, kj::LowLevelAsyncIoProvider::TAKE_OWNERSHIP);
         connection = std::make_unique<Connection>(loop, kj::mv(stream));
         init_client = connection->m_rpc_system->bootstrap(ServerVatId().vat_id).castAs<InitInterface>();
         Connection* connection_ptr = connection.get();
@@ -907,16 +908,12 @@ void _Listen(const std::shared_ptr<Listener>& listener, EventLoop& loop, InitImp
         }));
 }
 
-//! Given stream file descriptor and an init object, handle requests on the
-//! stream by calling methods on the Init object.
+//! Given a stream and an init object, handle requests on the stream by calling
+//! methods on the Init object.
 template <typename InitInterface, typename InitImpl>
-void ServeStream(EventLoop& loop, int fd, InitImpl& init)
+void ServeStream(EventLoop& loop, Stream stream, InitImpl& init)
 {
-    _Serve<InitInterface>(
-        loop,
-        loop.m_io_context.lowLevelProvider->wrapSocketFd(fd, kj::LowLevelAsyncIoProvider::TAKE_OWNERSHIP),
-        init,
-        [] {});
+    _Serve<InitInterface>(loop, kj::mv(stream), init, [] {});
 }
 
 //! Given listening socket identifier and an init object, handle incoming
--- a/include/mp/util.h
+++ b/include/mp/util.h
@@ -12,6 +12,8 @@
 #include <cstring>
 #include <exception>
 #include <functional>
+#include <kj/async-io.h>
+#include <kj/memory.h>
 #include <kj/string-tree.h>
 #include <mutex>
 #include <string>
@@ -255,6 +257,8 @@ std::string ThreadName(const char* exe_name);
 //! Escape binary string for use in log so it doesn't trigger unicode decode
 //! errors in python unit tests.
 std::string LogEscape(const kj::StringTree& string, size_t max_size);
+
+using Stream = kj::Own<kj::AsyncIoStream>;
 
 using ProcessId = int;
 using SocketId = int;
--- a/include/mp/util.h
+++ b/include/mp/util.h
@@ -276,10 +276,9 @@ std::tuple<ProcessId, SocketId> SpawnProcess(SpawnConnectInfoToArgsFn&& connect_
 //! returning a socket id for communicating with the parent process.
 SocketId StartSpawned(const SpawnConnectInfo& connect_info);
 
-//! Call execvp with vector args.
-//! Not safe to call in a post-fork child of a multi-threaded process.
-//! Currently only used by mpgen at build time.
-void ExecProcess(const std::vector<std::string>& args);
+//! Start a process and return its process id. Caller should call WaitProcess
+//! on the returned id.
+ProcessId StartProcess(const std::vector<std::string>& args);
 
 //! Wait for a process to exit and return its exit code.
 int WaitProcess(ProcessId pid);
--- a/include/mp/util.h
+++ b/include/mp/util.h
@@ -145,7 +145,10 @@ struct PtrOrValue {
     std::variant<T*, T> data;
 
     template <typename... Args>
-    PtrOrValue(T* ptr, Args&&... args) : data(ptr ? ptr : std::variant<T*, T>{std::in_place_type<T>, std::forward<Args>(args)...}) {}
+    PtrOrValue(T* ptr, Args&&... args) : data(std::in_place_type<T*>, ptr)
+    {
+        if (!ptr) data.template emplace<T>(std::forward<Args>(args)...);
+    }
 
     T& operator*() { return data.index() ? std::get<T>(data) : *std::get<T*>(data); }
     T* operator->() { return &**this; }
--- a/include/mp/util.h
+++ b/include/mp/util.h
@@ -259,17 +259,22 @@ using ProcessId = int;
 using SocketId = int;
 constexpr SocketId SocketError{-1};
 
+//! Information about parent process passed to child process as a command-line
+//! argument. On unix this is the child socket fd number formatted as a string.
+using SpawnConnectInfo = std::string;
+
 //! Callback type used by SpawnProcess below.
-using FdToArgsFn = std::function<std::vector<std::string>(SocketId fd)>;
+using SpawnConnectInfoToArgsFn = std::function<std::vector<std::string>(const SpawnConnectInfo&)>;
 
 //! Spawn a new process that communicates with the current process over a socket
-//! pair. Returns pid through an output argument, and file descriptor for the
-//! local side of the socket.
-//! The fd_to_args callback is invoked in the parent process before fork().
-//! It must not rely on child pid/state, and must return the command line
-//! arguments that should be used to execute the process. Embed the remote file
-//! descriptor number in whatever format the child process expects.
-SocketId SpawnProcess(ProcessId& pid, FdToArgsFn&& fd_to_args);
+//! pair. Calls connect_info_to_args callback with a connection string that
+//! needs to be passed to the child process, and executes the argv command line
+//! it returns. Returns child process id and socket id.
+std::tuple<ProcessId, SocketId> SpawnProcess(SpawnConnectInfoToArgsFn&& connect_info_to_args);
+
+//! Initialize spawned child process using the SpawnConnectInfo string passed to it,
+//! returning a socket id for communicating with the parent process.
+SocketId StartSpawned(const SpawnConnectInfo& connect_info);
 
 //! Call execvp with vector args.
 //! Not safe to call in a post-fork child of a multi-threaded process.
--- a/include/mp/util.h
+++ b/include/mp/util.h
@@ -255,6 +255,8 @@ std::string ThreadName(const char* exe_name);
 //! errors in python unit tests.
 std::string LogEscape(const kj::StringTree& string, size_t max_size);
 
+using ProcessId = int;
+
 //! Callback type used by SpawnProcess below.
 using FdToArgsFn = std::function<std::vector<std::string>(int fd)>;
 
@@ -265,7 +267,7 @@ using FdToArgsFn = std::function<std::vector<std::string>(int fd)>;
 //! It must not rely on child pid/state, and must return the command line
 //! arguments that should be used to execute the process. Embed the remote file
 //! descriptor number in whatever format the child process expects.
-int SpawnProcess(int& pid, FdToArgsFn&& fd_to_args);
+int SpawnProcess(ProcessId& pid, FdToArgsFn&& fd_to_args);
 
 //! Call execvp with vector args.
 //! Not safe to call in a post-fork child of a multi-threaded process.
@@ -273,7 +275,7 @@ int SpawnProcess(int& pid, FdToArgsFn&& fd_to_args);
 void ExecProcess(const std::vector<std::string>& args);
 
 //! Wait for a process to exit and return its exit code.
-int WaitProcess(int pid);
+int WaitProcess(ProcessId pid);
 
 inline char* CharCast(char* c) { return c; }
 inline char* CharCast(unsigned char* c) { return (char*)c; }
--- a/include/mp/util.h
+++ b/include/mp/util.h
@@ -5,6 +5,7 @@
 #ifndef MP_UTIL_H
 #define MP_UTIL_H
 
+#include <array>
 #include <capnp/schema.h>
 #include <cassert>
 #include <cstdlib>
@@ -275,6 +276,10 @@ std::tuple<ProcessId, SocketId> SpawnProcess(SpawnConnectInfoToArgsFn&& connect_
 //! Initialize spawned child process using the SpawnConnectInfo string passed to it,
 //! returning a socket id for communicating with the parent process.
 SocketId StartSpawned(const SpawnConnectInfo& connect_info);
+
+//! Create a socket pair that can be used to communicate within a process or
+//! between parent and child processes.
+std::array<SocketId, 2> SocketPair();
 
 //! Start a process and return its process id. Caller should call WaitProcess
 //! on the returned id.
--- a/include/mp/util.h
+++ b/include/mp/util.h
@@ -256,9 +256,11 @@ std::string ThreadName(const char* exe_name);
 std::string LogEscape(const kj::StringTree& string, size_t max_size);
 
 using ProcessId = int;
+using SocketId = int;
+constexpr SocketId SocketError{-1};
 
 //! Callback type used by SpawnProcess below.
-using FdToArgsFn = std::function<std::vector<std::string>(int fd)>;
+using FdToArgsFn = std::function<std::vector<std::string>(SocketId fd)>;
 
 //! Spawn a new process that communicates with the current process over a socket
 //! pair. Returns pid through an output argument, and file descriptor for the
@@ -267,7 +269,7 @@ using FdToArgsFn = std::function<std::vector<std::string>(int fd)>;
 //! It must not rely on child pid/state, and must return the command line
 //! arguments that should be used to execute the process. Embed the remote file
 //! descriptor number in whatever format the child process expects.
-int SpawnProcess(ProcessId& pid, FdToArgsFn&& fd_to_args);
+SocketId SpawnProcess(ProcessId& pid, FdToArgsFn&& fd_to_args);
 
 //! Call execvp with vector args.
 //! Not safe to call in a post-fork child of a multi-threaded process.
--- a/include/mp/util.h
+++ b/include/mp/util.h
@@ -94,6 +94,16 @@ using RemoveCvRef = std::remove_cv_t<std::remove_reference_t<T>>;
 template <typename T>
 using Decay = std::decay_t<T>;
 
+//! Concept satisfied when T's .get() method returns exactly type U.
+//! Used to constrain overloads that handle a specific capnp field type.
+template <typename T, typename U>
+concept FieldTypeIs = std::is_same_v<decltype(std::declval<T>().get()), U>;
+
+//! Concept satisfied when T's .get() method returns a capnp interface type
+//! (i.e., a type that exposes a nested ::Calls type in generated code).
+template <typename T>
+concept InterfaceField = requires { typename Decay<decltype(std::declval<T>().get())>::Calls; };
+
 //! SFINAE helper, see using Require below.
 template <typename SfinaeExpr, typename Result_>
 struct _Require
--- a/src/mp/proxy.cpp
+++ b/src/mp/proxy.cpp
@@ -101,7 +101,7 @@ void EventLoopRef::reset(bool relock) MP_NO_TSA
 
 ProxyContext::ProxyContext(Connection* connection) : connection(connection), loop{*connection->m_loop} {}
 
-Connection::~Connection()
+Connection::~Connection() noexcept(false)
 {
     // Connection destructor is always called on the event loop thread. If this
     // is a local disconnect, it will trigger I/O, so this needs to run on the
@@ -120,6 +120,31 @@ Connection::~Connection()
     // executing. In that case, Cap'n Proto will destroy the ProxyServer objects
     // after the calls finish.
     m_rpc_system.reset();
+
+    // shutdownWrite is needed on Windows so pending data in the m_stream socket
+    // will be sent instead of discarded when m_stream is destroyed. On unix,
+    // this doesn't seem to be needed because data is sent more reliably.
+    //
+    // Sending pending data is important if the connection is a socketpair
+    // because when one side of the socketpair is closed, the other side doesn't
+    // seem to receive any onDisconnect event. So it is important for the other
+    // side to instead receive Cap'n Proto "release" messages (see `struct
+    // Release` in capnp/rpc.capnp) from local Client objects being destroyed so
+    // the remote side can free resources and shut down cleanly. Without this,
+    // when one side of a socket pair is closed the other side may not receive
+    // these messages, preventing the remote side from freeing ProxyServer
+    // resources and shutting down cleanly.
+    try {
+        m_stream->shutdownWrite();
+    } catch (const kj::Exception& e) {
+        // Ignore ENOTCONN: on macOS/FreeBSD (unlike Linux), shutdown(SHUT_WR)
+        // returns ENOTCONN if the peer already closed the connection. This is
+        // expected when the destructor is triggered by a remote disconnect.
+        // Also ignore UNIMPLEMENTED: shutdownWrite() may not be supported on
+        // all stream types (e.g. streams that do not support half-close).
+        if (e.getType() != kj::Exception::Type::DISCONNECTED &&
+            e.getType() != kj::Exception::Type::UNIMPLEMENTED) throw;
+    }
 
     // ProxyClient cleanup handlers are in sync list, and ProxyServer cleanup
     // handlers are in the async list.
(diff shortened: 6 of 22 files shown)


include/mp/util.h
```diff
@@ -271,6 +274,10 @@ using ConnectInfoToArgsFn = std::function<std::vector<std::string>(const Connect
 //! it returns. Returns child process id and socket id.
 std::tuple<ProcessId, SocketId> SpawnProcess(ConnectInfoToArgsFn&& connect_info_to_args);
 
+//! Spawn a process and return its process id. Caller should call WaitProcess
+//! on the returned id.
+ProcessId SpawnProcess(const std::vector<std::string>& args);
```
ViniciusCestarii: Is this declaration proposital? It has no definition anywhere

src/mp/util.cpp
```diff
@@ -176,16 +177,20 @@ SocketId StartSpawned(const ConnectInfo& connect_info)
     return std::stoi(connect_info);
 }
 
-void ExecProcess(const std::vector<std::string>& args)
+ProcessId ExecProcess(const std::vector<std::string>& args)
```
ViniciusCestarii: Now that this function does fork() + execvp() rather than just execvp, is ExecProcess still the intended name?

include/mp/util.h
```diff
@@ -253,17 +253,22 @@ using ProcessId = int;
...
-//! It must not rely on child pid/state, and must return the command line
-//! arguments that should be used to execute the process. Embed the remote file
-//! descriptor number in whatever format the child process expects.
-SocketId SpawnProcess(ProcessId& pid, FdToArgsFn&& fd_to_args);
+//! pair. Calls connect_info_to_args callback with a connection string that
+//! needs to be passed to the child process, and executes the argv command line
+//! it returns. Returns child process id and socket id.
+std::tuple<ProcessId, SocketId> SpawnProcess(ConnectInfoToArgsFn&& connect_info_to_args);
+
+//! Initialize spawned child process using the ConnectInfo string passed to it,
```
ViniciusCestarii: StartSpawned reads as imperative, but the body just parses an int out of the connect-info string (and the header comment seems to describe more work than the function actually does). Would something like ParseConnectInfo fit better? 

I might be missing the full picture here.

ViniciusCestarii: Yes, I was missing the full picture. Just looked [PR 231](https://github.com/bitcoin-core/libmultiprocess/pull/231/changes/4f58c8c981d31cc58d593d70aaff000081589953#diff-cd70e7485d379aa68f4f8a8e77f7706614f158702f634077794a46744040d60eR272-R292)

include/mp/util.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3210462616

In commit "util, refactor: Fix PtrOrValue constructor for move-only types on MSVC" (1060a95de217025f4ea0a7f56a3936aec57834c1)

> Is this declaration proposital? It has no definition anywhere

Good catch. No this declaration is just a mistake accidentally to this commit. Will remove

src/mp/util.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3210769635

In commit "util, refactor: Handle forking inside ExecProcess" (b16f8c4b47ba80088a26e62c5251d0a88b0d45cb)

> Now that this function does fork() + execvp() rather than just execvp, is ExecProcess still the intended name?

Good point, I don't think an ExecProcess is a bad name but it could be misleading, and something like RunProcess could be better.

include/mp/util.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3210828548

> the header comment seems to describe more work than the function actually does

Yes, you linked to the relevant change, but just to answer the question, while on unix this only needs to parse the descriptor number, on windows it needs to do [more](https://github.com/ryanofsky/libmultiprocess/blob/7fd5ec40bc8c2a1fa0e2645d2b587ce2c1c3d17d/src/mp/util.cpp#L276-L293) and read a socket descriptor over a pipe.

ryanofsky (commented):
Thanks for the review! Will update with suggestions.

src/mp/util.cpp
ViniciusCestarii: Looking at this again, it was a pretty minor nit. ExecProcess is a fine name

Sjors:
I opened https://github.com/stratum-mining/sv2-tp/pull/110 to test the changes in the Template Provider.

include/mp/util.h
```diff
@@ -249,6 +249,8 @@ std::string ThreadName(const char* exe_name);
 //! errors in python unit tests.
 std::string LogEscape(const kj::StringTree& string, size_t max_size);
 
+using ProcessId = int;
```
Sjors, line 276: In 36c91a0c73d5fc5ac616c43ed2a0fdfdea115b81 _util, refactor: Add ProcessId type alias and use it_: the commit message could mention that Windows uses a different underlying type.

include/mp/util.h
```diff
@@ -250,9 +250,11 @@ std::string ThreadName(const char* exe_name);
 std::string LogEscape(const kj::StringTree& string, size_t max_size);
 
 using ProcessId = int;
+using SocketId = int;
+constexpr SocketId SocketError{-1};
```
Sjors, line 278: In 94af41bb55b773787ad1cf195c3b65a68568a1f6 _util, refactor: Add SocketId type alias and use it_: `SocketError` is unused in this commit. Maybe do:

```diff
diff --git a/include/mp/proxy-io.h b/include/mp/proxy-io.h
index d7b9f0e546..cae4ac08e9 100644
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -310,8 +310,8 @@ public:

     //! Pipe read handle used to wake up the event loop thread.
-    int m_wait_fd = -1;
+    SocketId m_wait_fd = SocketError;

     //! Pipe write handle used to wake up the event loop thread.
-    int m_post_fd = -1;
+    SocketId m_post_fd = SocketError;

     //! Number of clients holding references to ProxyServerBase objects that
diff --git a/src/mp/proxy.cpp b/src/mp/proxy.cpp
index eb0e841291..4783e1f110 100644
--- a/src/mp/proxy.cpp
+++ b/src/mp/proxy.cpp
@@ -217,6 +217,6 @@ EventLoop::~EventLoop()
     KJ_ASSERT(m_post_fn == nullptr);
     KJ_ASSERT(!m_async_fns);
-    KJ_ASSERT(m_wait_fd == -1);
-    KJ_ASSERT(m_post_fd == -1);
+    KJ_ASSERT(m_wait_fd == SocketError);
+    KJ_ASSERT(m_post_fd == SocketError);
     KJ_ASSERT(m_num_clients == 0);

@@ -269,6 +269,6 @@ void EventLoop::loop()
     KJ_SYSCALL(::close(post_fd));
     const Lock lock(m_mutex);
-    m_wait_fd = -1;
-    m_post_fd = -1;
+    m_wait_fd = SocketError;
+    m_post_fd = SocketError;
     m_async_fns.reset();
     m_cv.notify_all();
```

Although IIUC this will go away in a later commit.

Sjors (commented):
Studied two first three commits...

include/mp/util.h
```diff
@@ -253,17 +253,22 @@ using ProcessId = int;
 using SocketId = int;
 constexpr SocketId SocketError{-1};
 
+//! Information about parent process passed to child process as a command-line
+//! argument. On unix this is the child socket fd number formatted as a string.
```
Sjors, line 281: In beaa50a046190277a91cdf9ca87c169fd2516c55 _util, refactor: Add ConnectInfo type alias and use it_: maybe call it `SpawnConnectInfo`?

include/mp/util.h
```diff
@@ -270,10 +270,9 @@ std::tuple<ProcessId, SocketId> SpawnProcess(ConnectInfoToArgsFn&& connect_info_
 //! returning a socket id for communicating with the parent process.
 SocketId StartSpawned(const ConnectInfo& connect_info);
 
-//! Call execvp with vector args.
-//! Not safe to call in a post-fork child of a multi-threaded process.
```
Sjors, line 271: In b16f8c4b47ba80088a26e62c5251d0a88b0d45cb _util, refactor: Handle forking inside ExecProcess_: I think you still need some sort of warning because the `execvp failed` branch in the child is not [async-signal-safe](https://man7.org/linux/man-pages/man7/signal-safety.7.html).

include/mp/util.h
```diff
@@ -270,10 +270,9 @@ std::tuple<ProcessId, SocketId> SpawnProcess(ConnectInfoToArgsFn&& connect_info_
...
-//! Currently only used by mpgen at build time.
```
Sjors, line 272: In b16f8c4 util, refactor: Handle forking inside ExecProcess: this comment shouldn't be dropped, as it's a warning against using this in run time without dealing get signal-safety issue above first.

src/mp/util.cpp
```diff
@@ -119,10 +119,7 @@ std::string LogEscape(const kj::StringTree& string, size_t max_size)
 
 std::tuple<ProcessId, SocketId> SpawnProcess(ConnectInfoToArgsFn&& connect_info_to_args)
 {
-    SocketId fds[2];
-    if (socketpair(AF_UNIX, SOCK_STREAM, 0, fds) != 0) {
-        throw std::system_error(errno, std::system_category(), "socketpair");
```
Sjors, line 123: In 022b29b776759a246296cad578fbde97d47149a3 _util, refactor: Add SocketPair() and use it in SpawnProcess_: changing `throw std::system_error` and `!= 0` check, to `KJ_SYSCALL` seems fine, but still a slight behavior change?

src/mp/util.cpp
```diff
@@ -158,6 +159,16 @@ std::tuple<ProcessId, SocketId> SpawnProcess(ConnectInfoToArgsFn&& connect_info_
             }
         }
 
+        // Explicitly clear FD_CLOEXEC on the child's socket before calling
+        // exec, so the fd survives into the spawned process regardless of how
+        // the socket was created.
+        int flags = fcntl(fds[0], F_GETFD);
+        if (flags == -1) throw std::system_error(errno, std::system_category(), "fcntl F_GETFD");
```
Sjors: In 24c5e57fddec43c3e46fff36f93fe5f9f35d05bd _util: Clear FD_CLOEXEC on child socket before exec_: to not make things more non-fork-safe:

```diff
diff --git a/src/mp/util.cpp b/src/mp/util.cpp
index 15400215e4..b2a545ef5f 100644
--- a/src/mp/util.cpp
+++ b/src/mp/util.cpp
@@ -62,4 +62,12 @@ size_t MaxFd()
 }

+template <size_t N>
+[[noreturn]] void ChildProcessError(const char (&msg)[N])
+{
+    const ssize_t writeResult = ::write(STDERR_FILENO, msg, N - 1);
+    (void)writeResult;
+    _exit(126);
+}
+
 } // namespace

@@ -144,8 +152,5 @@ std::tuple<ProcessId, SocketId> SpawnProcess(ConnectInfoToArgsFn&& connect_info_
             throw std::system_error(errno, std::system_category(), "close");
         }
-        static constexpr char msg[] = "SpawnProcess(child): close(fds[1]) failed\n";
-        const ssize_t writeResult = ::write(STDERR_FILENO, msg, sizeof(msg) - 1);
-        (void)writeResult;
-        _exit(126);
+        ChildProcessError("SpawnProcess(child): close(fds[1]) failed\n");
     }

@@ -164,8 +169,12 @@ std::tuple<ProcessId, SocketId> SpawnProcess(ConnectInfoToArgsFn&& connect_info_
         // the socket was created.
         int flags = fcntl(fds[0], F_GETFD);
-        if (flags == -1) throw std::system_error(errno, std::system_category(), "fcntl F_GETFD");
+        if (flags == -1) {
+            ChildProcessError("SpawnProcess(child): fcntl(fds[0], F_GETFD) failed\n");
+        }
         if (flags & FD_CLOEXEC) {
             flags &= ~FD_CLOEXEC;
-            if (fcntl(fds[0], F_SETFD, flags) == -1) throw std::system_error(errno, std::system_category(), "fcntl F_SETFD");
+            if (fcntl(fds[0], F_SETFD, flags) == -1) {
+                ChildProcessError("SpawnProcess(child): fcntl(fds[0], F_SETFD) failed\n");
+            }
         }
```

Sjors: Was there something in particular that led you this (hardening) commit?

include/mp/util.h
```diff
@@ -250,6 +252,8 @@ std::string ThreadName(const char* exe_name);
 //! errors in python unit tests.
 std::string LogEscape(const kj::StringTree& string, size_t max_size);
 
+using Stream = kj::Own<kj::AsyncIoStream>;
```
Sjors, line 274: In 091f5e16dcd9f08d92aafbae4feee03566d0041e _proxy, refactor: Change ConnectStream and ServeStream to accept stream objects_: would it make sense to introduce this earlier, in 3c81cf27ead7add93c69a86ffb97bb51b9fe799a?

include/mp/proxy-io.h
```diff
@@ -211,6 +211,18 @@ class Logger
 
 std::string LongThreadName(const char* exe_name);
 
+inline SocketId StreamSocketId(const Stream& stream)
```
Sjors: In 7cb83a5d53476585c12dccabf84cc56294f64c5e _cmake: Fix CapnProto tool paths broken by Ubuntu Noble packaging bug_: `StreamSocketId` is unused and stays that way? Given that it's potentially unsafe to throw, maybe it's better to drop this?

src/mp/proxy.cpp
```diff
@@ -99,6 +99,28 @@ Connection::~Connection()
...
+    // seem to receive any onDisconnect event. So it is important for the other
+    // side to instead receive Cap'n Proto "release" messages (see `struct
+    // Release` in capnp/rpc.capnp) from local Client objects being destroyed so
+    // the remote side can free resources and shut down cleanly. Without this,
+    // when one side of a socket pair is closed the other side may not receive
+    // these messages, preventing the remote side from freeing ProxyServer
+    // resources and shutting down cleanly.
+    try {
+        m_stream->shutdownWrite();
+    } catch (const kj::Exception& e) {
```
Sjors: In bfc2db7b517e59e2413b8920341d320bb1228dd2 _proxy: Call shutdownWrite() in Connection destructor_: maybe squash 28e4c7fd2eb992d632a3b394eab30d3e236208ca into this?

include/mp/util.h
Sjors: In 1060a95de217025f4ea0a7f56a3936aec57834c1 _util, refactor: Fix PtrOrValue constructor for move-only types on MSVC_: this seems unrelated?

Sjors (commented):
Reviewed the rest as well, except the CMake changes in c9aa8060ec2a7b91d88bf5516320bfa41b14d4a5 and 7cb83a5d53476585c12dccabf84cc56294f64c5e.

Concept ACK

include/mp/util.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3210462616

Removed now

src/mp/util.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3210769635

> Looking at this again, it was a pretty minor nit. ExecProcess is a fine name

Wound up renaming to StartProcess, which I think should be an improvement

include/mp/util.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3429153726

> In [36c91a0](https://github.com/bitcoin-core/libmultiprocess/commit/36c91a0c73d5fc5ac616c43ed2a0fdfdea115b81) _util, refactor: Add ProcessId type alias and use it_: the commit message could mention that Windows uses a different underlying type.

Thanks, added this to commit message

include/mp/util.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3429261166

> In [94af41b](https://github.com/bitcoin-core/libmultiprocess/commit/94af41bb55b773787ad1cf195c3b65a68568a1f6) _util, refactor: Add SocketId type alias and use it_: `SocketError` is unused in this commit. Maybe do:

Thanks, applied patch!

> Although IIUC this will go away in a later commit.

Also true, but seems nice to be complete here

include/mp/util.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3429639471

> In [beaa50a](https://github.com/bitcoin-core/libmultiprocess/commit/beaa50a046190277a91cdf9ca87c169fd2516c55) _util, refactor: Add ConnectInfo type alias and use it_: maybe call it `SpawnConnectInfo`?

Makes sense, renamed

include/mp/util.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3435427603

> In [b16f8c4](https://github.com/bitcoin-core/libmultiprocess/commit/b16f8c4b47ba80088a26e62c5251d0a88b0d45cb) _util, refactor: Handle forking inside ExecProcess_: I think you still need some sort of warning because the `execvp failed` branch in the child is not [async-signal-safe](https://man7.org/linux/man-pages/man7/signal-safety.7.html).

Agree there is an issue here, although the warning doesn't really help because it is telling the caller to do something it has no control over, and the function itself is breaking the safety rule internally. Fixed by switching to posix_spawn.

include/mp/util.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3435465779

> In [b16f8c4](https://github.com/bitcoin-core/libmultiprocess/commit/b16f8c4b47ba80088a26e62c5251d0a88b0d45cb) util, refactor: Handle forking inside ExecProcess: this comment shouldn't be dropped, as it's a warning against using this in run time without dealing get signal-safety issue above first.

Note this review comment seems to be on same lines of code as https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3435427603, so I think this is a duplicate suggestion

src/mp/util.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3435633212

> In [022b29b](https://github.com/bitcoin-core/libmultiprocess/commit/022b29b776759a246296cad578fbde97d47149a3) _util, refactor: Add SocketPair() and use it in SpawnProcess_: changing `throw std::system_error` and `!= 0` check, to `KJ_SYSCALL` seems fine, but still a slight behavior change?

Thanks, noted in commit message

src/mp/util.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3435703266

> In [24c5e57](https://github.com/bitcoin-core/libmultiprocess/commit/24c5e57fddec43c3e46fff36f93fe5f9f35d05bd) _util: Clear FD_CLOEXEC on child socket before exec_: to not make things more non-fork-safe:

Good catch. An easier to way to make it safe was just to move the CLOEXEC code before the fork() call so now doing that

ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3435794870

> Was there something in particular that led you this (hardening) commit?

I don't remember exactly. It was definitely a bug of some kind. I think an earlier version of the code was using `newTwoWayPipe` which wraps a socket pair and sets CLOEXEC. Also originally `SpawnProcess` required callers to construct the socket pair, taking the child socket as an argument instead of both sockets. But this made calling code more complicated, so now `SpawnProcess` returns a parent socket instead of accepting a child socket.

Now since `SpawnProcess` is creating the socket pair it knows whether CLOEXEC is set so this code is not necessary. But I think it's good to use CLOEXEC to prevent accidental leaks so I did keep some of this and integrated it better with the earlier commit introducing SocketPair.

include/mp/util.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3437759806

> In [091f5e1](https://github.com/bitcoin-core/libmultiprocess/commit/091f5e16dcd9f08d92aafbae4feee03566d0041e) _proxy, refactor: Change ConnectStream and ServeStream to accept stream objects_: would it make sense to introduce this earlier, in [3c81cf2](https://github.com/bitcoin-core/libmultiprocess/commit/3c81cf27ead7add93c69a86ffb97bb51b9fe799a)?

It could make sense but I feel it's a little clearer if the EventLoop code is using `kj::Own<kj::AsyncIoStream>` and `kj::Own<kj::OutputStream>` types directly and not tied to the `mp::Stream` type for external callers and meant to be more opaque.

But I did extend this commit to use `Stream` stream type in ConnectStream and ServeStream functions since these are external functions.

include/mp/proxy-io.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3437771873

> In [7cb83a5](https://github.com/bitcoin-core/libmultiprocess/commit/7cb83a5d53476585c12dccabf84cc56294f64c5e) _cmake: Fix CapnProto tool paths broken by Ubuntu Noble packaging bug_: `StreamSocketId` is unused and stays that way? Given that it's potentially unsafe to throw, maybe it's better to drop this?

Now removed. I think this was used previously when code was using `newTwoWayPipe` to create streams and using `StreamSocketId` to turn them into socket ids. Now it does the opposite calling `SocketPair` starting from socket ids, and calling `MakeStream` to turn them into streams.

Probably will not go back to the previous approach since it made SpawnProcess more complicated as described https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3435794870, so ok to drop `StreamSocketId`.

There are some reasons why `StreamSocketId` might be useful: it is an inverse to the `MakeStream` function, and could be used in the `EventLoop` constructor to assign `m_post_writer`. But there isn't a need for it now.

src/mp/proxy.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3437864426

> In [bfc2db7](https://github.com/bitcoin-core/libmultiprocess/commit/bfc2db7b517e59e2413b8920341d320bb1228dd2) _proxy: Call shutdownWrite() in Connection destructor_: maybe squash [28e4c7f](https://github.com/bitcoin-core/libmultiprocess/commit/28e4c7fd2eb992d632a3b394eab30d3e236208ca) into this?

Didn't squash these because I think 28e4c7fd2eb992d632a3b394eab30d3e236208ca is good documentation of the macos REDUCE_EXPORTS bug and will be probably be useful to reference in the future (the bug can happen very easily just by throwing and catching exceptions.)

include/mp/util.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3437888592

> In [1060a95](https://github.com/bitcoin-core/libmultiprocess/commit/1060a95de217025f4ea0a7f56a3936aec57834c1) _util, refactor: Fix PtrOrValue constructor for move-only types on MSVC_: this seems unrelated?

Thanks, dropped this

ryanofsky (commented):
Thanks for the reviews!


Rebased 7cb83a5d53476585c12dccabf84cc56294f64c5e -> 68ed129cc3c5803323ea75df28c806cd49f64710 ([`pr/wins.1`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.1) -> [`pr/wins.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.2), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/wins.1-rebase..pr/wins.2)) implementing review suggestions and fixing conflict with #279

Updated 68ed129cc3c5803323ea75df28c806cd49f64710 -> a43e5a85bbf8f5551e41be8873c2c16b94ecd4c4 ([`pr/wins.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.2) -> [`pr/wins.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.3), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/wins.2..pr/wins.3)) to fix environ undeclared on macOS/BSD: add explicit declaration before posix_spawn() https://github.com/bitcoin-core/libmultiprocess/actions/runs/27974097183/job/82787313836

Updated a43e5a85bbf8f5551e41be8873c2c16b94ecd4c4 -> 6cc729f9312641c0299b774228f89494dd85494a ([`pr/wins.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.3) -> [`pr/wins.4`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.4), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/wins.3..pr/wins.4)) to fix clang-tidy readability-redundant-declaration on environ declaration https://github.com/bitcoin-core/libmultiprocess/actions/runs/27978804911/job/82803323635

src/mp/util.cpp
```diff
@@ -176,16 +176,14 @@ SocketId StartSpawned(const SpawnConnectInfo& connect_info)
...
 {
     const std::vector<char*> argv{MakeArgv(args)};
-    if (execvp(argv[0], argv.data()) != 0) {
-        perror("execvp failed");
-        if (errno == ENOENT && !args.empty()) {
-            std::cerr << "Missing executable: " << fs::weakly_canonical(args.front()) << '\n';
-        }
-        _exit(1);
+    ProcessId pid;
+    if (int err = posix_spawn(&pid, argv[0], nullptr, nullptr, argv.data(), ::environ)) {
```
Sjors: In ddd3945be422d3b2addd2b25778eaff01ba41923 _util, refactor: Do not fork() and exec() separately_: not `posix_spawnp`?

include/mp/util.h
```diff
@@ -276,6 +277,10 @@ std::tuple<ProcessId, SocketId> SpawnProcess(SpawnConnectInfoToArgsFn&& connect_
 //! returning a socket id for communicating with the parent process.
 SocketId StartSpawned(const SpawnConnectInfo& connect_info);
 
+//! Create a socket pair that can be used to communicate within a process or
+//! between parent and child processes.
+std::array<SocketId, 2> SocketPair();
```
Sjors, line 299: In d9a5a68872167dd1bb0412d1174dcee694182527 _util, refactor: Add SocketPair() and use it in SpawnProcess_: it seems a bit odd to set `FD_CLOEXEC` and then unset it again for the child. The unset code also assume it was the only flag, since it sets `0`. Maybe it's better to give this function a boolean argument for whether to set this flag on the child?

Sjors (commented):
ACK 6cc729f9312641c0299b774228f89494dd85494a

I still didn't review the cmake commits though.

include/mp/proxy-io.h
enirox001: In 0ec1534 ("util, refactor: Add SocketId type alias and use it"):

Should this parameter be using `SocketId` instead of `int` to maintain type consistency with the rest of the code?

```diff --git a/include/mp/proxy-io.h b/include/mp/proxy-io.h
index 91d02cc..29bcb54 100644
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -876,14 +876,14 @@ void ServeStream(EventLoop& loop, Stream stream, InitImpl& init)
     _Serve<InitInterface>(loop, kj::mv(stream), init);
 }

-//! Given listening socket file descriptor and an init object, handle incoming
+//! Given listening socket identifier and an init object, handle incoming
 //! connections and requests by calling methods on the Init object.
 template <typename InitInterface, typename InitImpl>
-void ListenConnections(EventLoop& loop, int fd, InitImpl& init)
+void ListenConnections(EventLoop& loop, SocketId socket_id, InitImpl& init)
 {
     loop.sync([&]() {
         _Listen<InitInterface>(loop,
-            loop.m_io_context.lowLevelProvider->wrapListenSocketFd(fd, kj::LowLevelAsyncIoProvider::TAKE_OWNERSHIP),
+            loop.m_io_context.lowLevelProvider->wrapListenSocketFd(socket_id, kj::LowLevelAsyncIoProvider::TAKE_OWNERSHIP),
             init);
     });
 }
```

example/calculator.cpp
```diff
@@ -51,14 +50,10 @@ int main(int argc, char** argv)
         std::cout << "Usage: mpcalculator <fd>\n";
         return 1;
     }
-    mp::SocketId fd;
-    if (std::from_chars(argv[1], argv[1] + strlen(argv[1]), fd).ec != std::errc{}) {
-        std::cerr << argv[1] << " is not a number or is larger than an int\n";
-        return 1;
-    }
+    mp::SocketId socket{mp::StartSpawned(argv[1])};
```
enirox001, line 53: In c7a939b _"util, refactor: Add SpawnConnectInfo type alias and use it"_

Didn't we lose our error handling here? `std::stoi` throws exceptions on bad input, so shouldn't we add a try-catch to prevent a crash?

include/mp/util.h
```diff
@@ -255,6 +255,8 @@ std::string ThreadName(const char* exe_name);
 //! errors in python unit tests.
 std::string LogEscape(const kj::StringTree& string, size_t max_size);
 
+using ProcessId = int;
```
enirox001, line 276: in 5f578bb _"util, refactor: Add ProcessId type alias and use it"_

nit: Since this refactor adds `ProcessId`, should we also add `ProcessError{-1}` and use it for the fork() failure check, matching `SocketId` and `SocketError`

```diff --git a/include/mp/util.h b/include/mp/util.h
index f647d5d..05e1b4b 100644
--- a/include/mp/util.h
+++ b/include/mp/util.h
@@ -276,6 +276,7 @@ using Stream = kj::Own<kj::AsyncIoStrea
m>;
 using ProcessId = int;
 using SocketId = int;
 constexpr SocketId SocketError{-1};
+constexpr ProcessId ProcessError{-1};
 
 //! Information about parent process passed to child proce
ss as a command-line
 //! argument. On unix this is the child socket fd number f
ormatted as a string.
diff --git a/src/mp/util.cpp b/src/mp/util.cpp
index 1721e12..342acc8 100644
--- a/src/mp/util.cpp
+++ b/src/mp/util.cpp
@@ -134,7 +134,7 @@ std::tuple<ProcessId, SocketId> SpawnPr
ocess(SpawnConnectInfoToArgsFn&& connect_
     KJ_SYSCALL(fcntl(fds[0], F_SETFD, 0));
 
     ProcessId pid = fork();
-    if (pid == -1) {
+    if (pid == ProcessError) {
         throw std::system_error(errno, std::system_categor
y(), "fork");
     }
     // Parent process closes the descriptor for socket 0, 
child closes the
```

src/mp/proxy.cpp
```diff
@@ -67,10 +66,9 @@ void EventLoopRef::reset(bool relock) MP_NO_TSA
         loop->m_num_clients -= 1;
         if (loop->done()) {
             loop->m_cv.notify_all();
-            int post_fd{loop->m_post_fd};
             loop_lock->unlock();
             char buffer = 0;
-            KJ_SYSCALL(write(post_fd, &buffer, 1)); // NOLINT(bugprone-suspicious-semicolon)
+            loop->m_post_writer->write(&buffer, 1);
```
enirox001: In 85e13a6 _"proxy, refactor: Replace EventLoop wakeup fd integers with KJ stream objects"_

Before this refactor, this method copied `m_post_fd` while holding `m_mutex` and then wrote to the local fd after unlocking. 

Now it unlocks and then dereferences `loop->m_post_writer`. Is that safe if the event loop wakes up and tears down its streams while reset() is still executing? 

Perhaps we could keep a local writer before unlocking?

src/mp/proxy.cpp
```diff
@@ -266,10 +266,9 @@ void EventLoop::loop()
     m_task_set.reset();
     MP_LOG(*this, Log::Info) << "EventLoop::loop bye.";
     wait_stream = nullptr;
-    KJ_SYSCALL(::close(post_fd));
     const Lock lock(m_mutex);
-    m_wait_fd = SocketError;
-    m_post_fd = SocketError;
+    m_wait_stream = nullptr;
+    m_post_stream = nullptr;
```
enirox001, line 326: In 85e13a6 "proxy, refactor: Replace EventLoop wakeup fd integers with KJ stream objects"

Should `m_post_writer` also be reset during EventLoop teardown? It is initialized from `m_post_stream`, but the loop mehtod currently clears `m_wait_stream` and `m_post_stream` without clearing `m_post_writer`. I was thinking something along the lines of:

```diff --git a/src/mp/proxy.cpp b/src/mp/proxy.cpp
index c7b602e..e14455c 100644
--- a/src/mp/proxy.cpp
+++ b/src/mp/proxy.cpp
@@ -298,6 +298,7 @@ void EventLoop::loop()
     const Lock lock(m_mutex);
     m_wait_stream = nullptr;
     m_post_stream = nullptr;
+    m_post_writer = nullptr;
     m_async_fns.reset();
     m_cv.notify_all();
 }
```

include/mp/util.h
enirox001: In 174604a _"proxy, refactor: Change ConnectStream and ServeStream to accept stream objects"_

Had a concern that the name `Stream` is fairly broad, so using it too widely could hide useful information at call sites. For example, inside `EventLoop` it seems helpful to see the concrete KJ type details directly, otherwise future code might start using `mp::Stream` everywhere just because the alias exists.

I agree with keeping it limited to external facing methods like `ConnectStream` and `ServeStream`, where the abstraction is useful, and avoiding it in internal code where the underlying type is more informative.

src/mp/proxy.cpp
```diff
@@ -99,6 +99,28 @@ Connection::~Connection()
...
+    // when one side of a socket pair is closed the other side may not receive
+    // these messages, preventing the remote side from freeing ProxyServer
+    // resources and shutting down cleanly.
+    try {
+        m_stream->shutdownWrite();
+    } catch (const kj::Exception& e) {
+        // Ignore ENOTCONN: on macOS/FreeBSD (unlike Linux), shutdown(SHUT_WR)
+        // returns ENOTCONN if the peer already closed the connection. This is
+        // expected when the destructor is triggered by a remote disconnect.
+        if (e.getType() != kj::Exception::Type::DISCONNECTED) throw;
```
enirox001: In c5faf68 _"proxy: Call shutdownWrite() in Connection destructor"_

Is it a good idea to allow exceptions to escape from a destructor here?

Also, what happens if `Connection` wraps a stream that does not support `shutdownWrite()`, wouldn’t this try to perform an operation that isn't permitted and potentially terminate the process?

I think this should either only call shutdownWrite() for streams where it is valid, or catch/ignore the expected failure in the destructor

enirox001 (commented):
Code Review ACK c5faf68

Initial look at the chnages implemented here

Left a few suggestions and questions inline.

src/mp/util.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3461229472

> In [ddd3945](https://github.com/bitcoin-core/libmultiprocess/commit/ddd3945be422d3b2addd2b25778eaff01ba41923) _util, refactor: Do not fork() and exec() separately_: not `posix_spawnp`?

Good catch! Switched to posix_spawnp here (so `PATH`) is still used

include/mp/util.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3461362488

> In [d9a5a68](https://github.com/bitcoin-core/libmultiprocess/commit/d9a5a68872167dd1bb0412d1174dcee694182527) _util, refactor: Add SocketPair() and use it in SpawnProcess_: it seems a bit odd to set `FD_CLOEXEC` and then unset it again for the child. The unset code also assume it was the only flag, since it sets `0`. Maybe it's better to give this function a boolean argument for whether to set this flag on the child?

Good catch on assuming no other flags were set, this should be fixed now by calling F_GETFD. It still seems a little less error prone to explicitly unset FD_CLOEXEC right before forking, instead of adding an option and treating the socketpair fds differently

include/mp/proxy-io.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3534942571

Good catch, added SocketId here

example/calculator.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3537681496

> Didn't we lose our error handling here? `std::stoi` throws exceptions on bad input, so shouldn't we add a try-catch to prevent a crash?

Yes previous error handling provided more context. Changed to catch and throw a more descriptive exception. It can be up to callers whether to crash in this case (which probably indicates a bug not a runtime error) and crashing in the examples seems ok

include/mp/util.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3545410258

IMO it wouldn't make sense to a ProcessError constant because this constant would only be used in posix code where -1 is clearer. `SocketError` is different because it actually needs to be used by cross-platform code and I think the value is also slightly different in windows and posix implementations (unsigned vs signed).

src/mp/proxy.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3545433625

> Now it unlocks and then dereferences `loop->m_post_writer`. Is that safe if the event loop wakes up and tears down its streams while reset() is still executing?

This a good catch so added comments here and restored the previous pattern since the issues are a little subtle.

Dereferencing `m_post_writer` should be safe because the loop can't exit until the write happens. Added a comment describing this. 

Dereferencing `loop` should also be safe because the m_post_writer member also won't be changed (when it is set to null) until after the write() call happens so that should also be safe.

But IIRC the previous pattern of copying members out of m_loop while holding the mutex was needed to prevent false positives from TSAN, which could complain about members being written after they were read without synchronization between, because it was not aware of the synchronization provided by socketpair reads and writes.

Now comments are added explaining most of this.

src/mp/proxy.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3545453602

Thanks, it does seem better to clear m_post_writer here for completeness, so added this

include/mp/util.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3545471249

Agreed, glad the reasoning made sense!

src/mp/proxy.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3545563225

> In [c5faf68](https://github.com/bitcoin-core/libmultiprocess/commit/c5faf68d371d3d7d84e5b89743073924022716ac) _"proxy: Call shutdownWrite() in Connection destructor"_
> 
> Is it a good idea to allow exceptions to escape from a destructor here?

I think probably the answer to this is yes since the pattern is used consistently in capnproto code https://github.com/capnproto/capnproto/blob/v2/kjdoc/tour.md#exceptions-in-destructors and this seems better than alternatives of suppressing errors or terminating. 

But you're right libmultiprocess isn't doing a great job dealing with these exceptions in many cases, and in this case noexcept(false) is needed here so have added that.

Also for defensiveness have added a check to handle the case where shutdownWrite() is not implemented, even though for all stream types we support (socketpairs, UNIX sockets, and TCP sockets) shutdownwrite and half-closes are supported

ryanofsky (commented):
Thanks for the reviews! I dropped the cmake commits here since they require a different type of review and would make more sense to review separately.


Rebased 6cc729f9312641c0299b774228f89494dd85494a -> 56220f87a55becd88904f4ba9eace9bb34a4bdf4 ([`pr/wins.4`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.4) -> [`pr/wins.5`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.5), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/wins.4-rebase..pr/wins.5)) due to conflicts, also applying review suggestions and dropping the cmake commits

Updated 56220f87a55becd88904f4ba9eace9bb34a4bdf4 -> 491f1e80ea1b20ab549607e71ca0cf661ea0ffb7 ([`pr/wins.5`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.5) -> [`pr/wins.6`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.6), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/wins.5..pr/wins.6)) fixing comments and commit organization

Rebased 491f1e80ea1b20ab549607e71ca0cf661ea0ffb7 -> 254c3e36372f0b53ed625d6e0b1d948df130da64 ([`pr/wins.6`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.6) -> [`pr/wins.7`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.7), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/wins.6-rebase..pr/wins.7)) due to conflict with #306

ryanofsky:
@Sjors @enirox001 Are you able to Re-ACK this? I'd like to merge it to unblock windows support. This PR and https://github.com/bitcoin/bitcoin/pull/35084 contain most of the code changes needed to support windows, with the other changes just being non-invasive `#ifdef WIN32` additions.

There are not many changes since your last reviews. They are:

- Rebasing
- Dropping cmake commits
- Catching `std::stio` exception for more descriptive spawn error messages
- Catching another exception (in `~Connection` and declaring it `noexcept(false)`. These changes just precautionary, the new exception and other exceptions in general are not expected here.
- In `EventLoopRef::reset` adding comment and tweaking `post_writer` call to avoid potential TSAN error (not a real race).
- Tweaking `F_SETFD` call to preserve previous flags
- Changing `posix_spawn` to `posix_spawnp` to preserve previous PATH search behavior.

src/mp/util.cpp
```diff
@@ -181,16 +181,14 @@ SocketId StartSpawned(const SpawnConnectInfo& connect_info)
...
 {
     const std::vector<char*> argv{MakeArgv(args)};
-    if (execvp(argv[0], argv.data()) != 0) {
-        perror("execvp failed");
-        if (errno == ENOENT && !args.empty()) {
-            std::cerr << "Missing executable: " << fs::weakly_canonical(args.front()) << '\n';
-        }
-        _exit(1);
+    ProcessId pid;
+    if (int err = posix_spawnp(&pid, argv[0], nullptr, nullptr, argv.data(), ::environ)) {
```
enirox001, line 200: nit: the commit message still references `posix_spawn`

doc/versions.md
```diff
@@ -7,9 +7,12 @@ Library versions are tracked with simple
 Versioning policy is described in the [version.h](../include/mp/version.h)
 include.
 
-## v13
+## v14
 - Current unstable version.
 
+## [v13.0](https://github.com/bitcoin-core/libmultiprocess/commits/v13.0)
+- Adds support for nonunix platforms, making API changes that are not backwards compatible.
```
Sjors: Did you mean v14 here?

include/mp/proxy-io.h
```diff
@@ -214,6 +214,12 @@ class Logger
 
 std::string LongThreadName(const char* exe_name);
 
+//! Wrap a socket file descriptor as an async stream, taking ownership of the fd.
+inline Stream MakeStream(kj::AsyncIoContext& io_context, SocketId socket)
+{
+    return io_context.lowLevelProvider->wrapSocketFd(socket, kj::LowLevelAsyncIoProvider::TAKE_OWNERSHIP);
```
Sjors: Claude warned about this:

> WARNING: A UnixEventPort can only be used in the thread and process that created it.

https://github.com/capnproto/capnproto/blob/v0.9.2/c++/src/kj/async-unix.h#L64

And now we're creating this outside the event-loop thread.

Image

Sjors:
Everything else in 254c3e36372f0b53ed625d6e0b1d948df130da64 looks good to me, just two questions above.

doc/versions.md
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3595386053

> Did you mean v14 here?

Thanks good catch, I forgot to update this after https://github.com/bitcoin/bitcoin/pull/35720. (This is one reason I really want to move documentation to a support branch: so the project can actually have good documentation that is updated freely and not tangled in with code changes.)

ci/configs/olddeps.bash
```diff
@@ -1,5 +1,5 @@
 CI_DESC="CI job using old Cap'n Proto and cmake versions"
 CI_DIR=build-olddeps
 export CXXFLAGS="-Werror -Wall -Wextra -Wpedantic -Wno-unused-parameter -Wno-error=array-bounds"
-NIX_ARGS=(--argstr capnprotoVersion "0.7.1" --argstr cmakeVersion "3.12.4")
+NIX_ARGS=(--argstr capnprotoVersion "0.9.2" --argstr cmakeVersion "3.12.4")
```
enirox001, line 4: Unsure, but should this commit not come before a4a680c0eaf6a5403bc9118095026e333392ee6b? Seeing that `kj::AsyncIoStream::getFd()` was first used in that commit

include/mp/proxy-io.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3595512630

> And now we're creating this outside the event-loop thread.

Yes this is a pretty good catch. It mainly affects https://github.com/bitcoin/bitcoin/pull/35084 not this PR because only the downsteam PR real calls MakeStream.

This PR only uses MakeStream in a test where it is called from the EventLoop thread which should be ok, and from an example where it is called from a different thread, which could have a theoretical problem, but still seems to work.

I think I will change MakeStream call to take a EventLoop parameter and call sync to avoid this problem. I think probably it's only a theoretical problem because just wrapping a file descriptor should not affect shared i/o state. But worth fixing regardless

enirox001 (commented):
ACK 254c3e3 Changes look fine to me. Left a few questions, but they are non-blocking nits.

ryanofsky (commented):
Updated 254c3e36372f0b53ed625d6e0b1d948df130da64 -> 6e29e2279d020a31eab5a9f533ce978368a2a551 ([`pr/wins.7`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.7) -> [`pr/wins.8`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.8), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/wins.7..pr/wins.8)) fixing version documentation

include/mp/proxy-io.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3595512630

I looked into this more and wrapSocketFd doesn't currently mutate shared state on any platform, so there should not be a practical issue currently. But to future proof this I changed MakeStream to have a different signature and call wrapSocketFd on the event loop thread. Nice catch and thanks for posting this!

ryanofsky (commented):
Updated 6e29e2279d020a31eab5a9f533ce978368a2a551 -> 995398123f270cd5bb5d2cf794ef3cce905efb48 ([`pr/wins.8`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.8) -> [`pr/wins.9`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.9), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/wins.8..pr/wins.9)) changing MakeStream signature to avoid potential thread safety issues in the future. Note that this requires an update to https://github.com/bitcoin/bitcoin/pull/35084 so CI will fail until that is pushed (should happen shortly)

Squashed 995398123f270cd5bb5d2cf794ef3cce905efb48 -> 1b0f605606254be1612e9d1b918cd8b860457372 ([`pr/wins.9`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.9) -> [`pr/wins.10`](https://github.com/ryanofsky/libmultiprocess/commits/pr/wins.10), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/wins.9..pr/wins.10)) just updating commit message and commit order as suggested

src/mp/util.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3595251053

> nit: the commit message still references `posix_spawn`

Thanks, updated to posix_spawnp

ci/configs/olddeps.bash
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3595593282

> Unsure, but should this commit not come before [a4a680c](https://github.com/bitcoin-core/libmultiprocess/commit/a4a680c0eaf6a5403bc9118095026e333392ee6b)? Seeing that `kj::AsyncIoStream::getFd()` was first used in that commit

This was intended because I thought it would make the commit sequence a little easier to understand but it also makes sense to order based on strict dependencies, so have reordered now

Sjors:
ACK 1b0f605606254be1612e9d1b918cd8b860457372

ViniciusCestarii:
ACK 1b0f605606254be1612e9d1b918cd8b860457372 I like that these changes make the code more readable too, require clauses and KJ Stream abstractions are nice

src/mp/proxy.cpp
```diff
@@ -258,7 +269,7 @@ void EventLoop::loop()
             m_cv.notify_all();
         } else if (done()) {
             // Intentionally do not break if m_post_fn was set, even if done()
-            // would return true, to ensure that the EventLoopRef write(post_fd)
+            // would return true, to ensure that the post() m_post_writer->write()
```
ryanofsky, line 272: In commit "proxy, refactor: Replace EventLoop wakeup fd integers with KJ stream objects" (e96d5d742abf4d1c4af95134eef1e975b8f57a9a)

Note for followup. Unclear why I changed this comment to reference post() instead of EventLoopRef to. I think this is just a mistake and this should reference EventLoopRef::reset() specifically. Both of these methods write but only reset() sets the done() state and needs this logic.

I was also looking into why newTwoWayPipe is used here instead of newOneWayPipe, and it doesn't seem like there is a good reason. And there seems to be a newer `kj::Executor` mechanism https://github.com/capnproto/capnproto/blob/v2/kjdoc/tour.md#threads that would avoid the need for these pipes entirely.

I think I want to focus on getting windows support fully implemented in #231 and tested in CI but after that it seems like there is room for simplification here with `kj::Executor`

include/mp/util.h
```diff
@@ -259,17 +259,22 @@ using ProcessId = int;
 using SocketId = int;
 constexpr SocketId SocketError{-1};
 
+//! Information about parent process passed to child process as a command-line
+//! argument. On unix this is the child socket fd number formatted as a string.
+using SpawnConnectInfo = std::string;
```
ryanofsky, line 264: In commit "util, refactor: Add SpawnConnectInfo type alias and use it" (1389cf3132f48c903f56c9fcd8cc6e73d770cb19)

Note for followup: Looking at this code I think these type `SpawnConnectInfo` aliases actually hurt readability more than they help. I think originally I was thinking `SpawnConnectInfo` might have to be different types on different platforms to deal with the windows implementation #231 which sends a serialized socketed handle over a pipe to the child process instead of passing a file descriptor. But either way the application only sees an opaque string so `std::string` is a clearer way to represent this than `SpawnConnectInfo`

ryanofsky (commented):
Thanks for the reviews! I plan to merge this soon (as-is) but rereviewed this myself and left some notes for followups

PR #276, opened 2026-05-12
build: prepare for subtree split

### LINKED ISSUE #232: libmultiprocess-ext repo
I think I'd like to create a new `bitcoin-core/libmultiprocess-ext` repo so we can drop some extraneous things from this repository, particularly documentation, examples, CI scripts, and the shell.nix file, and keep only the core c++ library and tests and CMake files here with a small readme. Advantages:

- This should make Bitcoin Core libmultiprocess subtree smaller and diffs to the subtree easier to review.
- This should make it easier to distinguish PR's which actually affect Bitcoin Core code from PR's that don't and make sure changes affecting Bitcoin Core get the best possible review.
- This should declutter the libmultiprocess repository and increase development velocity. There are a bunch of PRs here like #229, parts of #228, #225, #223, #212, and parts of #196 that I feel like would be better to just merge and iterate on in a unified branch than keep open as PRs that will get stale and do not have great visibility. Since the changes would not affect Bitcoin Core, I don't think they need to be kept open waiting for acks or prolonged review.

If this makes sense, this could be done incrementally. For example CI and documentation could be moved first, since those could be simple moves, while moving examples could require a little more more thought about how we want to structure and reuse the cmake code.

Sjors:
We can use a [subtree split](https://git-memo.readthedocs.io/en/latest/subtree.html) to export only a subset of this repository to downstream projects. The only thing we need to do is publish and maintain a branch, e.g. `lib`, generated as follows: 

```sh
git subtree split --prefix=lib -b lib
```

I published such a branch here: https://github.com/Sjors/libmultiprocess/commits/lib/

Commits that don't touch the prefix directory are skipped entirely. Others are stripped of content outside the prefix. This of course breaks PGP signatures, but the `lib` branch commits are deterministically generated. On the Bitcoin Core side I plan to update the subtree linter to verify this.

Because we can only export one directory, this PR moves the following library sources to `lib/`:
- CMakeLists.txt
- cmake
- include
- pkgconfig
- src
- test

This leaves examples, documentation and CI outside the export. Tests could be moved out as well, but it seems useful to offer downstream projects an easy way to run these tests as part of their own CI.

A new top-level CMakeLists.txt wrapper preserves 'cmake -S .' for standalone builds; this is also the only way to build the example/ subdirectory. See comments in CMakeLists.txt and lib/CMakeLists.txt for details on how lib/ adapts between the two build modes. _This can use some CMake expertise eyes_. Most of the existing `CMakeList.txt` content is simply moved, so it's best reviewed with `--color-moved=dimmed-zebra`.

The second commit can be simplified if we drop examples from `cmake/libmultiprocess.cmake` in Bitcoin Core in a separate PR first.

Bitcoin Core side: https://github.com/Sjors/bitcoin/pull/113 (will upstream later)

Fixes #232


ryanofsky:
Concept ACK. It would be great if only needed files were imported into bitcoin core and things like ci scripts, documentation, and examples were excluded, since this could speed up review in both repositories.

The approach here also seems workable although I think I'd probably still prefer the idea of moving unneeded files to a support branch as described https://github.com/bitcoin-core/libmultiprocess/issues/232#issuecomment-3443012312 instead of using a partial subtree. That approach seems nice because it wouldn't require any changes to subtree import steps, or to the verifier, and wouldn't impact hashes, and wouldn't require any files except the support files to move. I also like the property of that approach where individual PRs only update the support branch or master branch, changes to master can't sneak in a support PR, and review effort can be focused on the right places (relatively more scrunity on master branch, and more freedom to experiment with CI improvements on the support branch).

Sjors:
Closing in favor of the approach in #287.

PR #277, opened 2026-05-14
Add std::unordered_set support and a helper BuildList to dedup list build handlers

ViniciusCestarii:
Add std::unordered_set support and a helper BuildList to dedup list build handlers that is being used for map, set and vector.

While looking bitcoin/bitcoin#29409, found a TODO noting that libmultiprocess lacked std::unordered_set support, requiring downstream that PR to implement the build/read functions locally.

I believe there could be more dedup adding a ReadList too. I could do that as a follow-up if desirable

### DIFF
--- a/test/mp/test/foo.h
+++ b/test/mp/test/foo.h
@@ -21,10 +21,10 @@ namespace test {
 struct FooStruct
 {
     std::string name;
-    std::set<int> setint;
-    std::vector<bool> vbool;
+    std::set<int> set_int;
+    std::vector<bool> v_bool;
     std::optional<int> optional_int;
-    std::unordered_set<int> unorderedsetint;
+    std::unordered_set<int> unordered_set_int;
 };
 
 enum class FooEnum : uint8_t { ONE = 1, TWO = 2, };
--- a/test/mp/test/foo.h
+++ b/test/mp/test/foo.h
@@ -12,6 +12,7 @@
 #include <optional>
 #include <string>
 #include <set>
+#include <unordered_set>
 #include <vector>
 
 namespace mp {
@@ -23,6 +24,7 @@ struct FooStruct
     std::set<int> setint;
     std::vector<bool> vbool;
     std::optional<int> optional_int;
+    std::unordered_set<int> unorderedsetint;
 };
 
 enum class FooEnum : uint8_t { ONE = 1, TWO = 2, };
--- a/include/mp/proxy-types.h
+++ b/include/mp/proxy-types.h
@@ -281,6 +281,17 @@ struct ListOutput<::capnp::List<T, kind>>
 };
 
 template <typename LocalType, typename Value, typename Output>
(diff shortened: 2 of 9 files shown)


ryanofsky (approved):
Code review ACK 00c9a8e91515bd367e28a2d5b1e235101d967aaf. This is a well-implemented change that simplifies https://github.com/bitcoin/bitcoin/pull/29409, and could help other places, since unordered_sets are used for various things in bitcoin core, and it makes sense to allow them to be easily passed over capnproto interfaces.

> I believe there could be more dedup adding a ReadList too. I could do that as a follow-up if desirable

Yes this would be a nice followup, especially since type-set.h and type-unordered-set.h files are almost exact duplicates of each other and other read functions are very similar. Might be possible to write a generic ReadList that works for all containers by taking an `init(size)` callback and an `emplace(args...)` callback.

ViniciusCestarii:
The new optional_int field on master uses a proper camelCase capnp name (optionalInt) with $Proxy.name mapping to a snake_case C++ member, which reads cleaner than the older lowercase style (setint, vbool). Should I follow that pattern here and rename unorderedsetint to an unorderedSetInt capnp field with a snake_case C++ member?

No need to touch the existing setint/vbool fields here. I believe that cleanup could be a separate PR.

ryanofsky:
> Should I follow that pattern here and rename unorderedsetint to an unorderedSetInt capnp field with a snake_case C++ member?

Anything seems fine to me, would choose whatever seems cleanest to you. Feel free to include a commit renaming existing fields in this PR too if you'd like

ViniciusCestarii:
Forced-push 9de4b885aa60b91a015f670a37bcbc64333c1d0c rebasing and fixing conflict with master and added the commit 9de4b885aa60b91a015f670a37bcbc64333c1d0c "test: use camelCase + $Proxy.name for FooStruct fields"

test/mp/test/foo.h
```diff
@@ -21,10 +21,10 @@ namespace test {
 struct FooStruct
 {
     std::string name;
-    std::set<int> setint;
-    std::vector<bool> vbool;
+    std::set<int> set_int;
+    std::vector<bool> v_bool;
```
ryanofsky, line 25: In commit "test: use camelCase + $Proxy.name for FooStruct fields" (9de4b885aa60b91a015f670a37bcbc64333c1d0c)

Not important but `vector_bool` would seem more consistent with other field names

ryanofsky (approved):
Code review ACK 9de4b885aa60b91a015f670a37bcbc64333c1d0c. Only changes since last review were renaming and adding commit user better field names in the test.

test/mp/test/foo.h
ViniciusCestarii: I agree. If in the future I touch these code lines again I will apply this suggestion. Thanks for the feedback!

PR #279, opened 2026-05-26
doc: Add comments to FIELD_* constants in proxy.h

ViniciusCestarii:
Add documentation comments to the five `FIELD_*` bit flag constants in `proxy.h`, which were previously undocumented

### DIFF
--- a/include/mp/proxy.h
+++ b/include/mp/proxy.h
@@ -304,20 +304,31 @@ struct ProxyServerMethodTraits : public ProxyMethodTraits<MethodParams>
 {
 };
 
-static constexpr int FIELD_IN = 1;
-static constexpr int FIELD_OUT = 2;
-static constexpr int FIELD_OPTIONAL = 4;
-static constexpr int FIELD_REQUESTED = 8;
-static constexpr int FIELD_BOXED = 16;
+static constexpr int FIELD_IN = 1;        //!< See Accessor::in.
+static constexpr int FIELD_OUT = 2;       //!< See Accessor::out.
+static constexpr int FIELD_OPTIONAL = 4;  //!< See Accessor::optional.
+static constexpr int FIELD_REQUESTED = 8; //!< See Accessor::requested.
+static constexpr int FIELD_BOXED = 16;    //!< See Accessor::boxed.
 
 //! Accessor type holding flags that determine how to access a message field.
 template <typename Field, int flags>
 struct Accessor : public Field
 {
+    //! Field is present from the Cap'n Proto Params struct (client -> server).
     static const bool in = flags & FIELD_IN;
+    //! Field is present from the Cap'n Proto Results struct (server -> client).
     static const bool out = flags & FIELD_OUT;
+    //! Field has a companion has{Name} boolean field in the Cap'n Proto struct.
+    //! This is used to represent optional primitive values (e.g. C++
+    //! std::optional<int>) because Cap'n Proto doesn't allow primitive fields to
+    //! be unset.
     static const bool optional = flags & FIELD_OPTIONAL;
+    //! Results field has a companion want{Name} boolean field in the Params
+    //! struct. Used for optional output parameters (e.g. C++ int*) and set to
+    //! true if the caller passed a non-null pointer and wants the result.
     static const bool requested = flags & FIELD_REQUESTED;
+    //! Field is a Cap'n Proto pointer type (struct, list, text, data,
+    //! interface) as opposed to a primitive type (bool, int, float, enum).
     static const bool boxed = flags & FIELD_BOXED;
 };
 


include/mp/proxy.h
```diff
@@ -304,11 +304,11 @@ struct ProxyServerMethodTraits : public ProxyMethodTraits<MethodParams>
 {
 };
 
-static constexpr int FIELD_IN = 1;
-static constexpr int FIELD_OUT = 2;
-static constexpr int FIELD_OPTIONAL = 4;
-static constexpr int FIELD_REQUESTED = 8;
-static constexpr int FIELD_BOXED = 16;
+static constexpr int FIELD_IN = 1;        //!< Field is read from the Cap'n Proto Params struct (client -> server).
+static constexpr int FIELD_OUT = 2;       //!< Field is read from the Cap'n Proto Results struct (server -> client).
```
ryanofsky: In commit "doc: Add comments to FIELD_* constants in proxy.h" (d70648079db6fac354c842e52a037f77bc29f102)

I think it would be better to say "present" instead of "read". "read" is accurate but these flags are used to decide whether to write fields as well, so "present" reflects the wider context

include/mp/proxy.h
```diff
@@ -304,11 +304,11 @@ struct ProxyServerMethodTraits : public ProxyMethodTraits<MethodParams>
 {
 };
 
-static constexpr int FIELD_IN = 1;
-static constexpr int FIELD_OUT = 2;
-static constexpr int FIELD_OPTIONAL = 4;
-static constexpr int FIELD_REQUESTED = 8;
-static constexpr int FIELD_BOXED = 16;
+static constexpr int FIELD_IN = 1;        //!< Field is read from the Cap'n Proto Params struct (client -> server).
```
ryanofsky: In commit "doc: Add comments to FIELD_* constants in proxy.h" (d70648079db6fac354c842e52a037f77bc29f102)

I think it would be better to move these comments to the Accessor struct below where they are used, and written as preceding comments (`//!`) trailing comments (`//!<`) so there is more space for the descriptions.

Could keep `//!<` comments here referring to the Accessor struct below though.

include/mp/proxy.h
```diff
@@ -304,11 +304,11 @@ struct ProxyServerMethodTraits : public ProxyMethodTraits<MethodParams>
...
+static constexpr int FIELD_OUT = 2;       //!< Field is read from the Cap'n Proto Results struct (server -> client).
+static constexpr int FIELD_OPTIONAL = 4;  //!< Field has a `has<Field>` sibling in the Cap'n Proto schema; value may be absent and presence must be checked via `has()`.
```
ryanofsky: In commit "doc: Add comments to FIELD_* constants in proxy.h" (d70648079db6fac354c842e52a037f77bc29f102)

Would replace the word "schema" with "struct" to be more specific. Also `has<Field>` looks like c++ template syntax which could be confusing. And I'm not sure the "presence must be checked" comment is meaningful because it seems focused on reading the field when the flag controls how the field needs to be written as well as read.

Would suggest something more like "Field has a companion `has{Name}` boolean field in the Cap'n Proto struct. This used to represent optional primitive values (e.g. C++ `std::optional<int>`) because Cap'n Proto doesn't allow primitive fields to be unset."

include/mp/proxy.h
```diff
@@ -304,11 +304,11 @@ struct ProxyServerMethodTraits : public ProxyMethodTraits<MethodParams>
...
+static constexpr int FIELD_REQUESTED = 8; //!< Field has a `want<Field>` sibling in the Cap'n Proto Params struct; client opts in to receiving it.
```
ryanofsky: In commit "doc: Add comments to FIELD_* constants in proxy.h" (d70648079db6fac354c842e52a037f77bc29f102)

This doesn't really say what the flag is used for. Maybe would write "Results field has a companion `want{Name}` boolean field in the Params struct. Used for optional output parameters (e.g. C++ `int*`) and set to true if the caller passed a non-null and wants the result."

include/mp/proxy.h
```diff
@@ -304,11 +304,11 @@ struct ProxyServerMethodTraits : public ProxyMethodTraits<MethodParams>
...
+static constexpr int FIELD_BOXED = 16;    //!< Field is a Cap'n Proto pointer type (struct, list, text, data, interface); value may be absent and presence must be checked via `has()`.
```
ryanofsky: In commit "doc: Add comments to FIELD_* constants in proxy.h" (d70648079db6fac354c842e52a037f77bc29f102)

I think absense/presence is not really relevant here. Would suggest something more like "Field is a Cap'n Proto pointer type (struct, list, text, data, interface) as opposed to a primitive type (bool, int, float, enum)."

ryanofsky (commented):
Code review d70648079db6fac354c842e52a037f77bc29f102. These comments are helpful but could be misleading in a few ways so I left some recommendations below.

ViniciusCestarii:
Forced-push e863c6cdf61437485012fc4219bc9dd8e20b218f to add suggested changes

include/mp/proxy.h
```diff
@@ -304,20 +304,31 @@ struct ProxyServerMethodTraits : public ProxyMethodTraits<MethodParams>
...
+static constexpr int FIELD_OUT = 2;       //!< See Accessor::out.
+static constexpr int FIELD_OPTIONAL = 4;  //!< See Accessor::optional.
+static constexpr int FIELD_REQUESTED = 8; //!< See Accessor::requested.
+static constexpr int FIELD_BOXED = 16;    //!< See Accessor::boxed.
 
 //! Accessor type holding flags that determine how to access a message field.
 template <typename Field, int flags>
 struct Accessor : public Field
 {
+    //! Field is present from the Cap'n Proto Params struct (client -> server).
```
ryanofsky, line 317: In commit "doc: Add comments to FIELD_* constants in proxy.h" (e863c6cdf61437485012fc4219bc9dd8e20b218f)

IMO "present in" would sound more natural than "present from" but this seems ok

ryanofsky (commented):
Code review ACK e863c6cdf61437485012fc4219bc9dd8e20b218f. Thanks for the comments!

include/mp/proxy.h
ViniciusCestarii: I agree. Thanks for the feedback

PR #280, opened 2026-05-28
build: exclude mptest target from compile commands

Sanjana2906:
build: exclude mptest target from compile commands

Excludes mptest from compile_commands.json because it includes generated
.capnp.h files that don't exist at configure time, causing IWYU failures
when this repo is used as a subtree in Bitcoin Core.

Follows the pattern established for example targets.

Refs bitcoin/bitcoin#35361 (this change will resolve that issue once pulled into Bitcoin Core)

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -37,4 +37,5 @@ if(BUILD_TESTING AND TARGET CapnProto::kj-test)
 
   add_dependencies(mptests mptest)
   add_test(NAME mptest COMMAND mptest)
+  set_target_properties(mptest PROPERTIES EXPORT_COMPILE_COMMANDS OFF)
 endif()


hebasto:
I think that the [issue](https://github.com/bitcoin/bitcoin/issues/35361) should be addressed in https://github.com/bitcoin/bitcoin/blob/master/cmake/libmultiprocess.cmake, similar to the [following](https://github.com/bitcoin/bitcoin/blob/10dfdd4b9fab578e43ba211325967cbc28cb6652/cmake/libmultiprocess.cmake#L32-L37):
```cmake
  # Exclude examples from compilation database, because the examples are not
  # built by default, and they contain generated c++ code. Without this
  # exclusion, tools like clang-tidy and IWYU that make use of compilation
  # database would complain that the generated c++ source files do not exist. An
  # alternate fix could build "mpexamples" by default like "mptests" above.
  set_target_properties(mpcalculator mpprinter mpexample PROPERTIES EXPORT_COMPILE_COMMANDS OFF)
```

Sanjana2906:
@hebasto  Thank you for confirming the approach.

I've opened PR #35418 with exactly this change — adding `set_target_properties(mptest PROPERTIES EXPORT_COMPILE_COMMANDS OFF)` to `cmake/libmultiprocess.cmake`, following the same pattern as the examples.

Link: https://github.com/bitcoin/bitcoin/pull/35418

ryanofsky:
Should this be closed given existence of https://github.com/bitcoin/bitcoin/pull/35418 and #284?

Sanjana2906:
@ryanofsky Yes, thanks. The change is now being tracked in bitcoin/bitcoin#35418, so this PR is no longer needed. I'll close this one.

Issue #281, opened 2026-05-28
RFC: Server should execute requests on a thread pool when client does not specify an execution thread


ViniciusCestarii:
Referencing: https://btctranscripts.com/bitcoin-core-dev-tech/2026-05/external-interfaces

>Problem: IPC interface can be awkward to call due to threading
>- Clients have to choose which thread asynchronous calls execute on, and create threads for calls to execute on
>- Could be improved by server executing requests on a thread pool when client does not specify an execution thread

Currently, clients that call IPC methods must first call ThreadMap.makeThread to create a server-side worker thread and pass the resulting Thread handle in Context.thread on every call so it executes on a dedicated thread. Libmultiprocess abstracts this away for C++ clients, but clients that don't use libmultiprocess (which only supports C++) must handle this setup manually, making the IPC interface more awkward to call.

Proposed improvement: When a client passes a null/missing Context.thread, or when a method has no mp.Context parameter at all, the server should fall back to executing the request on a server-managed thread pool instead of blocking the event loop thread or failing with "invalid thread handle". This would simplify non-libmultiprocess clients that doesn't need a dedicated thread model.

Caveats:
- Callbacks require the client to still provide Context.callbackThread
- Response ordering is not guaranteed when requests from the same client land on different pool threads. The dedicated thread model avoids this by always queuing requests from the same client on one thread

PR #283, opened 2026-05-29
Add `makePool` method on `ThreadMap`

rustaceanrob:
This patch introduces a pool of threads to the `Connection` class, and allows this pool to be populated with the thread map via `makePool`. When a client thread is not set in a request context, it is delegated to the pool. This is unable to handle the guarentees with server-invoked callbacks that the current API offers, but these callbacks are not yet present in the interface.

The pool is implemented as ~round-robin as it is simplest~ shortest queue, but perhaps the pool could be a queue of requests with work-stealing for threads that are available.

This was raised to me by Rust users, as they did not particularly care where work is executed on the server-side, but they have to set the thread regardless.

Tests in Rust can be seen here: https://github.com/2140-dev/bitcoin-capnp-types/pull/24

ref: https://github.com/2140-dev/bitcoin-capnp-types/blob/master/tests/util/bitcoin_core.rs#L149
ref: #281

### DIFF
--- a/include/mp/proxy.capnp
+++ b/include/mp/proxy.capnp
@@ -45,6 +45,10 @@ interface ThreadMap $count(0) {
     # execute on. Clients create and name threads and pass the thread handle as
     # a call parameter.
     makeThread @0 (name :Text) -> (result :Thread);
+    # Pre-allocate a pool of server threads for implicit dispatch. When a
+    # request arrives with no context.thread set, the server dispatches it
+    # through this pool via a shared work queue.
+    makePool @1 (count :UInt32) -> ();
 }
 
 interface Thread {
--- a/include/mp/type-context.h
+++ b/include/mp/type-context.h
@@ -200,63 +200,95 @@ auto PassField(Priority<1>, TypeList<>, ServerContext& server_context, const Fn&
     // asynchronously with getLocalServer().
     const auto& params = server_context.call_context.getParams();
     Context::Reader context_arg = Accessor::get(params);
-    auto thread_client = context_arg.getThread();
-    auto result = server.m_context.connection->m_threads.getLocalServer(thread_client)
-        .then([&loop, invoke = kj::mv(invoke), req](const kj::Maybe<Thread::Server&>& perhaps) mutable {
-            // Assuming the thread object is found, pass it a pointer to the
-            // `invoke` lambda above which will invoke the function on that
-            // thread.
-            KJ_IF_MAYBE (thread_server, perhaps) {
-                auto& thread = static_cast<ProxyServer<Thread>&>(*thread_server);
-                MP_LOG(loop, Log::Debug)
-                    << "IPC server post request  #" << req << " {" << thread.m_thread_context.thread_name << "}";
-                return thread.template post<typename ServerContext::CallContext>(std::move(invoke));
-            } else {
-                MP_LOG(loop, Log::Error)
-                    << "IPC server error request #" << req << ", missing thread to execute request";
-                throw std::runtime_error("invalid thread handle");
-            }
-        }, [&loop, req](::kj::Exception&& e) -> kj::Promise<typename ServerContext::CallContext> {
-            // If you see the error "(remote):0: failed: remote exception:
-            // Called null capability" here, it probably means your Init class
-            // is missing a declaration like:
-            //
-            //   construct @0 (threadMap: Proxy.ThreadMap) -> (threadMap :Proxy.ThreadMap);
-            //
-            // which passes a ThreadMap reference from the client to the server,
-            // allowing the server to create threads to run IPC calls on the
-            // client, and also returns a ThreadMap reference from the server to
-            // the client, allowing the client to create threads on the server.
-            // (Typically the latter ThreadMap is used more often because there
-            // are more client-to-server calls.)
-            //
-            // If the other side of the connection did not previously get a
-            // ThreadMap reference from this side of the connection, when the
-            // other side calls `m_thread_map.makeThreadRequest()` in
-            // `BuildField` above, `m_thread_map` will be null, but that call
-            // will not fail immediately due to Cap'n Proto's request pipelining
-            // and delayed execution. Instead that call will return an invalid
-            // Thread reference, and when that reference is passed to this side
-            // of the connection as `thread_client` above, the
-            // `getLocalServer(thread_client)` call there will be the first
-            // thing to overtly fail, leading to an error here.
-            //
-            // Potentially there are also other things that could cause errors
-            // here, but this is the most likely cause.
-            //
-            // The log statement here is not strictly necessary since the same
-            // exception will also be logged in serverInvoke, but this logging
-            // may provide extra context that could be helpful for debugging.
-            MP_LOG(loop, Log::Info)
-                << "IPC server error request #" << req << " CapabilityServerSet<Thread>::getLocalServer call failed, did you forget to provide a ThreadMap to the client prior to this IPC call?";
-            return kj::mv(e);
-        });
-    // Use connection m_canceler object to cancel the result promise if the
-    // connection is destroyed. (By default Cap'n Proto does not cancel requests
-    // on disconnect, since it's possible clients might want to make requests
-    // and immediately disconnect without waiting for results, but not want the
-    // requests to be canceled.)
-    return server.m_context.connection->m_canceler.wrap(kj::mv(result));
+    if (!context_arg.hasThread()) {
+        // No client thread specified — dispatch through the server thread
+        // pool, picking the slot with the smallest in-flight depth.
+        auto* connection = server.m_context.connection;
+        auto& pool = connection->m_thread_pool;
+        if (pool.empty()) {
+            MP_LOG(loop, Log::Error)
+                << "IPC server error request #" << req << ", no thread specified and no pool configured";
+            throw std::runtime_error("no thread specified and no pool configured");
+        }
+        auto* slot = &pool[0];
+        for (size_t i = 1; i < pool.size(); ++i) {
+            if (pool[i].depth < slot->depth) slot = &pool[i];
+        }
+        ++slot->depth;
+        auto result = connection->m_threads.getLocalServer(slot->client)
+            .then([&loop, invoke = kj::mv(invoke), req](const kj::Maybe<Thread::Server&>& perhaps) mutable -> kj::Promise<typename ServerContext::CallContext> {
+                KJ_IF_MAYBE (thread_server, perhaps) {
+                    auto& thread = static_cast<ProxyServer<Thread>&>(*thread_server);
+                    MP_LOG(loop, Log::Debug)
+                        << "IPC server post request  #" << req << " {" << thread.m_thread_context.thread_name << "}";
+                    return thread.template post<typename ServerContext::CallContext>(std::move(invoke));
+                } else {
+                    MP_LOG(loop, Log::Error)
+                        << "IPC server error request #" << req << ", pool thread not found";
+                    throw std::runtime_error("pool thread not found");
+                }
+            })
+            .attach(kj::defer([slot] { --slot->depth; }));
+        return connection->m_canceler.wrap(kj::mv(result));
+    } else {
+        auto thread_client = context_arg.getThread();
+        auto result = server.m_context.connection->m_threads.getLocalServer(thread_client)
+            .then([&loop, invoke = kj::mv(invoke), req](const kj::Maybe<Thread::Server&>& perhaps) mutable {
+                // Assuming the thread object is found, pass it a pointer to the
+                // `invoke` lambda above which will invoke the function on that
+                // thread.
+                KJ_IF_MAYBE (thread_server, perhaps) {
+                    auto& thread = static_cast<ProxyServer<Thread>&>(*thread_server);
+                    MP_LOG(loop, Log::Debug)
+                        << "IPC server post request  #" << req << " {" << thread.m_thread_context.thread_name << "}";
+                    return thread.template post<typename ServerContext::CallContext>(std::move(invoke));
+                } else {
+                    MP_LOG(loop, Log::Error)
+                        << "IPC server error request #" << req << ", missing thread to execute request";
+                    throw std::runtime_error("invalid thread handle");
+                }
+            }, [&loop, req](::kj::Exception&& e) -> kj::Promise<typename ServerContext::CallContext> {
+                // If you see the error "(remote):0: failed: remote exception:
+                // Called null capability" here, it probably means your Init class
+                // is missing a declaration like:
+                //
+                //   construct @0 (threadMap: Proxy.ThreadMap) -> (threadMap :Proxy.ThreadMap);
+                //
+                // which passes a ThreadMap reference from the client to the server,
+                // allowing the server to create threads to run IPC calls on the
+                // client, and also returns a ThreadMap reference from the server to
+                // the client, allowing the client to create threads on the server.
+                // (Typically the latter ThreadMap is used more often because there
+                // are more client-to-server calls.)
+                //
+                // If the other side of the connection did not previously get a
+                // ThreadMap reference from this side of the connection, when the
+                // other side calls `m_thread_map.makeThreadRequest()` in
+                // `BuildField` above, `m_thread_map` will be null, but that call
+                // will not fail immediately due to Cap'n Proto's request pipelining
+                // and delayed execution. Instead that call will return an invalid
+                // Thread reference, and when that reference is passed to this side
+                // of the connection as `thread_client` above, the
+                // `getLocalServer(thread_client)` call there will be the first
+                // thing to overtly fail, leading to an error here.
+                //
+                // Potentially there are also other things that could cause errors
+                // here, but this is the most likely cause.
+                //
+                // The log statement here is not strictly necessary since the same
+                // exception will also be logged in serverInvoke, but this logging
+                // may provide extra context that could be helpful for debugging.
+                MP_LOG(loop, Log::Info)
+                    << "IPC server error request #" << req << " CapabilityServerSet<Thread>::getLocalServer call failed, did you forget to provide a ThreadMap to the client prior to this IPC call?";
+                return kj::mv(e);
+            });
+        // Use connection m_canceler object to cancel the result promise if the
+        // connection is destroyed. (By default Cap'n Proto does not cancel requests
+        // on disconnect, since it's possible clients might want to make requests
+        // and immediately disconnect without waiting for results, but not want the
+        // requests to be canceled.)
+        return server.m_context.connection->m_canceler.wrap(kj::mv(result));
+    }
 }
 } // namespace mp
 
--- a/src/mp/proxy.cpp
+++ b/src/mp/proxy.cpp
@@ -26,6 +26,7 @@
 #include <kj/function.h>
 #include <kj/memory.h>
 #include <kj/string.h>
+#include <cstdint>
 #include <map>
 #include <memory>
 #include <optional>
@@ -36,6 +37,7 @@
 #include <tuple>
 #include <unistd.h>
 #include <utility>
+#include <vector>
 
 namespace mp {
 
@@ -414,6 +416,29 @@ kj::Promise<void> ProxyServer<Thread>::getName(GetNameContext context)
 }
 
 ProxyServer<ThreadMap>::ProxyServer(Connection& connection) : m_connection(connection) {}
+
+kj::Promise<void> ProxyServer<ThreadMap>::makePool(MakePoolContext context)
+{
+    if (!m_connection.m_thread_pool.empty()) {
+        throw std::runtime_error("makePool called on connection with existing pool");
+    }
+    EventLoop& loop{*m_connection.m_loop};
+    const uint32_t count = context.getParams().getCount();
+    for (uint32_t i = 0; i < count; ++i) {
+        const std::string thread_name = "pool/" + std::to_string(i);
+        std::promise<ThreadContext*> thread_context;
+        std::thread thread([&loop, &thread_context, thread_name]() {
+            g_thread_context.thread_name = ThreadName(loop.m_exe_name) + " (" + thread_name + ")";
+            g_thread_context.waiter = std::make_unique<Waiter>();
+            Lock lock(g_thread_context.waiter->m_mutex);
+            thread_context.set_value(&g_thread_context);
+            g_thread_context.waiter->wait(lock, [] { return !g_thread_context.waiter; });
+        });
+        auto thread_server = kj::heap<ProxyServer<Thread>>(m_connection, *thread_context.get_future().get(), std::move(thread));
+        m_connection.m_thread_pool.push_back({m_connection.m_threads.add(kj::mv(thread_server))});
+    }
+    return kj::READY_NOW;
+}
 
 kj::Promise<void> ProxyServer<ThreadMap>::makeThread(MakeThreadContext context)
 {
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -477,6 +477,13 @@ public:
     //! ThreadMap.makeThread) used to service requests to clients.
     ::capnp::CapabilityServerSet<Thread> m_threads;
 
+    //! A thread created by makePool with associated pending work queue. Vector is filled once by makePool() and never resized.
+    struct PoolSlot {
+        Thread::Client client;
+        size_t depth{0};
+    };
+    std::vector<PoolSlot> m_thread_pool;
+
     //! Canceler for canceling promises that we want to discard when the
     //! connection is destroyed. This is used to interrupt method calls that are
     //! still executing at time of disconnection.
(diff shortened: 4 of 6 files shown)


include/mp/proxy.capnp
```diff
@@ -45,6 +45,10 @@ interface ThreadMap $count(0) {
     # execute on. Clients create and name threads and pass the thread handle as
     # a call parameter.
     makeThread @0 (name :Text) -> (result :Thread);
+    # Pre-allocate a pool of server threads for implicit dispatch. When a
+    # request arrives with no context.thread set, the server round-robins
+    # across the pool instead of returning an error.
+    makePool @1 (count :UInt32) -> ();
```
ViniciusCestarii, line 51:
```suggestion
    makePool @1 (name: Text, count :UInt32) -> ();
```

It would be usefull to have a name parameter like makeThread. Since each client may have its own thread pool, naming them would make log lines like "IPC server post request #{name} {pool/0}" attributable to a specific client

src/mp/proxy.cpp
```diff
@@ -415,6 +417,26 @@ kj::Promise<void> ProxyServer<Thread>::getName(GetNameContext context)
...
+            g_thread_context.waiter = std::make_unique<Waiter>();
+            Lock lock(g_thread_context.waiter->m_mutex);
+            thread_context.set_value(&g_thread_context);
+            g_thread_context.waiter->wait(lock, [] { return !g_thread_context.waiter; });
+        });
+        auto thread_server = kj::heap<ProxyServer<Thread>>(m_connection, *thread_context.get_future().get(), std::move(thread));
+        m_connection.m_thread_pool.push_back(m_connection.m_threads.add(kj::mv(thread_server)));
+    }
+    return kj::READY_NOW;
+}
```
ViniciusCestarii, line 441: Repeated calls silently grow the pool beyond the intended size, and since naming is based on the loop index starting from 0 each time, you'd end up with duplicate names like two threads both called pool/0 making logs ambiguous.

I'm not sure what would be the correct approach but I thought about:
- Return an error if m_connection.m_thread_pool is not empty
- Replace the old pool by cleaning the m_connection.m_thread_pool and create a fresh one with the new count. This makes sense if the client needs to resize its dedicated pool.

include/mp/proxy.capnp
rustaceanrob: Updated. How does that look?

ViniciusCestarii: Nice, I just have this nit to make it more consistent with makeThread:

```diff
         const std::string thread_name = pool_name + "/pool/" + std::to_string(i);
         std::promise<ThreadContext*> thread_context;
         std::thread thread([&loop, &thread_context, thread_name]() {
-            g_thread_context.thread_name = ThreadName(loop.m_exe_name) + " (" + thread_name + ")";
+            g_thread_context.thread_name = ThreadName(loop.m_exe_name) + " (from " + thread_name + ")";
             g_thread_context.waiter = std::make_unique<Waiter>();
             Lock lock(g_thread_context.waiter->m_mutex);
             thread_context.set_value(&g_thread_context);
```

src/mp/proxy.cpp
rustaceanrob: Added a guard statement:
```cpp
    if (!m_connection.m_thread_pool.empty()) {
        throw std::runtime_error("makePool called on connection with existing pool");
    }
```

I think throwing is appropriate here as I would like to get user feedback first. If they anticipate pool resizing during runtime will be valuable we can add resizing APIs

ryanofsky:
Concept ACK. This is a good idea that should make development of rust & python clients easier, while making the behavior opt-in and not affecting c++ clients or existing clients using current behavior.

With this feature, we may also want to add a way to mark certain methods that can take long time before returning to not use the thread pool, so they do not block other work. These methods could create their own threads to run on, or require threads to be specified.

> This is unable to handle the guarentees with server-invoked callbacks that the current API offers

Actually I think this is not an issue. If a server method is called with a [callbackThread](https://github.com/bitcoin-core/libmultiprocess/blob/3c69d125a175084e4056ec9db8badb625e5f31d8/include/mp/proxy.capnp#L62-L64) value it can still pass the same value back to the client when making the callback, no matter where the server thread is running.

> perhaps the pool could be a queue of requests with work-stealing for threads that are available

Yes I need to look more closely at current implementation, but would seem better to have a single queue of waiting tasks instead of giving each thread its own queue. Otherwise execution order could be very unpredictable and if a single request is slow, it could create a backlog of requests assigned to the same thread. It seems like it should be possible to have a shared queue without too much complexity, so probably worth looking into.

xyzconstant:
Concept ACK. This feature significantly reduces integration burden for clients written in other programming languages.

I just glanced at it, planning to review it deeply soon.

One thing to note: This PR will definitely need release notes.

Sjors:
Would be nice to have a before and after usage example (with `2140-dev/bitcoin-capnp-types`).

rustaceanrob:
~Currently stale but I made an example [here](https://github.com/2140-dev/bitcoin-capnp-types/pull/24) within the tests. No more `set_thread` required.~ See [comment](https://github.com/bitcoin-core/libmultiprocess/pull/283#issuecomment-4635478657)

rustaceanrob:
Should I take a hack at implementing proper work delegation throughout the pool or would reviewers prefer I keep it as is for now?

xyzconstant:
re: https://github.com/bitcoin-core/libmultiprocess/pull/283#issuecomment-4633191181

> Should I take a hack at implementing proper work delegation throughout the pool or would reviewers prefer I keep it as is for now?

On my side, I'd love to see a more elaborate pool.

I was thinking about something like https://github.com/bitcoin/bitcoin/blob/master/src/util/threadpool.h as reference

xyzconstant:
Just tested 0ac739b0fd0869ec153ba86d875fe95065860182 on [peer-observer's ipc-extractor](https://github.com/xyzconstant/peer-observer/commit/c67385a6d987d3999f55571a523e5586912a2941) with a `ChainNotifications` callback as per bitcoin/bitcoin#29409

It's working as expected and, notably, reduces the overhead of creating a `thread ` instance for each IPC call.

rustaceanrob:
Re-implemented as a "shortest queue" pool, where the thread with the lowest count of pending jobs is selected, tiebreaker by lowest index. I'd say this is more than sufficient for now (I would imagine users would not have more outstanding jobs than allocated threads in the current mining use case).

src/mp/proxy.cpp
```diff
@@ -415,6 +417,31 @@ kj::Promise<void> ProxyServer<Thread>::getName(GetNameContext context)
...
+        std::promise<ThreadContext*> thread_context;
+        std::thread thread([&loop, &thread_context, thread_name]() {
+            g_thread_context.thread_name = ThreadName(loop.m_exe_name) + " (from " + thread_name + ")";
+            g_thread_context.waiter = std::make_unique<Waiter>();
+            Lock lock(g_thread_context.waiter->m_mutex);
+            thread_context.set_value(&g_thread_context);
+            g_thread_context.waiter->wait(lock, [] { return !g_thread_context.waiter; });
+        });
+        auto thread_server = kj::heap<ProxyServer<Thread>>(m_connection, *thread_context.get_future().get(), std::move(thread));
+        m_connection.m_thread_pool.push_back({m_connection.m_threads.add(kj::mv(thread_server)), 0});
```
ViniciusCestarii: On a828771f995abaf0a918cd3006cf1a54c928490e

nit: no need for explict 0, because PoolSlot.depth already defaults to 0

include/mp/type-context.h
```diff
@@ -201,20 +201,44 @@ auto PassField(Priority<1>, TypeList<>, ServerContext& server_context, const Fn&
...
+                            MP_LOG(loop, Log::Debug)
+                                << "IPC server post request  #" << req << " {" << pool_thread.m_thread_context.thread_name << "}";
+                            return pool_thread.template post<typename ServerContext::CallContext>(std::move(invoke))
+                                .attach(kj::mv(guard));
+                        } else {
+                            MP_LOG(loop, Log::Error)
+                                << "IPC server error request #" << req << ", pool thread not found";
+                            throw std::runtime_error("pool thread not found");
+                        }
+                    });
```
ViniciusCestarii: On a828771f995abaf0a918cd3006cf1a54c928490e

I believe it's better to move the depth-decrement guard up so it's created right after `++slot->depth` and attached to the whole dispatch chain instead of just the inner `post()`. That keeps the increment and its decrement paired with no async hop between them, so the slot's `depth` is always released however the chain ends.

And you can use kj::defer instead of a custom struct.

  ```cpp
  ++slot->depth;
  auto guard = kj::defer([slot] { --slot->depth; });
  return connection->m_threads.getLocalServer(slot->client)
      .then([&loop, invoke = kj::mv(invoke), req](const kj::Maybe<Thread::Server&>& pool_perhaps) mutable {
          // ... dispatch to pool_thread ...
      })
      .attach(kj::mv(guard));
  ```

ViniciusCestarii (commented):
Shortest-queue selection is looking good

rustaceanrob:
Nice feedback, updated baab0fe33e6e3a2b83acb25819547e24d89a187f

include/mp/type-context.h
```diff
@@ -201,20 +201,43 @@ auto PassField(Priority<1>, TypeList<>, ServerContext& server_context, const Fn&
...
+                    .then([&loop, invoke = kj::mv(invoke), req](const kj::Maybe<Thread::Server&>& pool_perhaps) mutable {
+                        KJ_IF_MAYBE (pt, pool_perhaps) {
+                            auto& pool_thread = static_cast<ProxyServer<Thread>&>(*pt);
+                            MP_LOG(loop, Log::Debug)
+                                << "IPC server post request  #" << req << " {" << pool_thread.m_thread_context.thread_name << "}";
+                            return pool_thread.template post<typename ServerContext::CallContext>(std::move(invoke));
+                        } else {
+                            MP_LOG(loop, Log::Error)
+                                << "IPC server error request #" << req << ", pool thread not found";
+                            throw std::runtime_error("pool thread not found");
```
xyzconstant: I'm wondering how this block could be exercised. A test for this would be great.

include/mp/type-context.h
```diff
@@ -201,20 +201,43 @@ auto PassField(Priority<1>, TypeList<>, ServerContext& server_context, const Fn&
...
+                    MP_LOG(loop, Log::Error)
+                        << "IPC server error request #" << req << ", no thread specified and no pool configured";
+                    throw std::runtime_error("no thread specified and no pool configured");
+                }
+                auto* slot = &pool[0];
+                for (size_t i = 1; i < pool.size(); ++i) {
+                    if (pool[i].depth < slot->depth) slot = &pool[i];
+                }
+                ++slot->depth;
+                auto guard = kj::defer([slot] { --slot->depth; });
```
xyzconstant: nit: `KJ_DEFER` macro would be more consistent IMO.

xyzconstant (commented):
Thanks for the changes @rustaceanrob

I exercised this test again: https://github.com/bitcoin-core/libmultiprocess/pull/283#issuecomment-4635478657 and downstream IPC tests, everything is working as expected.

I am still reviewing but I have left a couple of comments.

Note that `doc/design.md` needs to be updated too to reflect this new feature

include/mp/type-context.h
rustaceanrob: `KJ_DEFER` doesn't support move assignment

rustaceanrob:
> Note that doc/design.md needs to be updated too to reflect this new feature

Perhaps as a follow up? Or at least I'd like to get a few approach ACKs so we know this design won't change

include/mp/type-context.h
rustaceanrob: This can only happen if a worker is removed from `m_threads` and not `m_thread_pool`, which only be possible if we added a shrink/resize method. I could add something like this, but for now I'm not sure if it's necessary. FWIW the previous `else` path also seems untested.

```diff
diff --git a/test/mp/test/test.cpp b/test/mp/test/test.cpp
index 9be8742..cd5d5f4 100644
--- a/test/mp/test/test.cpp
+++ b/test/mp/test/test.cpp
@@ -598,5 +598,38 @@ KJ_TEST("Call async IPC method without thread or pool errors correctly")
     KJ_EXPECT(error_thrown);
 }

+KJ_TEST("Pool dispatch with torn-down slot surfaces error")
+{
+    TestSetup setup;
+    ProxyClient<messages::FooInterface>* foo = setup.client.get();
+    setup.server->m_impl->m_fn = [] {};
+
+    std::promise<void> done;
+    bool error_thrown{false};
+    foo->m_context.loop->sync([&] {
+        foo->m_context.connection->m_thread_pool.push_back(
+            {Thread::Client{KJ_EXCEPTION(FAILED, "pool slot torn down")}});
+
+        auto request{foo->m_client.callFnAsyncRequest()};
+        request.initContext();
+        foo->m_context.loop->m_task_set->add(
+            request.send().then(
+                [&](auto&&) { done.set_value(); },
+                [&](kj::Exception&& e) {
+                    error_thrown = true;
+                    KJ_EXPECT(std::string_view{e.getDescription().cStr()}.find(
+                        "pool thread not found") != std::string_view::npos);
+                    done.set_value();
+                }));
+    });
+    done.get_future().get();
+    KJ_EXPECT(error_thrown);
+}
+
```

xyzconstant: Makes sense thanks

include/mp/type-context.h
xyzconstant: I actually meant something like this:

```diff
diff --git a/include/mp/type-context.h b/include/mp/type-context.h
index d241d95..e31ceee 100644
--- a/include/mp/type-context.h
+++ b/include/mp/type-context.h
@@ -223,9 +223,9 @@ auto PassField(Priority<1>, TypeList<>, ServerContext& server_context, const Fn&
                     if (pool[i].depth < slot->depth) slot = &pool[i];
                 }
                 ++slot->depth;
-                auto guard = kj::defer([slot] { --slot->depth; });
                 return connection->m_threads.getLocalServer(slot->client)
-                    .then([&loop, invoke = kj::mv(invoke), req](const kj::Maybe<Thread::Server&>& pool_perhaps) mutable {
+                    .then([&loop, invoke = kj::mv(invoke), req, slot](const kj::Maybe<Thread::Server&>& pool_perhaps) mutable {
+                        KJ_DEFER(--slot->depth);
                         KJ_IF_MAYBE (pt, pool_perhaps) {
                             auto& pool_thread = static_cast<ProxyServer<Thread>&>(*pt);
                             MP_LOG(loop, Log::Debug)
@@ -236,8 +236,7 @@ auto PassField(Priority<1>, TypeList<>, ServerContext& server_context, const Fn&
                                 << "IPC server error request #" << req << ", pool thread not found";
                             throw std::runtime_error("pool thread not found");
                         }
-                    })
-                    .attach(kj::mv(guard));
+                    });
             }
         }, [&loop, req](::kj::Exception&& e) -> kj::Promise<typename ServerContext::CallContext> {
             // If you see the error "(remote):0: failed: remote exception:
```

However, this is a nit

xyzconstant: The diff above compiles and passes the tests, but so does this other:

```diff
diff --git a/include/mp/type-context.h b/include/mp/type-context.h
index d241d95..0a99610 100644
--- a/include/mp/type-context.h
+++ b/include/mp/type-context.h
@@ -223,7 +223,6 @@ auto PassField(Priority<1>, TypeList<>, ServerContext& server_context, const Fn&
                     if (pool[i].depth < slot->depth) slot = &pool[i];
                 }
                 ++slot->depth;
-                auto guard = kj::defer([slot] { --slot->depth; });
                 return connection->m_threads.getLocalServer(slot->client)
                     .then([&loop, invoke = kj::mv(invoke), req](const kj::Maybe<Thread::Server&>& pool_perhaps) mutable {
                         KJ_IF_MAYBE (pt, pool_perhaps) {
@@ -236,8 +235,7 @@ auto PassField(Priority<1>, TypeList<>, ServerContext& server_context, const Fn&
                                 << "IPC server error request #" << req << ", pool thread not found";
                             throw std::runtime_error("pool thread not found");
                         }
-                    })
-                    .attach(kj::mv(guard));
+                    });
             }
         }, [&loop, req](::kj::Exception&& e) -> kj::Promise<typename ServerContext::CallContext> {
             // If you see the error "(remote):0: failed: remote exception:
```

ryanofsky:
> Or at least I'd like to get a few approach ACKs so we know this design won't change

Definite approach ACK for baab0fe33e6e3a2b83acb25819547e24d89a187f & sorry I haven't reviewed this more closely yet.

This is a simple and elegant way of adding a thread pool while changing existing code as little as possible.

My one concern about this is I don't think scheduling algorithm is very robust. But it's probably good enough for practical purposes, and it can be improved later. An example of a problem would be that if that if there are 10 threads, and a client makes 100 method calls, and 99 of the method calls return instantly, but one call waits for a new block to be connected and takes minutes to return. In that case, some of the 99 method calls which could complete instantaneously will be stuck queued behind the one slow call, even though there are threads sitting idle which could execute them. To fix this, it would be best to just add all incoming requests to a single queue, and have threads read from the common queue, instead of giving each thread its own individual queue. I asked claude to implement this to get an idea of what it would look like bd1f80c31b95c2041bd5feb13ae710a78698dcec ([branch](https://github.com/ryanofsky/libmultiprocess/commits/pr/tpool)) but it's a more complicated approach and completely untested and I'd be happy to save something like this for a followup.

Other more minor things:

- Would like to drop the `name` parameter from makePool, even though it could be useful for debugging. Since it only makes sense to have one thread pool per connection, the name string more of a way to identify the connection than the thread pool, and it could probably be useful in other contexts even in cases where no thread pool is needed. So I think it'd be best just to drop the name from the makePool method. And if having a name is helpful for debugging, add a separate `setConnectionName` method to name the connection. (We could also consider renaming `ThreadMap` interface to `Connection` to generalize it.)

- I think implementation should avoid calling `m_threads.getLocalServer(context_arg.getThread())` if `context_arg.hasThread()` is false. Should be no need to do a lookup on a null capability. This should just be a minor readability/efficiency improvement.

Otherwise this looks very good as far as I can tell and I do want to get this reviewed and merged.

rustaceanrob:
dd05eeab0c83d7fe5cb91804f826e7af87033415 removes `name` argument. I'd like to leave `setConnectionName` and similar refactors to a followup. Will get to the second comment soon.

rustaceanrob:
@ryanofsky, r.e. second comment, is this what you mean?

```diff
diff --git a/include/mp/type-context.h b/include/mp/type-context.h
index d241d95..9751de7 100644
--- a/include/mp/type-context.h
+++ b/include/mp/type-context.h
@@ -202,7 +202,10 @@ auto PassField(Priority<1>, TypeList<>, ServerContext& server_context, const Fn&
     Context::Reader context_arg = Accessor::get(params);
     auto thread_client = context_arg.getThread();
     auto* connection = server.m_context.connection;
-    auto result = connection->m_threads.getLocalServer(thread_client)
+    auto thread_promise = context_arg.hasThread()
+        ? connection->m_threads.getLocalServer(thread_client)
+        : kj::Promise<kj::Maybe<Thread::Server&>>(nullptr);
+    auto result = thread_promise
         .then([&loop, invoke = kj::mv(invoke), req, connection](const kj::Maybe<Thread::Server&>& perhaps) mutable {
             // If the client specified a thread, dispatch to it directly.
             KJ_IF_MAYBE (thread_server, perhaps) {
```

ryanofsky:
> @ryanofsky, r.e. second comment, is this what you mean?

That's close but not exactly what I mean. I'd like to preserve the current "invalid thread handle" error in the case where `context_arg.hasThread()` is true but getLocalServer returns null. I'm actually not sure whether that case is possible to hit, but if it is possible I'd want it to be an error, not to silently fall back to using the thread pool when a thread was passed but couldn't be mapped to a local object for some reason.

So idea would be more like `if (!context_arg.hasThread()) { ... your new code here ...} else { ... existing code ....}` but hopefully not needing to move existing code too much.

rustaceanrob:
Unfortunately a larger diff but all existing code moved into the `else` block here d77e897eaf5e8bd8de9ea11b64233d9a6d354f25

Eunovo:
Concept ACK

This removes the requirement for non-libmultiprocess clients to manage thread handles explicitly.

A server-wide threadpool is worth considering. The current per-connection pool allocates O(connections × pool_size) OS threads. External clients can open an unpredictable number of connections, so this scales without bound. A shared pool would cap total thread count at a fixed size across all connections.

include/mp/proxy.capnp
```diff
@@ -45,6 +45,10 @@ interface ThreadMap $count(0) {
     # execute on. Clients create and name threads and pass the thread handle as
     # a call parameter.
     makeThread @0 (name :Text) -> (result :Thread);
+    # Pre-allocate a pool of server threads for implicit dispatch. When a
+    # request arrives with no context.thread set, the server dispatches it
+    # through this pool via a shared work queue.
+    makePool @1 (count :UInt32) -> ();
```
xyzconstant, line 51: nit: Relocate the comment below `makePool` definition line to maintain consistency with the file.

xyzconstant (approved):
Code review ACK d77e897eaf5e8bd8de9ea11b64233d9a6d354f25

I think the changes are clean and ready to be merged.

Ran local and IPC tests on downstream, compiled a `bitcoin-node` instance to verify it wasn't affected, and finally tested the changes in a Rust client (see: https://github.com/xyzconstant/peer-observer/commit/c414765a0af113832c95ef2b26ba0e25fb9416fc).

This significantly improves how clients written in other languages consume `libmultiprocess` processes by eliminating the need to create a new thread for each IPC call, because the serving processes now automatically assign threads to requests with no threads specified.

I explored another alternative to the current approach but felt this was the most effective solution.

As a next step, based on [ryanofsky's suggestion (comment)](https://github.com/bitcoin-core/libmultiprocess/pull/283#issuecomment-4732869389), it might be helpful to rename the `ThreadMap` interface to something like `Connection` or `Runtime` for better clarity.

xyzconstant:
Following on from my review:

> it might be helpful to rename the `ThreadMap` interface to something like `Connection` or `Runtime` for better clarity.

I've made [some progress](https://github.com/rustaceanrob/libmultiprocess/compare/26-5-29-pooled-threads...xyzconstant:libmultiprocess:rename-threadmap-to-connection) on top of these changes. Happy to open a follow-up PR if/once this one gets merged.

ViniciusCestarii:
tACK d77e897eaf5e8bd8de9ea11b64233d9a6d354f25

Tested with https://github.com/ViniciusCestarii/bitcoin-ipc-demo/tree/make-pool-tester and it successfully worked passing simultaneous calls to the threads with smaller queues.

src/mp/proxy.cpp
```diff
@@ -415,6 +417,29 @@ kj::Promise<void> ProxyServer<Thread>::getName(GetNameContext context)
...
+    if (!m_connection.m_thread_pool.empty()) {
+        throw std::runtime_error("makePool called on connection with existing pool");
+    }
+    EventLoop& loop{*m_connection.m_loop};
+    const uint32_t count = context.getParams().getCount();
+    for (uint32_t i = 0; i < count; ++i) {
+        const std::string thread_name = "pool/" + std::to_string(i);
+        std::promise<ThreadContext*> thread_context;
+        std::thread thread([&loop, &thread_context, thread_name]() {
+            g_thread_context.thread_name = ThreadName(loop.m_exe_name) + " (from " + thread_name + ")";
```
ryanofsky: In commit "Add `makePool` method on `ThreadMap`" (d77e897eaf5e8bd8de9ea11b64233d9a6d354f25)

Not important would probably be a little better to drop `"from " + ` here.

Format for client thread names is

```bash
client_exe_name-123/client_thread_name-456
```

And for server thread names

```bash
server_exe_name-234/server_thread_name-567 (from client_exe_name-123/client_thread_name-456)
```

Where numbers are process/thread ids. This code would make the server thread name look like:

```bash
server_exe_name-234/server_thread_name-567 (from pool/0)
```

Which makes pool/0 look ilke a client process.

ryanofsky (approved):
Code review ACK d77e897eaf5e8bd8de9ea11b64233d9a6d354f25. Looks good! Sorry for the delay getting to this and I plan to merge this soon after looking over the other review comments.

I still do have the same concern about scheduling algorithm not being very robust (https://github.com/bitcoin-core/libmultiprocess/pull/283#issuecomment-4732869389) since requests that should be fast could get stuck queued behind much slower requests and be delayed for a long time (potentially indefinitely) even if workers are free which could execute them. A good fix for this would be to add incoming requests to a shared queue and have workers read from the queue. But it sounds like current design does work well in practice, so I'm not very concerned about this and this could be improved in a followup

rustaceanrob:
Dropped `from` string 0e49d91186717d757d43c9be3ba939b0c1305491

FYI will not be able to rebase again until July 6th if more changes are required

ryanofsky (approved):
Code review ACK 0e49d91186717d757d43c9be3ba939b0c1305491 just tweaking pool thread names since last review

xyzconstant:
re-ACK 0e49d91186717d757d43c9be3ba939b0c1305491

ViniciusCestarii:
re-ACK  0e49d91186717d757d43c9be3ba939b0c1305491

PR #273, opened 2026-04-22
proxy-client: tolerate exceptions from remote destroy during cleanup

xyzconstant:
Fixes #219.

Catches exceptions from `Sub::destroy(*this)` in `~ProxyClientBase`'s cleanup lambda so a failed remote destroy call doesn't crash the process. Regression test included.

PR #286, opened 2026-06-03
proxy-client: fix TSan data race in clientDestroy

ryanofsky:
#273 added a new test which exposed a longstanding race condition in `clientDestroy` logging code, causing the sanitize CI job to now fail with a TSAN error. Fix the race by simplifying the logging code.

Details about the fix are in the commit message. TSAN error looked like

```
[ TEST ] test.cpp:437: Destroying ProxyClient<> with destroy method after peer disconnect
...
  Write of size 8 at 0x721400005020 by thread T24 (mutexes: write M0):
    #0 std::__1::__function::__func<mp::ProxyClientBase<mp::test::messages::FooCallback, mp::test::FooCallback>::ProxyClientBase(mp::test::messages::FooCallback::Client, mp::Connection*, bool)::{lambda()#1}, void ()>::operator()() /home/runner/work/libmultiprocess/libmultiprocess/include/mp/proxy-io.h:522 (mptest+0x242b20)
    #1 void mp::Unlock<mp::Lock, std::__1::function<void ()>&>(mp::Lock&, std::__1::function<void ()>&) /nix/store/6z6xj37ljxwrwd5w3rs0zyfkrxd48lzl-libcxx-21.1.2-dev/include/c++/v1/__functional/function.h:274 (mptest+0x35582d)
    #2 mp::Connection::~Connection() /home/runner/work/libmultiprocess/libmultiprocess/src/mp/proxy.cpp:150 (mptest+0x34e596)
    ...
    #11 mp::test::TestSetup::TestSetup(bool)::{lambda()#1}::operator()() const /home/runner/work/libmultiprocess/libmultiprocess/test/mp/test/test.cpp:112 (mptest+0x1504fe)

  Previous read of size 8 at 0x721400005020 by thread T26:
    #0 void mp::clientDestroy<mp::ProxyClient<mp::test::messages::FooCallback> >(mp::ProxyClient<mp::test::messages::FooCallback>&) /home/runner/work/libmultiprocess/libmultiprocess/include/mp/proxy-types.h:647 (mptest+0x342d2a)
     ...
    #21 std::__1::__function::__func<mp::ProxyServerBase<mp::test::messages::FooInterface, mp::test::FooImplementation>::~ProxyServerBase()::{lambda()#1}, void ()>::operator()() /nix/store/6z6xj37ljxwrwd5w3rs0zyfkrxd48lzl-libcxx-21.1.2-dev/include/c++/v1/__functional/function.h:174 (mptest+0x347c76)
     ...
    #23 void* std::__1::__thread_proxy[abi:ne210101]<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, mp::EventLoop::startAsyncThread()::$_0> >(void*) /home/runner/work/libmultiprocess/libmultiprocess/src/mp/proxy.cpp:309 (mptest+0x353ef2)

  Thread T24 (tid=10032, running) created by main thread at:
    ...
    #4 mp::test::TestCase437::run() /home/runner/work/libmultiprocess/libmultiprocess/test/mp/test/test.cpp:449 (mptest+0x1459a7)

```

https://github.com/bitcoin-core/libmultiprocess/actions/runs/26851297744/job/79183796152

Note: #273 made it possible for this TSAN error to trigger, but it didn't trigger most of the time, so it was not detected in CI. There is also an [interaction](https://github.com/bitcoin-core/libmultiprocess/pull/286#issuecomment-4631871296) with #218 and #273 + #218 together make it much more likely to trigger

### DIFF
--- a/include/mp/proxy-types.h
+++ b/include/mp/proxy-types.h
@@ -644,11 +644,7 @@ struct CapRequestTraits<::capnp::Request<_Params, _Results>>
 template <typename Client>
 void clientDestroy(Client& client)
 {
-    if (client.m_context.connection) {
-        MP_LOG(*client.m_context.loop, Log::Debug) << "IPC client destroy " << CxxTypeName(client);
-    } else {
-        KJ_LOG(INFO, "IPC interrupted client destroy", CxxTypeName(client));
-    }
+    MP_LOG(*client.m_context.loop, Log::Debug) << "IPC client destroy " << CxxTypeName(client);
 }
 
 template <typename Server>
--- a/src/mp/gen.cpp
+++ b/src/mp/gen.cpp
@@ -6,6 +6,7 @@
 #include <mp/util.h>
 
 #include <algorithm>
+#include <capnp/schema.capnp.h> // IWYU pragma: keep
 #include <capnp/schema.h>
 #include <capnp/schema-parser.h>
 #include <cerrno>
@@ -329,12 +330,17 @@ static void Generate(kj::StringPtr src_prefix,
     cpp_client << "#include <" << include_path << ".h>\n";
     cpp_client << "#include <" << include_path << ".proxy.h>\n";
     cpp_client << "#include <" << include_path << ".proxy-types.h>\n";
+    cpp_client << "#include <capnp/capability.h>\n";
+    cpp_client << "#include <capnp/common.h>\n";
     cpp_client << "#include <capnp/generated-header-support.h>\n";
     cpp_client << "#include <cstring>\n";
-    cpp_client << "#include <vector>\n";
+    cpp_client << "#include <functional>\n";
     cpp_client << "#include <kj/common.h>\n";
+    cpp_client << "#include <map>\n";
     cpp_client << "#include <mp/proxy.h>\n";
     cpp_client << "#include <mp/util.h>\n";
+    cpp_client << "#include <string>\n";
+    cpp_client << "#include <vector>\n";
     cpp_client << "#include <" << PROXY_TYPES << ">\n";
     cpp_client << "// IWYU pragma: end_keep\n\n";
     cpp_client << "namespace mp {\n";
@@ -346,8 +352,6 @@ static void Generate(kj::StringPtr src_prefix,
     cpp_types << "#include <" << include_path << ".h> // IWYU pragma: keep\n";
     cpp_types << "#include <" << include_path << ".proxy.h>\n";
     cpp_types << "#include <" << include_path << ".proxy-types.h> // IWYU pragma: keep\n";
-    cpp_types << "#include <kj/common.h>\n";
-    cpp_types << "#include <mp/util.h>\n";
     cpp_types << "#include <" << PROXY_TYPES << ">\n\n";
     cpp_types << "namespace mp {\n";
 


ryanofsky:
Added 1 commits 25bb3e67f3918bf03a5317da3138204886337347 -> 0963634f5b01b05a5fdd4bd810e915a8052892e1 ([`pr/drace.1`](https://github.com/ryanofsky/libmultiprocess/commits/pr/drace.1) -> [`pr/drace.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/drace.2), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/drace.1...pr/drace.2)) to fix IWYU errors https://github.com/bitcoin-core/libmultiprocess/actions/runs/26919582954/job/79416996848?pr=286

Updated 0963634f5b01b05a5fdd4bd810e915a8052892e1 -> 039e5ac14bd8bb8a0532431474fbfd70bf0a2557 ([`pr/drace.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/drace.2) -> [`pr/drace.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/drace.3), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/drace.2..pr/drace.3)) to fix IWYU errors https://github.com/bitcoin-core/libmultiprocess/actions/runs/26924152551/job/79430685686?pr=286

w0xlt:
Concept ACK,

enirox001 (commented):
ACK 039e5ac

Built and ran `mptest` with  `-fsanitize=thread`, all 11 tests pass cleanly.

The fix makes sense to me. Instead of adding synchronization for what constitutes a log line, we drop the branch and unconditionally log via `MP_LOG`.

ryanofsky:
Thanks for the reviews. I also went back and tried to debug why CI did not catch the TSAN error. It turns out to be because an interaction between #273 and #218. The new test added in #273 does not trigger the TSAN error deterministically. When it is run standalone without #218 there's usually no error, although I did see a TSAN error once running the test maybe 10 times. But when #273 and #218 are combined, the test triggers TSAN failures most of the time. This seems to happen because #218 changes `clientDestroy` to call `CxxTypeName` which can change thread timing.

ryanofsky:
Started to merge this but I saw some more IWYU errors locally that seem better to fix. I think the errors don't happen in CI because nixpkgs are older and use an older version of capnproto. I also ran IWYU locally before my last push, but didn't see these errors because I was using a combined branch with #287 with example files removed. Errors look like:

```
/home/ryan/mp/build-default/example/calculator.capnp.proxy-client.c++ should add these lines:
#include <capnp/common.h>                    // for MessageSize
#include <string>                            // for basic_string
```


Updated 039e5ac14bd8bb8a0532431474fbfd70bf0a2557 -> 90982f75c6bfd764e9e8c1c1fb63c3858e0f6c48 ([`pr/drace.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/drace.3) -> [`pr/drace.4`](https://github.com/ryanofsky/libmultiprocess/commits/pr/drace.4), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/drace.3..pr/drace.4)) touching up IWYU commit to fix more local IWYU errors

maflcko:
lgtm ACK 90982f75c6bfd764e9e8c1c1fb63c3858e0f6c48

A bit tedious, that different capnp versions require different set of includes, but I guess there is nothing that can be done about that in this repo.

ryanofsky:
> A bit tedious, that different capnp versions require different set of includes, but I guess there is nothing that can be done about that in this repo.

Yes, this repo is taking a different strategy for IWYU than the Bitcoin core repo so it is easier to run IWYU locally.

With Bitcoin core AFAIK you need to run IWYU in a container and can't run it as part of your normal development process because headers must be changed to match what a single CI machine enforces, and pragmas are discouraged.

With this repo there is no special IWYU configuration. IWYU can be run in any configuration and "pragma: keeps" can be added freely to turn off complaints from other configurations. From my perspective this is good because there's no need for a container to run IWYU, and you can waste less time debugging IWYU errors and figuring out what "correct" includes are. You get the main benefits of IWYU (faster compiles, cleaner headers) at the cost of needing a few mysterious pragmas. I think this is about as much as you can hope for with IWYU.

maflcko:
Right, it is obviously more correct to run iwyu for the full range of  all supported third-party dependency versions (especially if they have different include requirements).

Maybe it could make sense to expand the CI to cover both oldeps-iwyu and nightly-iwyu (or so) here?

I think for Bitcoin Core this doesn't matter, because the dependencies don't change the include requirements across versions. (E.g. boost multi index is just a single include header)

If that were to change, we should do the same. But no strong opinion, and this seems to be trailing off a bit :sweat_smile:

ViniciusCestarii (commented):
ACK, 90982f75c6bfd764e9e8c1c1fb63c3858e0f6c48

This change removes differentiation on the logging when the connection was torn down "IPC interrupted client destroy" but I believe this is fine and simpler than locking a mutex.

xyzconstant:
tACK 90982f75c6bfd764e9e8c1c1fb63c3858e0f6c48

w0xlt:
post-merge ACK

PR #289, opened 2026-06-04
cmake: make target_capnp_sources use CURRENT dirs

ryanofsky:
Tweak `target_capnp_sources` to use `CMAKE_CURRENT_SOURCE_DIR` and `CMAKE_CURRENT_BINARY_DIR` instead of `CMAKE_SOURCE_DIR` and `CMAKE_BINARY_DIR` to compute include directory paths.

This allows `target_capnp_sources` to work properly inside libmultiprocess when libmultiprocess is used as a subproject and `CMAKE_SOURCE_DIR` point somewhere outside of it, as happens in support branch used by #287.

This is also a good change to make more generally because `CMAKE_CURRENT_SOURCE_DIR` is a more predictable path than `CMAKE_SOURCE_DIR` and is passed in by all current `target_capnp_sources` callers.

### DIFF
--- a/cmake/TargetCapnpSources.cmake
+++ b/cmake/TargetCapnpSources.cmake
@@ -27,10 +27,8 @@ Arguments:
 
       #include <schema.capnp.h>
 
-    The specified include_prefix should be ${CMAKE_SOURCE_DIR} or a
-    subdirectory of it to include files relative to the project root. It can
-    be ${CMAKE_CURRENT_SOURCE_DIR} to include files relative to the current
-    source directory.
+    Pass ${CMAKE_CURRENT_SOURCE_DIR} or a subdirectory of it to include files
+    relative to the current source directory (the typical usage).
 
 Additional Unnamed Arguments:
 
@@ -46,9 +44,9 @@ Optional Keyword Arguments:
 Example:
   # Assuming `my_library` is a target and `lib/` contains `.capnp` schema
   # files with imports from `include/`.
-  target_capnp_sources(my_library "${CMAKE_SOURCE_DIR}"
+  target_capnp_sources(my_library "${CMAKE_CURRENT_SOURCE_DIR}"
                        lib/schema1.capnp lib/schema2.capnp
-                       IMPORT_PATHS ${CMAKE_SOURCE_DIR}/include)
+                       IMPORT_PATHS ${CMAKE_CURRENT_SOURCE_DIR}/include)
 
 #]=]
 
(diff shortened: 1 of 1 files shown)


hebasto (approved):
ACK 620f297f311718f3d186b716676779e046e3255c, tested with the Bitcoin Core master branch where `relative_path` is now  empty for all call sites.

From [Professional CMake: A Practical Guide, 22nd Edition, Section 8.5](https://discourse.cmake.org/t/professional-cmake-a-practical-guide-22nd-edition/15470):
> Where possible, avoid using the `CMAKE_SOURCE_DIR` and `CMAKE_BINARY_DIR` variables, as these typically break the ability of the project to be incorporated into a larger project hierarchy.

While touching this file, could we add some more improvements?
1. Document the `ONLY_CAPNP` option.
2. Use absolute paths when specifying the `OUTPUT` files. 
From [Professional CMake: A Practical Guide, 22nd Edition, Section 20.3](https://discourse.cmake.org/t/professional-cmake-a-practical-guide-22nd-edition/15470):
    > If the output files are specified with no path or with a relative path, they are normally treated as relative to the current binary directory. In specific circumstances, they can be treated as relative to the current source directory instead ... To avoid that ambiguity, projects should always use absolute paths when specifying the `OUTPUT` files.

hebasto (approved):
re-ACK 615a94fe3a213380eb50cde82e4b4d718bf63dd5.

ryanofsky:
re: https://github.com/bitcoin-core/libmultiprocess/pull/289#pullrequestreview-4427856835

Thanks for the review!

> While touching this file, could we add some more improvements?

Thanks I had claude implement both suggestions and added the first here. I didn't add the second one (c1f35afeac28c4bda58378da2caf9ad32407f7ba) because the advice to prefer absolute paths in this case sounds faulty to me, without knowing further details. (It sounds like the reason could be to hide cmake bugs, but I don't have the book.) There are many cases where absolute paths are better to use, but also cases where they are worse. In general they can make commands harder to read and mistakes and inconsistencies harder to notice. Switching to them can discard semantic information and make code more complicated fragile and buggy it needs to strip and re-add them. The bug this PR fixes happens in code failing to stripping out and add path prefixes, so is actually an example of a bug that would not happen if relative paths were used.


Added 1 commits 620f297f311718f3d186b716676779e046e3255c -> 615a94fe3a213380eb50cde82e4b4d718bf63dd5 ([`pr/starget.1`](https://github.com/ryanofsky/libmultiprocess/commits/pr/starget.1) -> [`pr/starget.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/starget.2), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/starget.1...pr/starget.2)) documenting `ONLY_CAPNP` option

PR #290, opened 2026-06-05
cmake: Skip `PATH` environment variable when discovering CapnProto

### LINKED ISSUE #14: OpenSuse 16 Leap capnp issue
Steps to reproduce:

`zypper in -y zeromq-devel libevent-devel boost-devel sqlite3-devel  qt6-base-devel qt6-tools-devel qt6-linguist-devel qrencode-devel      find bison gcc-c++ libtool make autoconf automake python3 clang llvm  libcapnp-devel capnproto patch xz      curl wget htop git vim ccache  && git clone https://github.com/bitcoin/bitcoin.git  --depth=1 ./b-c && cd b-c &&    rm -rf ./bld-cmake && cmake -B ./bld-cmake -DCMAKE_C_COMPILER='clang' -DCMAKE_CXX_COMPILER='clang++' -DBUILD_GUI=ON -DBUILD_FUZZ_BINARY=ON -DBUILD_BENCH=ON -DWITH_ZMQ=ON -DBUILD_UTIL_CHAINSTATE=ON -DBUILD_KERNEL_LIB=ON                            && cmake --build ./bld-cmake --parallel $(nproc)`

Failure:

```
-- Configuring done (0.5s)
-- Generating done (0.1s)
-- Build files have been written to: /root/b-c/build-cdash
Command exited with the value: 0
SetCTestConfiguration:BuildDirectory:/root/b-c/build-cdash
SetCTestConfiguration:SourceDirectory:/root/b-c
SetMakeCommand:/usr/bin/cmake --build . --config "Release"
SetCTestConfiguration:MakeCommand:/usr/bin/cmake --build . --config "Release"
Build project
MakeCommand:/usr/bin/cmake --build . --config "Release"
Run command: "/usr/bin/cmake" "--build" "." "--config" "Release"
[1/589] Compiling Cap'n Proto schema include/mp/proxy.capnp
FAILED: src/ipc/libmultiprocess/include/mp/proxy.capnp.c++ src/ipc/libmultiprocess/include/mp/proxy.capnp.h /root/b-c/build-cdash/src/ipc/libmultiprocess/include/mp/proxy.capnp.c++ /root/b-c/build-cdash/src/ipc/libmultiprocess/include/mp/proxy.capnp.h 
cd /root/b-c/build-cdash/src/ipc/libmultiprocess && //bin/capnp compile -o //bin/capnpc-c++:/root/b-c/build-cdash/src/ipc/libmultiprocess --src-prefix /root/b-c/src/ipc/libmultiprocess -I /root/b-c/src/ipc/libmultiprocess -I /lib64/cmake/CapnProto/../../../include /root/b-c/src/ipc/libmultiprocess/include/mp/proxy.capnp
//bin/capnp compile: -I /lib64/cmake/CapnProto/../../../include: no such directory
Try '//bin/capnp compile --help' for more information.

hebasto:
On systems with UsrMerge, the `PATH` environment variable can [cause](https://github.com/maflcko/bitcoin-core-nightly/issues/14#issuecomment-4632269615) CMake to produce include paths that contain root-level symlinks. The `capnp` compiler resolves paths lexically and fails to process them properly, resulting in a "no such directory" error.

Fixes https://github.com/maflcko/bitcoin-core-nightly/issues/14.

CI runs in Bitcoin Core repo: https://github.com/hebasto/bitcoin/actions/runs/27019344100/job/79742757484.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ endif()
 include("cmake/compat_find.cmake")
 
 find_package(Threads REQUIRED)
-find_package(CapnProto 0.7 QUIET NO_MODULE)
+find_package(CapnProto 0.7 QUIET NO_MODULE NO_SYSTEM_ENVIRONMENT_PATH)
  if(NOT CapnProto_FOUND)
    message(FATAL_ERROR
      "Cap'n Proto is required but was not found.\n"


ryanofsky:
So the problem on these systems is that `find_package` detects capnproto in `/lib64/cmake/CapnProto` and constructs the include path `/lib64/cmake/CapnProto/../../../include`.

Both of these paths are valid and resolve to `/usr/lib64/cmake/CapnProto` and `/usr/include` respectively by the operating system because `/lib64` is a symlink to `/usr/lib64`

The problem is that capnproto is not using the operating system to resolve the include path. Instead it tries to simplify the include path lexically, resulting in `/include` which doesn't exist, instead of `/usr/include` which is what it actually resolves to. This is a bug in capnproto, because it's invalid to just drop path components followed by `/..` when symlinks are involved. You have to resolve symlinks first, then drop path components. It sounds like a bug that should be reported and fixed in capnproto.

Trying to work around in this libmultiprocess is a hack because libmultiprocess is not involved in constructing or interpreting these paths. The fix is also fragile. It's interesting that CMake locates the package in `/lib64/cmake/CapnProto` if NO_SYSTEM_ENVIRONMENT_PATH is not specified, and `/usr/lib64/cmake/CapnProto` otherwise but it is fragile to rely on this behavior since both paths are valid and it seems like `/ib64` could wind up being searched any number of ways even if PATH search is disabled.

It also seems like users can fix this by setting PATH variables more sanely (https://github.com/maflcko/bitcoin-core-nightly/issues/14#issuecomment-4632269615) to point to real directories instead of /bin and /sbin symlinks created for backwards compatibility. Or to at least order the symlinks after the real directories.

If we want to address this issue at all in libmultiprocess, i think the think to do would be detect when CAPNP_INCLUDE_DIRECTORY can't be resolved lexically and recommend users fix their PATH or use -DCMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH=OFF as a workaround.

But this seems like something that should be resolved by users of these systems or by capnproto, not something libmultiprocess should be involved with.

hebasto:
@ryanofsky 

Fair enough. Closing.

ryanofsky:
Thanks and to be sure I don't object to this change if it's documented properly either as a workaround or as a reasoned design decision to override cmake's default treatment of the PATH variable. I just want to avoid special cases in the build scripts that could make them hard to maintain & safely modify.

PR #292, opened 2026-06-09
type-number: fix clang-tidy modernize-use-nullptr

ViniciusCestarii:
Prevent warning --modernize-use-nullptr when using clang-tidy on downstream projects (Noticed when pushing https://github.com/2140-dev/bitcoin/pull/42, this change was also necessary for https://github.com/bitcoin/bitcoin/pull/29409 so CI doens't error)

### DIFF
--- a/include/mp/type-number.h
+++ b/include/mp/type-number.h
@@ -82,7 +82,7 @@ decltype(auto) CustomReadField(TypeList<LocalType>,
     InvokeContext& invoke_context,
     Input&& input,
     ReadDest&& read_dest,
-    typename std::enable_if<std::is_floating_point<LocalType>::value>::type* enable = 0)
+    typename std::enable_if<std::is_floating_point<LocalType>::value>::type* enable = nullptr)
 {
     auto value = input.get();
     static_assert(std::is_same<LocalType, decltype(value)>::value, "floating point type mismatch");


maflcko:
Looks like CI fails in mptest, which may be unrelated.

Also, I wonder why the CI passes fine without this change, when it does claim to run clang-tidy: https://github.com/bitcoin-core/libmultiprocess/blob/61de6975362a7070276da47cc2aa2c2b8909ed49/ci/configs/llvm.bash#L8

ViniciusCestarii:
> Also, I wonder why the CI passes fine without this change, when it does claim to run clang-tidy:

The CI does run clang-tidy, but it only warns about code that actually gets compiled. The float overload of CustomReadField (the one with = 0) is a template, so it only turns into real code when something uses a float type. libmultiprocess's own foo.capnp has no Float64 field, so that code is never generated and clang-tidy never looks at it. Bitcoin Core's .capnp does use Float64, so mpgen generates the code, and then clang-tidy can see the type and warn with modernize-use-nullptr.

Adding a Float64 field on foo.capnp and foo.h would have triggered this error here. I believe there other types that are not being tested too.

ViniciusCestarii:
> Looks like CI fails in mptest, which may be unrelated.

The same thing happened on https://github.com/bitcoin-core/libmultiprocess/pull/278 which only changes a markdown file. I will give a look on what's happening

ryanofsky (approved):
Code review ACK 91a951f59ac9e43e408456bf052b9fb1179563a6. Thanks for the PR. It could also be a good to add a test that would trigger this error. I guess would just need to add a test passing a floating point value.

maflcko:
Added test in https://github.com/bitcoin-core/libmultiprocess/pull/294

PR #295, opened 2026-06-10
Mark Waiter m_cv as guarded by m_mutex

maflcko:
This change fixes a lost wakeup bug in the "Make simultaneous IPC calls on single remote thread" test which is probably responsible for the hang reported https://github.com/bitcoin/bitcoin/issues/35491. The bug has existed since the test was introduced in 1238170f68e8fa7ae41c79465df5cdae34d568e9.

The change adds an `MP_GUARDED_BY(m_mutex)` annotation to `Waiter::m_cv`, and locks the mutex before notifying the condition variable in the test to satisfy the guarded-by requirement. 

Locking the mutex prevents a hang in the test that could happen because if there is no lock between the `running -= 1` state update and notifying the condition variable, there is no guarantee that the test is actively waiting on the condition variable when the notify call is made, so the notify might do nothing, and the test could wait forever.

(_edited by ryanofsky_)

### DIFF
--- a/test/mp/test/test.cpp
+++ b/test/mp/test/test.cpp
@@ -510,6 +510,7 @@ KJ_TEST("Make simultaneous IPC calls on single remote thread")
                 [&running, &tc, i](auto&& results) {
                     assert(results.getResult() == static_cast<int32_t>(100 * (i+1)));
                     running -= 1;
+                    Lock lock(tc.waiter->m_mutex);
                     tc.waiter->m_cv.notify_all();
                 }));
         }
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -384,7 +384,7 @@ struct Waiter
     }
 
     template <class Predicate>
-    void wait(Lock& lock, Predicate pred)
+    void wait(Lock& lock, Predicate pred) MP_REQUIRES(m_mutex)
     {
         m_cv.wait(lock.m_lock, [&]() MP_REQUIRES(m_mutex) {
             // Important for this to be "while (m_fn)", not "if (m_fn)" to avoid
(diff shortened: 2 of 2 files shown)


test/mp/test/test.cpp
```diff
@@ -510,6 +510,7 @@ KJ_TEST("Make simultaneous IPC calls on single remote thread")
                 [&running, &tc, i](auto&& results) {
                     assert(results.getResult() == static_cast<int32_t>(100 * (i+1)));
                     running -= 1;
+                    Lock lock(tc.waiter->m_mutex);
```
ryanofsky, line 513: In commit "Mark Waiter m_cv as guarded by m_mutex" (fa35501c4f0ac57ffe18479791ddf209c817aa39)

Note if we wanted to make this more optimal and avoid the "hurry up and wait" scenario described in https://en.cppreference.com/cpp/thread/condition_variable/notify_one, we would unlock before notifying as recommended there, which the MP_GUARDED_BY annotation prevents:

```diff
--- a/test/mp/test/test.cpp
+++ b/test/mp/test/test.cpp
@@ -511,6 +511,7 @@ KJ_TEST("Make simultaneous IPC calls on single remote thread")
                     assert(results.getResult() == static_cast<int32_t>(100 * (i+1)));
                     running -= 1;
                     Lock lock(tc.waiter->m_mutex);
+                    lock.unlock();
                     tc.waiter->m_cv.notify_all();
                 }));
         }
```

Still worth having the annotation here, but this shows why the annotation could be undesirable in other cases.

ryanofsky (approved):
Code review ACK fa35501c4f0ac57ffe18479791ddf209c817aa39. Thanks for the fix!

re: https://github.com/bitcoin-core/libmultiprocess/pull/295#issue-4631671320

> This allows to catch possible deadlock bugs at compile time.
> 
> For [bitcoin/bitcoin#35491](https://github.com/bitcoin/bitcoin/issues/35491)

I think PR description should be clearer that this does fix a specific deadlock not just "possible deadlock bugs", so I might edit it before merging (hope you don't mind)

PR #297, opened 2026-06-11
test: add map serialization round-trip coverage

ViniciusCestarii:
Add test to ensure map serialization is working and will keep working. Useful for #285.

Also renames `v_bool` to `vector_bool`.

Best reviewed commit by commit.

### DIFF
--- a/test/mp/test/foo.capnp
+++ b/test/mp/test/foo.capnp
@@ -59,6 +59,7 @@ struct FooStruct $Proxy.wrap("mp::test::FooStruct") {
     optionalInt @3 :Int32 $Proxy.name("optional_int");
     hasOptionalInt @4 :Bool;
     unorderedSetInt @5 :List(Int32) $Proxy.name("unordered_set_int");
+    mapStringInt @6 :List(StringIntPair) $Proxy.name("map_string_int");
 }
 
 struct FooCustom $Proxy.wrap("mp::test::FooCustom") {
@@ -80,4 +81,9 @@ struct FooMutable {
 struct Pair(T1, T2) {
     first @0 :T1;
     second @1 :T2;
+}
+
+struct StringIntPair {
+    first @0 :Text;
+    second @1 :Int32;
 }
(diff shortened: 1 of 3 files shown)


ViniciusCestarii:
Forced-push 645d90b to include map on test.cpp for IWYU

w0xlt:
Concept ACK

xyzconstant:
Looks clean to me, tACK 645d90b5461fcca540df781859ee8e8c06cb514a

test/mp/test/foo.capnp
```diff
@@ -78,11 +77,6 @@ struct FooMutable {
     message @0 :Text;
 }
 
-struct Pair(T1, T2) {
```
ryanofsky, line 80: In commit "test: drop mapSize, now covered by map round-trip test" (490585f044355d09dcac054b7f77a4efdffdec48)

I'm not really a fan the commit dropping the mapsize test. It seem to me the generic Pair() type is good to keep because it could be used in other tests testing non-primitive maps, and also it makes helps make sure the code generator and library code works with generics in capnproto, not just non-generic types.

I also don't love renumbering the passDouble method. It seems a little cleaner to me to either fully follow the compatibility rules for updating schema files, or fully not follow them and just update all the field numbers to match field order. It doesn't seem like there is a rationale for numbering fields out of order and also not following compatibility rules.

ryanofsky (approved):
Code review ACK 645d90b5461fcca540df781859ee8e8c06cb514a. Looks good but would slightly prefer to keep the generic Pair(T1, T2) type as described below

test/mp/test/foo.capnp
ViniciusCestarii: Good points, dropped the commit "test: drop mapSize, now covered by map round-trip test" 490585f. OBS: I still kept the StringIntPair because [only pointer types (structs, lists, blobs, and interfaces) can be used as generic parameters](https://capnproto.org/language.html#:~:text=Only%20pointer%20types%20(structs%2C%20lists%2C%20blobs%2C%20and%20interfaces)%20can%20be%20used%20as%20generic%20parameters).

ViniciusCestarii:
Thanks for reviewing! forced push 463d073cb8df9b7bcab33b3483af7e1759ac67a9 dropping commit "test: drop mapSize, now covered by map round-trip test" https://github.com/bitcoin-core/libmultiprocess/commit/490585f044355d09dcac054b7f77a4efdffdec48 because mapSize/Pair() test is useful for generic capnp coverage.

ryanofsky (approved):
Code review ACK 463d073cb8df9b7bcab33b3483af7e1759ac67a9

Thanks for the update!

PR #269, opened 2026-04-08
proxy: add local connection limit to ListenConnections

enirox001:
This adds an optional local connection limit to`ListenConnections()`.

Previously, `ListenConnections()` would accept incoming connections indefinitely. This branch adds an optional `max_connections` parameter so a listener can stop accepting new connections once a per-listener cap is reached, and resume accepting when an existing connection disconnects.

The limit is local to the listener instead of global to the `EventLoop`. This keeps the state and behavior scoped to the listening socket, and is closer to the direction discussed downstream for per-`-ipcbind` limits.

This also adds a test covering the behavior with `max_connections=1`, verifying that:
  - the first client is accepted normally
  - a second client is not accepted while the first remains connected
  - the second client is accepted after the first disconnects

_Note This PR includes a major version bump to `v12` due to the API addition. If #274 lands earlier and bumps the version to `v12` first, we will need to bump the version here again._

Issue #183, opened 2025-06-16
bitcoin-cli uncaught kj::ExceptionImpl with #32297


ryanofsky:
_Originally posted by @pinheadmz in [#182](https://github.com/bitcoin-core/libmultiprocess/issues/182#issuecomment-2977889064)_

> Not sure if my observation belongs here or in a new issue. Testing https://github.com/bitcoin/bitcoin/pull/32297 I tried to use `socat` to inspect the IPC request:
> 
> `socat -v UNIX-LISTEN:/Users/matthewzipkin/.bitcoin/regtest/node.sock STDOUT`
> 
> and then make a bitcoin-cli request even though bitcoin-node was not running:
> 
> `bccli -regtest getblockcount`
> 
> error:
> ```
> libc++abi: terminating due to uncaught exception of type kj::ExceptionImpl: /opt/homebrew/include/kj/memory.h:258: failed: expected ptr != nullptr; null Own<> dereference
> stack: 1070007cb 10489843b 10494a81f 10494a74b 104951f4f 104951def 104951d93 104951d33 104951c73 104951c17 10494f59b 1048acccb 1048ac9d7 104ddc727 104ddc077 10488ab97 10488aa0f 10488a8e7 104889977 1a01b6f93 1a01b1d33
> Abort trap: 6
> ```
> 
> Stepping through with lldb I found the error in `ConnectStream()`:
> 
> https://github.com/bitcoin-core/libmultiprocess/blob/019839758085cb094f107e9ac354cbda79b388e2/include/mp/proxy-io.h#L559-L580
> 
> What I believe happens is, we get a file descriptor way earlier in the stack (in bitcoin: `src/ipc/interfaces.cpp::connectAddress()` but by the time we get to `ConnectStream()` the listener has disconnected (or, just isn't bitcoin) and then things start going wrong in `ProxyClientBase()`

ryanofsky:
It could be useful to see the backtrace from lldb, but depending on where the exception is being thrown, I think [#32345](https://github.com/bitcoin/bitcoin/pull/32345) could help with this by turning the `kj::Exception` into an `ipc::Exception` that could be more easily caught in bitcoin-cli. It would make sense for the `CallIPC` method in https://github.com/bitcoin/bitcoin/pull/32297 to catch this exception and treat it the same as the connection failing.

xyzconstant:
[...]

PR #298, opened 2026-06-24
Fix error handling when creating clients (`mp::ConnectStream`)

xyzconstant:
Avoid use-after-free if the socket is disconnected before `ConnectStream` connects (#308), and avoid leaks and hangs if client `construct()` calls throw (#309). Also add tests to cover these and other client connection errors, as [suggested](https://github.com/bitcoin-core/libmultiprocess/issues/183#issuecomment-3216736943) by @ryanofsky.

The following cases are tested:

1. Connecting to a socket serving a valid init interface
2. Passing a disconnected socket (`ConnectStream` throws during the `construct()` call)
3. Passing a disconnected socket to an interface without `construct()` (the failure is deferred to the first IPC request)
4. Passing a disconnected socket and making no calls (the disconnect is still handled and the connection cleaned up)
5. Passing a live socket that disconnects after some data is received
6. Passing a socket from a listening socket (`accept()`) that disconnects after some data arrives

Additionally, a new `FooInit` test interface is added, and the `UnixListener` class introduced in #269 is extracted to a shared file so the new `connect_tests.cpp` file can use it.

Note: Clients that own their connection now delete it on unexpected disconnects, so calls after a server disconnect fail with "called after disconnect" instead of "interrupted by disconnect" (one test.cpp assertion updated accordingly).

### DIFF
--- a/test/mp/test/connect_tests.cpp
+++ b/test/mp/test/connect_tests.cpp
@@ -1,0 +1,171 @@
+// Copyright (c) The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+#include "unixlistener.h"
+#include <kj/async.h>
+#include <kj/common.h>
+#include <kj/debug.h>
+#include <kj/memory.h>
+#include <kj/test.h>
+#include <mp/proxy.h>
+#include <mp/proxy-io.h>
+#include <mp/test/foo.capnp.h>
+#include <mp/test/foo.capnp.proxy.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include <cstring> // IWYU pragma: keep
+#include <functional>
+#include <future>
+#include <memory>
+#include <optional>
+#include <stdexcept>
+#include <string_view>
+#include <thread>
+
+namespace mp {
+namespace test {
+namespace {
+
+//! Default event loop log handler used by tests, throws so the calling code
+//! can assert on errors.
+void DefaultLogHandler(mp::LogMessage log)
+{
+    if (log.level == mp::Log::Raise)
+        throw std::runtime_error(log.message);
+}
+
+class TestSetup
+{
+public:
+    int client_fd;
+    int server_fd;
+
+    mp::EventLoop* loop;
+    std::optional<mp::EventLoopRef> loop_ref;
+    //! Thread variable should be after other struct members so the thread does
+    //! not start until the other members are initialized.
+    std::thread loop_thread;
+
+    TestSetup(mp::LogFn log_handler = DefaultLogHandler)
+        : TestSetup(
+              [](int fds[2]) {
+                  KJ_REQUIRE(socketpair(AF_UNIX, SOCK_STREAM, 0, fds) != -1);
+              },
+              log_handler) {}
+
+    TestSetup(const std::function<void(int[2])>& init_sockets,
+              mp::LogFn log_handler = DefaultLogHandler)
+    {
+        std::promise<mp::EventLoop*> loop_promise;
+        loop_thread = std::thread([&, log_handler] {
+            mp::EventLoop loop("mptest-connect", log_handler);
+            loop_promise.set_value(&loop);
+            loop.loop();
+        });
+        loop = loop_promise.get_future().get();
+        loop_ref.emplace(*loop);
+
+        // Initialize and store sockets
+        int fds[2] = {-1, -1};
+        init_sockets(fds);
+
+        client_fd = fds[0];
+        server_fd = fds[1];
+    }
+
+    ~TestSetup()
+    {
+        loop_ref.reset();
+        loop_thread.join();
+    }
+};
+
+KJ_TEST("ConnectStream connects to a socket serving a valid init interface")
+{
+    TestSetup setup;
+
+    std::thread server_thread([&setup]() {
+        mp::EventLoop server_loop("mptest-valid-server", DefaultLogHandler);
+        std::unique_ptr<FooInit> init = std::make_unique<FooInit>();
+        ServeStream<messages::FooInit>(server_loop, MakeStream(server_loop, setup.server_fd), *init);
+        server_loop.loop();
+    });
+
+    auto init = ConnectStream<messages::FooInit>(*setup.loop, MakeStream(*setup.loop, setup.client_fd));
+
+    init.reset();
+    server_thread.join();
+    KJ_EXPECT(true);
+}
+
+KJ_TEST("ConnectStream throws when the socket disconnects after receiving data")
+{
+    TestSetup setup;
+
+    std::thread server_thread([&setup]() {
+        char buf[128];
+
+        ssize_t bytes_received =
+            recv(setup.server_fd, buf, sizeof(buf), 0);
+
+        if (bytes_received > 0) {
+            close(setup.server_fd);
+        }
+    });
+
+    try {
+        auto init = ConnectStream<messages::FooInit>(*setup.loop, MakeStream(*setup.loop, setup.client_fd));
+
+        if (server_thread.joinable()) server_thread.join();
+        KJ_EXPECT(false);
+    } catch (const std::runtime_error& e) {
+        if (server_thread.joinable()) server_thread.join();
+
+        std::string_view reason = e.what();
+        KJ_EXPECT(reason == "IPC client method called after disconnect.");
+    }
+}
+
+KJ_TEST("ConnectStream throws when a connection accepted from a listener disconnects after receiving data")
+{
+    UnixListener listener;
+
+    TestSetup setup([&listener](int fds[2]) {
+        fds[0] = listener.MakeConnectedSocket(); // client_fd
+        fds[1] = listener.release();             // server_fd
+    });
+
+    std::thread server_thread([&setup]() {
+        char buf[128];
+
+        int connection_fd = accept(setup.server_fd, nullptr, nullptr);
+
+        if (connection_fd >= 0) {
+            ssize_t bytes_received =
+                recv(connection_fd, buf, sizeof(buf), 0);
+
+            if (bytes_received > 0) {
+                close(connection_fd);
+            }
+        }
+        close(setup.server_fd);
+    });
+
+    try {
+        auto init = ConnectStream<messages::FooInit>(*setup.loop, MakeStream(*setup.loop, setup.client_fd));
+
+        if (server_thread.joinable()) server_thread.join();
+        KJ_EXPECT(false);
+    } catch (const std::runtime_error& e) {
+        if (server_thread.joinable()) server_thread.join();
+
+        std::string_view reason = e.what();
+        KJ_EXPECT(reason == "IPC client method called after disconnect.");
+    }
+}
+
+} // namespace
+} // namespace test
+} // namespace mp
--- a/test/mp/test/connect_tests.cpp
+++ b/test/mp/test/connect_tests.cpp
@@ -15,18 +15,24 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#include <chrono>
+#include <condition_variable>
 #include <cstring> // IWYU pragma: keep
 #include <functional>
 #include <future>
 #include <memory>
+#include <mutex>
 #include <optional>
 #include <stdexcept>
+#include <string>
 #include <string_view>
 #include <thread>
 
 namespace mp {
 namespace test {
 namespace {
+
+constexpr auto FAILURE_TIMEOUT = std::chrono::seconds{30};
 
 //! Default event loop log handler used by tests, throws so the calling code
 //! can assert on errors.
@@ -100,6 +106,68 @@ KJ_TEST("ConnectStream connects to a socket serving a valid init interface")
     KJ_EXPECT(true);
 }
 
+KJ_TEST("ConnectStream throws when the socket is already disconnected")
+{
+    TestSetup setup;
+
+    close(setup.server_fd);
+
+    try {
+        auto init = ConnectStream<messages::FooInit>(*setup.loop, MakeStream(*setup.loop, setup.client_fd));
+
+        KJ_EXPECT(false);
+    } catch (const std::runtime_error& e) {
+        std::string_view reason = e.what();
+
+        KJ_EXPECT(reason == "IPC client method call interrupted by disconnect.");
+    }
+}
+
+KJ_TEST("ConnectStream defers disconnect failure to the first IPC request for interfaces without construct()")
+{
+    TestSetup setup;
+
+    close(setup.server_fd);
+
+    // Without a construct() method no IPC call is made during client
+    // creation, so ConnectStream succeeds even though the peer is gone.
+    auto foo = ConnectStream<messages::FooInterface>(*setup.loop, MakeStream(*setup.loop, setup.client_fd));
+
+    try {
+        foo->add(1, 2);
+        KJ_EXPECT(false);
+    } catch (const std::runtime_error& e) {
+        std::string_view reason = e.what();
+
+        KJ_EXPECT(reason == "IPC client method called after disconnect.");
+    }
+}
+
+KJ_TEST("ConnectStream handles a disconnect when no client calls are made")
+{
+    std::mutex mutex;
+    std::condition_variable cv;
+    bool warned = false;
+
+    TestSetup setup([&](mp::LogMessage log) {
+        if (log.level == mp::Log::Warning && log.message.find("unexpected network disconnect") != std::string::npos) {
+            const std::lock_guard<std::mutex> lock(mutex);
+            warned = true;
+            cv.notify_all();
+        }
+        DefaultLogHandler(log);
+    });
+
+    close(setup.server_fd);
+
+    auto foo = ConnectStream<messages::FooInterface>(*setup.loop, MakeStream(*setup.loop, setup.client_fd));
+
+    // The disconnect handler registered by ProxyClientBase should run and
+    // delete the connection even when no calls are ever made.
+    std::unique_lock<std::mutex> lock(mutex);
+    KJ_EXPECT(cv.wait_for(lock, FAILURE_TIMEOUT, [&] { return warned; }));
+}
+
 KJ_TEST("ConnectStream throws when the socket disconnects after receiving data")
 {
     TestSetup setup;
@@ -124,7 +192,7 @@ KJ_TEST("ConnectStream throws when the socket disconnects after receiving data")
         if (server_thread.joinable()) server_thread.join();
 
         std::string_view reason = e.what();
-        KJ_EXPECT(reason == "IPC client method called after disconnect.");
+        KJ_EXPECT(reason == "IPC client method call interrupted by disconnect.");
     }
 }
 
@@ -162,7 +230,7 @@ KJ_TEST("ConnectStream throws when a connection accepted from a listener disconn
         if (server_thread.joinable()) server_thread.join();
 
         std::string_view reason = e.what();
-        KJ_EXPECT(reason == "IPC client method called after disconnect.");
+        KJ_EXPECT(reason == "IPC client method call interrupted by disconnect.");
     }
 }
 
--- a/test/mp/test/foo.capnp
+++ b/test/mp/test/foo.capnp
@@ -39,6 +39,10 @@ interface FooInterface $Proxy.wrap("mp::test::FooImplementation") {
     passDataPointers @22 (arg :List(Data)) -> (result :List(Data));
 }
 
+interface FooInit $Proxy.wrap("mp::test::FooInit") {
+    construct @0 () -> ();
+}
+
 interface FooCallback $Proxy.wrap("mp::test::FooCallback") {
     destroy @0 (context :Proxy.Context) -> ();
     call @1 (context :Proxy.Context, arg :Int32) -> (result :Int32);
--- a/test/mp/test/test.cpp
+++ b/test/mp/test/test.cpp
@@ -288,7 +288,16 @@ KJ_TEST("Calling IPC method after server connection is closed")
     KJ_EXPECT(foo->add(1, 2) == 3);
     setup.server_disconnect();
 
-    EXPECT_EXCEPTION(foo->add(1, 2), "IPC client method call interrupted by disconnect.");
+    try {
+        foo->add(1, 2);
+        KJ_EXPECT(false);
+    } catch (const std::runtime_error& e) {
+        std::string_view reason{e.what()};
+
+        // The disconnect handler may delete the connection before the
+        // call is processed or while the call is in flight, both errors are possible.
+        KJ_EXPECT(reason == "IPC client method called after disconnect." || reason == "IPC client method call interrupted by disconnect.");
+    }
 }
 
 KJ_TEST("Calling IPC method and disconnecting during the call")
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -10,7 +10,6 @@
 #include <fstream>
 #include <future>
 #include <iostream>
-#include <kj/async.h>
 #include <kj/common.h>
 #include <memory>
 #include <mp/proxy.h>
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -592,7 +592,28 @@ ProxyClientBase<Interface, Impl>::ProxyClientBase(typename Interface::Client cli
         });
     }
     });
-    Sub::construct(*this);
+    // If construct() fails, run the cleanup functions before rethrowing,
+    // because ~ProxyClientBase will not run for an object whose constructor
+    // threw, and the connection would otherwise be leaked.
+    try {
+        Sub::construct(*this);
+    } catch (...) {
+        MP_LOG(*m_context.loop, Log::Debug) << "Cleaning up " << CxxTypeName(*this) << " " << this << " after construct() failure";
+        CleanupRun(m_context.cleanup_fns);
+        throw;
+    }
+
+    // If this client owns the connection, delete the connection on disconnect.
+    if (destroy_connection) {
+        m_context.loop->sync([&] {
+            EventLoop& loop = *m_context.loop;
+            Connection* connection = m_context.connection;
+            connection->onDisconnect([&loop, connection] {
+                MP_LOG(loop, Log::Warning) << "IPC client: unexpected network disconnect.";
+                delete connection;
+            });
+        });
+    }
 }
 
 template <typename Interface, typename Impl>
@@ -832,6 +853,10 @@ kj::Promise<T> ProxyServer<Thread>::post(Fn&& fn)
 //! Given a stream, make a new ProxyClient object to send requests over it.
 //! Also create a new Connection object embedded in the client that is freed
 //! when the client is closed.
+//!
+//! If the init interface declares a construct() method, creating the client
+//! calls it, so this function may block making an IPC call and may throw if
+//! the call fails.
 template <typename InitInterface>
 std::unique_ptr<ProxyClient<InitInterface>> ConnectStream(EventLoop& loop, Stream stream)
 {
@@ -840,11 +865,6 @@ std::unique_ptr<ProxyClient<InitInterface>> ConnectStream(EventLoop& loop, Strea
     loop.sync([&] {
         connection = std::make_unique<Connection>(loop, kj::mv(stream));
         init_client = connection->m_rpc_system->bootstrap(ServerVatId().vat_id).castAs<InitInterface>();
-        Connection* connection_ptr = connection.get();
-        connection->onDisconnect([&loop, connection_ptr] {
-            MP_LOG(loop, Log::Warning) << "IPC client: unexpected network disconnect.";
-            delete connection_ptr;
-        });
     });
     return std::make_unique<ProxyClient<InitInterface>>(
         kj::mv(init_client), connection.release(), /* destroy_connection= */ true);
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -28,6 +28,7 @@ if(BUILD_TESTING AND TARGET CapnProto::kj-test)
     mp/test/foo.h
     mp/test/listen_tests.cpp
     mp/test/spawn_tests.cpp
+    mp/test/connect_tests.cpp
     mp/test/test.cpp
   )
   include(${PROJECT_SOURCE_DIR}/cmake/TargetCapnpSources.cmake)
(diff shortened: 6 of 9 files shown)


ryanofsky:
Thanks for following up to #183 with these tests. They do seem potentially useful. Here is feedback I'd have:

- The first 3 tests seems like they are mostly in good shape. You would just want to wrap these in try/catch to and assert expected exceptions are thrown. For the second test it would seem fine to accept both "called after disconnect" and "interrupted by disconnect" exceptions since we already have existing tests testing for each of these errors more specifically.

- The first 3 tests do overlap a lot with disconnect tests already in test.cpp, and it's a little questionable if having all 3 tests adds much value. The first test could be is nice because it directly tests connecting to a non-capnp server that disconnects ignoring whatever is sent. But the second and third tests just sending the same disconnect at later points in time and needing MSG_PEEK complexity would not seem to add as much value.

- For the 4th test having the client hang as long as server hangs is probably expected behavior. It would be good to make sure that client can still disconnect or cancel the calls if the the server hangs. Or that client calls are able to time out correctly.

- It could make sense to rebase this on [#269](https://github.com/bitcoin-core/libmultiprocess/pull/269) or rebase if that could help with the "Add the case with an actual mkdtemp/socket/bind/listen setup" follow up comment.

- Maybe (not sure) it could be interesting to have tests working in opposite direction with server providing dummy interface and clients connecting and disconnecting suddenly. Maybe it could also be useful to have tests sending garbage bytes and making sure clients and servers handle them cleanly. It might even be useful to use fuzzing for this though it might not be a good use of fuzzing resources since it would mostly only be fuzz testing capnproto code and only a little bit of libmultiprocess exception/cleanup handling code.

Overall the tests here seem reasonable to add. It seems good to have at least 1-2 tests verifying disconnects are processed when a capnproto client connects to a non-capnproto server.

(Relatedly, there are also other disconnect tests that could be added at different points during capnproto connections, which I started to write in https://github.com/bitcoin-core/libmultiprocess/pull/201#issuecomment-3286839911 and https://github.com/ryanofsky/libmultiprocess/commits/pr/distest.2 but was never able to really finish due to complexity of trying to set up and cover all of the relevant cases. Just mentioning this for completeness, though. There's probably not an obviously place to follow up with this at the moment.)

xyzconstant:
Thanks for the feedback @ryanofsky!

I've just rebased to master now that #269 has been merged and added a new commit to move `UnixListener` to its own dedicated file.

Also, I've made changes to the tests (please check the description), keeping the first 2 tests (Notice that I've dropped the `MSG_PEEK` setup as you suggested.) now catching errors in a `try/catch` block, plus a Unix domain socket that disconnects after some data arrives and a successful case connecting to a valid `libmultiprocess` server.

>It would be good to make sure that client can still disconnect or cancel the calls if the the server hangs. Or that client calls are able to time out correctly.

This is my current focus. I'm looking more into it, but I think this work might need its own branch so we keep test coverage scoped to this PR.

>Maybe (not sure) it could be interesting to have tests working in opposite direction with server providing dummy interface and clients connecting and disconnecting suddenly. Maybe it could also be useful to have tests sending garbage bytes and making sure clients and servers handle them cleanly. It might even be useful to use fuzzing for this though it might not be a good use of fuzzing resources since it would mostly only be fuzz testing capnproto code and only a little bit of libmultiprocess exception/cleanup handling code.

Good, these ideas are great and definitely worth exploring. Happy to tackle them after this.

xyzconstant:
This PR is now ready for review. I've updated the description as well!

xyzconstant:
The netbsd 9.4 job exposed a use-after-free race condition in `ConnectStream` on early disconnects, included a [new commit](https://github.com/bitcoin-core/libmultiprocess/pull/298/changes/91f015eb71fbc03773f02d6177cb2d099a4bef4b) in this PR that fixes it.

test/mp/test/connect_tests.cpp
```diff
@@ -0,0 +1,182 @@
...
+            mp::EventLoop loop("mptest-connect", [](mp::LogMessage log) {
+                if (log.level == mp::Log::Raise)
+                    throw std::runtime_error(log.message);
+            });
+            loop_promise.set_value(&loop);
+            loop.loop();
+        });
+        loop = loop_promise.get_future().get();
+
+        // Initalize and store sockets
```
maflcko: LLM Linter (✨ experimental)

Possible typos and grammar issues:

    Initalize -> Initialize [misspelling in the socket setup comment; intended meaning is clear but the word is misspelled]

2026-07-10 05:29:42

xyzconstant: Thanks!

xyzconstant:
> The netbsd 9.4 job exposed a use-after-free race condition in `ConnectStream` on early disconnects, included a [new commit](https://github.com/bitcoin-core/libmultiprocess/pull/298/changes/91f015eb71fbc03773f02d6177cb2d099a4bef4b) in this PR that fixes it.

Updated the description expanding more on this issue.

xyzconstant:
Added the `FooInit` test interface which declares `construct()` so we could trigger that call against a disconnected socket in tests. It exposed a `Connection` leak, which, for instance, prevented the `EventLoop::loop()` from ever exiting.

Updated the description with more details about it and included a fix in the second commit.

ryanofsky:
Code review 96c32f39b8d9fac2c6b36ffea98d8ff2235112e6

Nice tests and fixes! The changes here look pretty good. I would just suggest a number of things to make this easier to understand and review:

- Since this PR is not only adding test coverage, would change title to something like "Fix error handling when creating clients" and give a description like "Avoid use-after-free if socket is disconnected before `ConnectStream` connects, and avoid leaks and hangs if client `construct` calls throw. Also add tests to cover these and other client connection errors"

- Would suggest adding the tests before making the bugfixes, otherwise it's not clear how the tests and bugfixes relate. It looks like both bugs are caught by the "Passing a disconnected socket will throw" test which is a very short test, and none of the other tests depend on fixes. So it would be clearer to add all the new tests except "Passing a disconnected socket will throw" initially, to be clear they are unrelated to the bugs.

- PR description is very long because it goes into detail about the history of the bugs and you how debugged them. I think this information is useful and interesting but it makes it harder to get a quick understanding, so I'd suggest moving this detail into separate github issues describing each bug and adding "Fixes #<issue number>" to the relevant fix commits so they will be closed when this PR is merged.

- It would be good if description of the bugs mentioned they are not new and have always been present. I found current description of the bug in 91f015eb71fbc03773f02d6177cb2d099a4bef4b that begins with "Previously," confusing because I thought it was referring to something that was previously working but now broken, when actually it's fixing something that has always been broken.

- I think it might be clearer to fix both bugs and add the short "Passing a disconnected socket will throw" all in a single commit. The first fix 91f015eb71fbc03773f02d6177cb2d099a4bef4b looks correct but it would seem nicer if it just let `ProxyClientBase` destructor be fully responsible for destroying the connection when `destroy_connection` is true instead of requiring the caller to destroy it. So could add something like `if (destroy_connection) { sync([&]{ connection->onDisconnect(...); } }` the end of the `ProxyClientBase` constructor to make `ConnectStream` simpler than it was before, instead of more complicated.

- I think it would be good to add a comment to `ConnectStream` that it calls the `InitInterface.construct` method if one is present, so it may block or throw.

xyzconstant:
re: https://github.com/bitcoin-core/libmultiprocess/pull/298#issuecomment-4963531240

Thanks for the review @ryanofsky!

Addressed your feedback and force-pushed, I hope the commit history is cleaner now.

Opened issues (#308 and #309) and updated the PR title and description as well.

Also, please note that I've added 2 new tests (I just noticed I pushed them right after your latest review, so you might have missed them), and I think they're valuable for showcasing the difference with an Init interface without `construct()`. They're included in the latest commit along with the single "disconnected socket" case.

xyzconstant:
Rebased on master which now includes #274. Adapted the tests to the stream API (`MakeStream`) and `TestSetup` now holds an `EventLoopRef` because it needs it to keep `loop()` running.

test/mp/test/foo.capnp
```diff
@@ -39,6 +39,10 @@ interface FooInterface $Proxy.wrap("mp::test::FooImplementation") {
     passDataPointers @22 (arg :List(Data)) -> (result :List(Data));
 }
 
+interface FooInit $Proxy.wrap("mp::test::FooInit") {
```
ryanofsky, line 42: In commit "Add test coverage for ConnectStream" (31c1ac2203b4f5ea3c68fc26672b93fcaae2c9a9)

Am curious why this new FooInit interface is needed and existing Foo interface isn't used.

It also seems like a potentially complicating factor that could make the tests harder to debug & understand for this to have a construct method. I wonder if it could be dropped or at least the Thread map parameters could be dropped since it doesn't look like anything in these tests requires threadmaps

EDIT: Oh, I see in next commit it looks like there are new tests that rely on the construct call failing. I think it would be to only use the FooInit type for the tests which actually need the construct method, and use FooInterface for other tests. Also would be good to drop ThreadMap parameters as I believe they should not be needed.

ryanofsky (commented):
Code review f3355b5e501a384aa335c58f6309432a8f900dd2. I need to take more time to understand the tests but the bugfix looks right and useful for improving the stability of the C++ IPC client if it connects to a server that disconnects right away.

I left a minor suggestion below. Also would note that this conflicts with #231 and while I think conflicts mostly just come from moved code, it could be useful to try merging the two PRs and making sure the new tests do not introduce any unix-isms.

test/mp/test/foo.capnp
xyzconstant: Nice suggestion!

I addressed it in 44d191420c6fdc83d2c1da159a409c4c18151cbe by removing the `ThreadMap` parameters.

xyzconstant:
Rebased with master and fixed surfaced IWYU issues.

Additionally, addressed @ryanofsky's feedback by removing the `ThreadMap` parameter from the (new) `FooInit`'s construct method. Thanks for the review!

test/mp/test/foo.capnp
xyzconstant: Regarding the FooInit/FooInterface split, the tests already follow this. Only the 4 tests that require the `construct()` call use `FooInit`.

Also, I've replaced the `initThreadMap` call with a simple `add(1, 2)` in the _"ConnectStream defers disconnect failure to the first IPC request for interfaces without construct()"_ test case.

test/mp/test/connect_tests.cpp
```diff
@@ -0,0 +1,171 @@
...
+void DefaultLogHandler(mp::LogMessage log)
+{
+    if (log.level == mp::Log::Raise)
+        throw std::runtime_error(log.message);
+}
+
+class TestSetup
+{
+public:
+    int client_fd;
```
ryanofsky, line 42: In commit "Add test coverage for ConnectStream" (44d191420c6fdc83d2c1da159a409c4c18151cbe)

All these class members should have `m_` prefixes so it is clear when they are accessed from methods that they are class members and not local variables. Lack of prefixes also makes code confusing below because there are two different variables called `loop`

test/mp/test/connect_tests.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include <string_view>
+#include <thread>
+
+namespace mp {
+namespace test {
+namespace {
+
+//! Default event loop log handler used by tests, throws so the calling code
+//! can assert on errors.
+void DefaultLogHandler(mp::LogMessage log)
```
ryanofsky, line 33: In commit "Add test coverage for ConnectStream" (44d191420c6fdc83d2c1da159a409c4c18151cbe)

This seems to be slightly different than the DefaultLogHandler defined in listen_tests.cpp. Would be nice to define a shared on, maybe in a test.h file.

test/mp/test/connect_tests.cpp
```diff
@@ -0,0 +1,171 @@
...
+        throw std::runtime_error(log.message);
+}
+
+class TestSetup
+{
+public:
+    int client_fd;
+    int server_fd;
+
+    mp::EventLoop* loop;
```
ryanofsky, line 45: In commit "Add test coverage for ConnectStream" (44d191420c6fdc83d2c1da159a409c4c18151cbe)

Would probably drop `mp::` prefix throughout this file since tests are in the `mp` namespace.

test/mp/test/connect_tests.cpp
```diff
@@ -0,0 +1,171 @@
...
+    std::thread loop_thread;
+
+    TestSetup(mp::LogFn log_handler = DefaultLogHandler)
+        : TestSetup(
+              [](int fds[2]) {
+                  KJ_REQUIRE(socketpair(AF_UNIX, SOCK_STREAM, 0, fds) != -1);
+              },
+              log_handler) {}
+
+    TestSetup(const std::function<void(int[2])>& init_sockets,
```
ryanofsky, line 58: In commit "Add test coverage for ConnectStream" (44d191420c6fdc83d2c1da159a409c4c18151cbe)

Having the init_sockets callback and the client_fd and server_fd members seems unnecessarily complicated given that nothing else in the test setup class uses them. Would seem simpler to drop these

<p>

```diff
--- a/test/mp/test/connect_tests.cpp
+++ b/test/mp/test/connect_tests.cpp
@@ -9,16 +9,15 @@
 #include <kj/test.h>
 #include <mp/proxy.h>
 #include <mp/proxy-io.h>
+#include <mp/util.h>
 #include <mp/test/foo.capnp.h>
 #include <mp/test/foo.capnp.proxy.h>
 #include <sys/socket.h>
-#include <sys/types.h>
 #include <unistd.h>
 
 #include <chrono>
 #include <condition_variable>
 #include <cstring> // IWYU pragma: keep
-#include <functional>
 #include <future>
 #include <memory>
 #include <mutex>
@@ -45,9 +44,6 @@ void DefaultLogHandler(mp::LogMessage log)
 class TestSetup
 {
 public:
-    int client_fd;
-    int server_fd;
-
     mp::EventLoop* loop;
     std::optional<mp::EventLoopRef> loop_ref;
     //! Thread variable should be after other struct members so the thread does
@@ -55,14 +51,6 @@ public:
     std::thread loop_thread;
 
     TestSetup(mp::LogFn log_handler = DefaultLogHandler)
-        : TestSetup(
-              [](int fds[2]) {
-                  KJ_REQUIRE(socketpair(AF_UNIX, SOCK_STREAM, 0, fds) != -1);
-              },
-              log_handler) {}
-
-    TestSetup(const std::function<void(int[2])>& init_sockets,
-              mp::LogFn log_handler = DefaultLogHandler)
     {
         std::promise<mp::EventLoop*> loop_promise;
         loop_thread = std::thread([&, log_handler] {
@@ -72,13 +60,6 @@ public:
         });
         loop = loop_promise.get_future().get();
         loop_ref.emplace(*loop);
-
-        // Initialize and store sockets
-        int fds[2] = {-1, -1};
-        init_sockets(fds);
-
-        client_fd = fds[0];
-        server_fd = fds[1];
     }
 
     ~TestSetup()
@@ -91,15 +72,16 @@ public:
 KJ_TEST("ConnectStream connects to a socket serving a valid init interface")
 {
     TestSetup setup;
+    auto [client_fd, server_fd] = SocketPair();
 
-    std::thread server_thread([&setup]() {
+    std::thread server_thread([&]() {
         mp::EventLoop server_loop("mptest-valid-server", DefaultLogHandler);
         std::unique_ptr<FooInit> init = std::make_unique<FooInit>();
-        ServeStream<messages::FooInit>(server_loop, MakeStream(server_loop, setup.server_fd), *init);
+        ServeStream<messages::FooInit>(server_loop, MakeStream(server_loop, server_fd), *init);
         server_loop.loop();
     });
 
-    auto init = ConnectStream<messages::FooInit>(*setup.loop, MakeStream(*setup.loop, setup.client_fd));
+    auto init = ConnectStream<messages::FooInit>(*setup.loop, MakeStream(*setup.loop, client_fd));
 
     init.reset();
     server_thread.join();
@@ -109,11 +91,12 @@ KJ_TEST("ConnectStream connects to a socket serving a valid init interface")
 KJ_TEST("ConnectStream throws when the socket is already disconnected")
 {
     TestSetup setup;
+    auto [client_fd, server_fd] = SocketPair();
 
-    close(setup.server_fd);
+    close(server_fd);
 
     try {
-        auto init = ConnectStream<messages::FooInit>(*setup.loop, MakeStream(*setup.loop, setup.client_fd));
+        auto init = ConnectStream<messages::FooInit>(*setup.loop, MakeStream(*setup.loop, client_fd));
 
         KJ_EXPECT(false);
     } catch (const std::runtime_error& e) {
@@ -126,12 +109,13 @@ KJ_TEST("ConnectStream throws when the socket is already disconnected")
 KJ_TEST("ConnectStream defers disconnect failure to the first IPC request for interfaces without construct()")
 {
     TestSetup setup;
+    auto [client_fd, server_fd] = SocketPair();
 
-    close(setup.server_fd);
+    close(server_fd);
 
     // Without a construct() method no IPC call is made during client
     // creation, so ConnectStream succeeds even though the peer is gone.
-    auto foo = ConnectStream<messages::FooInterface>(*setup.loop, MakeStream(*setup.loop, setup.client_fd));
+    auto foo = ConnectStream<messages::FooInterface>(*setup.loop, MakeStream(*setup.loop, client_fd));
 
     try {
         foo->add(1, 2);
@@ -157,10 +141,11 @@ KJ_TEST("ConnectStream handles a disconnect when no client calls are made")
         }
         DefaultLogHandler(log);
     });
+    auto [client_fd, server_fd] = SocketPair();
 
-    close(setup.server_fd);
+    close(server_fd);
 
-    auto foo = ConnectStream<messages::FooInterface>(*setup.loop, MakeStream(*setup.loop, setup.client_fd));
+    auto foo = ConnectStream<messages::FooInterface>(*setup.loop, MakeStream(*setup.loop, client_fd));
 
     // The disconnect handler registered by ProxyClientBase should run and
     // delete the connection even when no calls are ever made.
@@ -171,20 +156,21 @@ KJ_TEST("ConnectStream handles a disconnect when no client calls are made")
 KJ_TEST("ConnectStream throws when the socket disconnects after receiving data")
 {
     TestSetup setup;
+    auto [client_fd, server_fd] = SocketPair();
 
-    std::thread server_thread([&setup]() {
+    std::thread server_thread([&]() {
         char buf[128];
 
         ssize_t bytes_received =
-            recv(setup.server_fd, buf, sizeof(buf), 0);
+            recv(server_fd, buf, sizeof(buf), 0);
 
         if (bytes_received > 0) {
-            close(setup.server_fd);
+            close(server_fd);
         }
     });
 
     try {
-        auto init = ConnectStream<messages::FooInit>(*setup.loop, MakeStream(*setup.loop, setup.client_fd));
+        auto init = ConnectStream<messages::FooInit>(*setup.loop, MakeStream(*setup.loop, client_fd));
 
         if (server_thread.joinable()) server_thread.join();
         KJ_EXPECT(false);
@@ -199,16 +185,14 @@ KJ_TEST("ConnectStream throws when the socket disconnects after receiving data")
 KJ_TEST("ConnectStream throws when a connection accepted from a listener disconnects after receiving data")
 {
     UnixListener listener;
+    TestSetup setup;
+    int client_fd = listener.MakeConnectedSocket();
+    int server_fd = listener.release();
 
-    TestSetup setup([&listener](int fds[2]) {
-        fds[0] = listener.MakeConnectedSocket(); // client_fd
-        fds[1] = listener.release();             // server_fd
-    });
-
-    std::thread server_thread([&setup]() {
+    std::thread server_thread([&]() {
         char buf[128];
 
-        int connection_fd = accept(setup.server_fd, nullptr, nullptr);
+        int connection_fd = accept(server_fd, nullptr, nullptr);
 
         if (connection_fd >= 0) {
             ssize_t bytes_received =
@@ -218,11 +202,11 @@ KJ_TEST("ConnectStream throws when a connection accepted from a listener disconn
                 close(connection_fd);
             }
         }
-        close(setup.server_fd);
+        close(server_fd);
     });
 
     try {
-        auto init = ConnectStream<messages::FooInit>(*setup.loop, MakeStream(*setup.loop, setup.client_fd));
+        auto init = ConnectStream<messages::FooInit>(*setup.loop, MakeStream(*setup.loop, client_fd));
 
         if (server_thread.joinable()) server_thread.join();
         KJ_EXPECT(false);
```
</p>
</details>

test/mp/test/connect_tests.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+        if (bytes_received > 0) {
+            close(setup.server_fd);
+        }
+    });
+
+    try {
+        auto init = ConnectStream<messages::FooInit>(*setup.loop, MakeStream(*setup.loop, setup.client_fd));
+
+        if (server_thread.joinable()) server_thread.join();
```
ryanofsky, line 121: In commit "Add test coverage for ConnectStream" (44d191420c6fdc83d2c1da159a409c4c18151cbe)

I don't think it makes sense to call `joinable` here and to repeat this same line. It would make more sense to simply call `server_thread.join()` unconditionally at the end of this function after the try/catch. 

Same comment also applies to test below

test/mp/test/connect_tests.cpp
```diff
@@ -0,0 +1,171 @@
...
+        std::unique_ptr<FooInit> init = std::make_unique<FooInit>();
+        ServeStream<messages::FooInit>(server_loop, MakeStream(server_loop, setup.server_fd), *init);
+        server_loop.loop();
+    });
+
+    auto init = ConnectStream<messages::FooInit>(*setup.loop, MakeStream(*setup.loop, setup.client_fd));
+
+    init.reset();
+    server_thread.join();
+    KJ_EXPECT(true);
```
ryanofsky, line 100: In commit "Add test coverage for ConnectStream" (44d191420c6fdc83d2c1da159a409c4c18151cbe)

Expecting true here seems unnecessary.

test/mp/test/connect_tests.cpp
```diff
@@ -0,0 +1,171 @@
...
+    TestSetup setup;
+
+    std::thread server_thread([&setup]() {
+        char buf[128];
+
+        ssize_t bytes_received =
+            recv(setup.server_fd, buf, sizeof(buf), 0);
+
+        if (bytes_received > 0) {
+            close(setup.server_fd);
```
ryanofsky, line 114: In commit "Add test coverage for ConnectStream" (44d191420c6fdc83d2c1da159a409c4c18151cbe)

I don't understand the reason for only closing the descriptor if bytes or received (and leaking otherwise)? Would make more sense to close it unconditionally. Same applies to test below. If there is a reason for this conditional it would be good to explain in a comment

test/mp/test/connect_tests.cpp
```diff
@@ -0,0 +1,171 @@
...
+    TestSetup setup;
+
+    std::thread server_thread([&setup]() {
+        mp::EventLoop server_loop("mptest-valid-server", DefaultLogHandler);
+        std::unique_ptr<FooInit> init = std::make_unique<FooInit>();
+        ServeStream<messages::FooInit>(server_loop, MakeStream(server_loop, setup.server_fd), *init);
+        server_loop.loop();
+    });
+
+    auto init = ConnectStream<messages::FooInit>(*setup.loop, MakeStream(*setup.loop, setup.client_fd));
```
ryanofsky, line 96: In commit "Add test coverage for ConnectStream" (44d191420c6fdc83d2c1da159a409c4c18151cbe)

Might be good to note in a comment that `FooInit` capnproto interface has a construct method, so this is not just connecting to the IPC server, but also testing that the construct IPC request completes successfully.

test/mp/test/connect_tests.cpp
```diff
@@ -100,6 +106,68 @@ KJ_TEST("ConnectStream connects to a socket serving a valid init interface")
     KJ_EXPECT(true);
 }
 
+KJ_TEST("ConnectStream throws when the socket is already disconnected")
+{
+    TestSetup setup;
+
+    close(setup.server_fd);
```
ryanofsky, line 113: In commit "Fix error handling when creating clients" (bb473690c97ceed78a482d6a97b480cec4a63190)

Would be good to use KJ_SYSCALL here and below to check close return value

test/mp/test/connect_tests.cpp
```diff
@@ -162,7 +230,7 @@ KJ_TEST("ConnectStream throws when a connection accepted from a listener disconn
         if (server_thread.joinable()) server_thread.join();
 
         std::string_view reason = e.what();
-        KJ_EXPECT(reason == "IPC client method called after disconnect.");
+        KJ_EXPECT(reason == "IPC client method call interrupted by disconnect.");
```
ryanofsky, line 233: In commit "Fix error handling when creating clients" (bb473690c97ceed78a482d6a97b480cec4a63190)

It would be helpful if commit message noted the reason existing tests in this commit are changing, that because the `onDisconnect` call now happens later, client code might detect disconnects happening during IPC calls instead of before them.

test/mp/test/test.cpp
```diff
@@ -288,7 +288,16 @@ KJ_TEST("Calling IPC method after server connection is closed")
...
-    EXPECT_EXCEPTION(foo->add(1, 2), "IPC client method call interrupted by disconnect.");
+    try {
+        foo->add(1, 2);
+        KJ_EXPECT(false);
+    } catch (const std::runtime_error& e) {
+        std::string_view reason{e.what()};
+
+        // The disconnect handler may delete the connection before the
+        // call is processed or while the call is in flight, both errors are possible.
+        KJ_EXPECT(reason == "IPC client method called after disconnect." || reason == "IPC client method call interrupted by disconnect.");
```
ryanofsky, line 299: In commit "Fix error handling when creating clients" (bb473690c97ceed78a482d6a97b480cec4a63190)

I'm confused about this change. I don't understand if disconnects are detected later now, why the "interrupted by disconnect" message would change to an "interrupted after connect". Would be good to have a more specific comment explaining this, or possibly reverting the change here if not necessary.

EDIT: Reason for this changes seems to be that `test.cpp` does not use ConnectStream so no disconnect handler was ever registered previously, and now one is registed since the disconnect handler was moved to ProxyClientBase. Would be helpful if commit message mentioned that as reason this test is changing.

test/mp/test/connect_tests.cpp
```diff
@@ -100,6 +106,68 @@ KJ_TEST("ConnectStream connects to a socket serving a valid init interface")
...
+    });
+
+    close(setup.server_fd);
+
+    auto foo = ConnectStream<messages::FooInterface>(*setup.loop, MakeStream(*setup.loop, setup.client_fd));
+
+    // The disconnect handler registered by ProxyClientBase should run and
+    // delete the connection even when no calls are ever made.
+    std::unique_lock<std::mutex> lock(mutex);
+    KJ_EXPECT(cv.wait_for(lock, FAILURE_TIMEOUT, [&] { return warned; }));
```
ryanofsky, line 168: In commit "Fix error handling when creating clients" (bb473690c97ceed78a482d6a97b480cec4a63190)

Note: technically it would be possible to simplify this test by having it wait for the client connection pointer to be null, instead of waiting for the "unexpected network disconnect" log message. But that approach would stop working after #336 which changes connections to use `shared_ptr` and no longer track lists of clients client objects. So it makes sense to keep the current approach.

ryanofsky (approved):
Code review ACK fae9a637e35df7020c5f42677e9f089754a95d2c. I left a lot of comments here about the tests, but the fix itself looks very good and this could be merged as-is.

test/mp/test/connect_tests.cpp
xyzconstant: Addressed this in https://github.com/bitcoin-core/libmultiprocess/pull/344/changes/70467c5a727ba50789865b0323f313167d2e500b (follow-up PR).

test/mp/test/connect_tests.cpp
xyzconstant: Nice suggestion! I moved this logger to a `common.h` file instead at https://github.com/bitcoin-core/libmultiprocess/pull/344/changes/ae5c6bc8637b79968b2fced5a556e2ca55d8af63 (follow-up PR).

test/mp/test/connect_tests.cpp
xyzconstant: Nice, addressed at https://github.com/bitcoin-core/libmultiprocess/pull/344/changes/5790becceac1e27b4f4c2bb6f2e81c4d684fef82 (follow-up PR).

test/mp/test/connect_tests.cpp
xyzconstant: Yeah, you're right, having them was indeed unnecessary. Addressed this at https://github.com/bitcoin-core/libmultiprocess/pull/344/changes/466afd0ecb4fc802ce28eaca7f4b02aad6efe1a3 (follow-up PR).

Thanks for the diff!

test/mp/test/connect_tests.cpp
xyzconstant: Nice catch. Addressed at https://github.com/bitcoin-core/libmultiprocess/pull/344/changes/24f17bd38a2ad4842f2e3c511ab50dd190ea1957 (follow-up PR).

test/mp/test/connect_tests.cpp
xyzconstant: Dropped at https://github.com/bitcoin-core/libmultiprocess/pull/344/changes/8d30ea847be9d0d9d267fdd22a127cea1f8ad9d6 (follow-up PR).

test/mp/test/connect_tests.cpp
xyzconstant: Good suggestion. Honestly, when I wrote this I thought `recv()` might somewhat "read 0 bytes", but I see now that I had it wrong. There's no reason to have this conditional, so dropped it at https://github.com/bitcoin-core/libmultiprocess/pull/344/changes/29770728c386b8aa700b863babc555d185d8a287 (follow-up PR).

test/mp/test/connect_tests.cpp
xyzconstant: Nice, done at https://github.com/bitcoin-core/libmultiprocess/pull/344/changes/76ab72f37d55f9512408ad292da1d00a059f3268 (follow-up PR).

test/mp/test/connect_tests.cpp
xyzconstant: Done at https://github.com/bitcoin-core/libmultiprocess/commit/29770728c386b8aa700b863babc555d185d8a287 (follow-up PR).

test/mp/test/connect_tests.cpp
xyzconstant: Sorry, couldn't reword the commit in time. Thanks for the feedback, though.

xyzconstant:
Thanks for the review @ryanofsky!

Unfortunately, I couldn't address your feedback in time before this PR was merged. Opened a follow-up PR to address your suggestions: #344.

I replied to some of your inline comments with links to the commit in the follow-up PR that tackles it.

PR #300, opened 2026-06-30
Server-wide ThreadPool

Eunovo:
This PR adds a `ThreadPool` owned by the `EventLoop`, enabling any server connection to dispatch work to a shared pool. The `ThreadPool` is tested with callbacks and recursive IPC calls.

**Motivation**
- Reduces boilerplate for non-libmultiprocess clients that don't need control over where work executes on the server (similar goal to https://github.com/bitcoin-core/libmultiprocess/pull/283).
- Provides a straightforward way to cap the server's total thread count.

**Difference from https://github.com/bitcoin-core/libmultiprocess/pull/283**
Unlike the approach in https://github.com/bitcoin-core/libmultiprocess/pull/283, external clients don't need to send a pool creation request — the pool is available server-wide from the start. This lays groundwork for a future where `ThreadMap` access is restricted to privileged connections only.

---

*Draft while I finish testing and gauge interest in a server-wide `ThreadPool`.*

### DIFF
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -156,6 +156,31 @@ public:
     WorkerThread m_worker;
 };
 
+//! Fixed-size pool of WorkerThreads. Requests are dispatched round-robin
+//! across the pool. Owned by EventLoop when num_pool_threads > 0.
+class ThreadPool
+{
+public:
+    ThreadPool(EventLoop& loop, int num_threads);
+
+    //! Dispatch fn to a pool WorkerThread (round-robin). Returns a promise
+    //! fulfilled on the event-loop thread when fn returns.
+    template<typename T, typename Fn>
+    kj::Promise<T> post(Fn&& fn);
+
+private:
+    std::vector<std::unique_ptr<WorkerThread>> m_threads;
+    size_t m_next_thread{0};
+};
+
+template<typename T, typename Fn>
+kj::Promise<T> ThreadPool::post(Fn&& fn)
+{
+    WorkerThread& worker = *m_threads[m_next_thread % m_threads.size()];
+    ++m_next_thread;
+    return worker.post<T>(std::forward<Fn>(fn));
+}
+
 //! Handler for kj::TaskSet failed task events.
 class LoggingErrorHandler : public kj::TaskSet::ErrorHandler
 {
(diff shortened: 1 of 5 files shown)


ryanofsky:
Thanks for the PR! Seems like there are a lot of nice things here: New recursive async test looks worthwhile (though I think it might be possible to simplify by using passFn/std::function instead of callback/Callback). Test cleanups and WorkerThread refactor also seem helpful and make code more readable.

Things I'd change:

- I think thread pools should be per-connection instead of per-eventloop. If there are only a limited number of threads the threads can easily get backed up or deadlocked with only a small number of blocking calls. If there is a shared thread pool then requests from one connection can block requests from a unrelated connections, which seems like it could cause unwanted interactions between clients and difficult to debug problems.

- Relatedly I think clients should choose number of threads in their thread pools rather than leaving it to the server to decide. Clients need to be trusted anyway right now and only clients can have a good idea of how many threads they may need. I don't think the server or server administrator could do much better than blind guessing. That's not to say the server couldn't have **caps** on number of threads clients create, but either way clients should be explicit about their needs and expectations.

- Similar to [#283](https://github.com/bitcoin-core/libmultiprocess/pull/283), scheduling algorithm does not seem very good because it just assigns tasks to threads whether the threads are blocked or not. A better approach would be to use a shared queue that threads pull from when they are idle as implemented in  bd1f80c31b95c2041bd5feb13ae710a78698dcec ([branch](https://github.com/ryanofsky/libmultiprocess/commits/pr/tpool)) from [#283 (comment)](https://github.com/bitcoin-core/libmultiprocess/pull/283#issuecomment-4732869389)

- I'm suspicious of the "getLocalServer call failed" error message and comment changing in a8e54134fa3a20437f6d14a3bbef0dec6c02c6cd. The goal of that error message is not to be triggered when `context_arg.getThread()` capability is null. (A better way to check whether the capability is null is to call `context_arg.hasThread()` instead.) The goal of the error message is to be triggered when `context_arg.getThread()` is an invalid capability promise passed from the client to the server using capnproto's [request pipelining](https://capnproto.org/rpc.html) feature. An invalid capability promise can be passed if the client calls `ThreadMap.makeThread` on a null ThreadMap object, and the the call is pipelined.

Overall, I think the solution for providing more automatic threading will combine different aspects of all this approach, bd1f80c31b95c2041bd5feb13ae710a78698dcec, and #283. I'm planning on merging #283 first in its current form (and bumping the libmultiprocess subtree in bitcoin core) and this PR could be a good followup to that.

PR #301, opened 2026-07-01
test: recursive async IPC calls and cleanups

Eunovo:
Test that async processing threads correctly handle incoming requests while waiting on another.
Also use the opportunity to refactor some of the tests to remove the try-catch blocks.
Broken out from https://github.com/bitcoin-core/libmultiprocess/pull/300

### DIFF
--- a/test/mp/test/test.cpp
+++ b/test/mp/test/test.cpp
@@ -38,6 +38,15 @@
 #include <utility>
 #include <vector>
 
+//! Assert that a call throws std::runtime_error with the given message.
+#define EXPECT_EXCEPTION(call, message)                                           \
+    try {                                                                         \
+        call;                                                                     \
+        KJ_EXPECT(false);                                                         \
+    } catch (const std::runtime_error& e) {                                       \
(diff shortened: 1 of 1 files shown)


test/mp/test/test.cpp
```diff
@@ -38,6 +38,15 @@
 #include <utility>
 #include <vector>
 
+//! Assert that a call throws std::runtime_error with the given message.
+#define EXPECT_EXCEPTION(call, message)                                           \
+    try {                                                                         \
+        call;                                                                     \
+        KJ_EXPECT(false);                                                         \
```
ryanofsky, line 45: In commit "test: recursive async IPC calls" (5519f7f9485c342d7c868846985750d6f152f359)

Could be better to improve output in this case with something like `KJ_FAIL_EXPECT("Expected exception was not thrown: " message)`

ryanofsky (approved):
Code review ACK 5519f7f9485c342d7c868846985750d6f152f359. Thanks for the new test and test cleanup!

PR #302, opened 2026-07-02
refactor: rename EventLoop::m_num_clients to m_num_refs

Sjors:
Rename the variable and document what it counts, so the `loop()` exit condition in `EventLoop::done()` is not misread as "exit when no clients are connected".

I found myself confused by this while reviewing #269.

### DIFF
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -255,7 +255,7 @@ public:
 
     //! Run event loop. Does not return until shutdown. This should only be
     //! called once from the m_thread_id thread. This will block until
-    //! the m_num_clients reference count is 0.
+    //! the m_num_refs reference count is 0.
     void loop();
 
     //! Run function on event loop thread. Does not return until function completes.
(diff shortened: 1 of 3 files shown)


include/mp/proxy-io.h
```diff
@@ -255,7 +255,7 @@ class EventLoop
 
     //! Run event loop. Does not return until shutdown. This should only be
     //! called once from the m_thread_id thread. This will block until
-    //! the m_num_clients reference count is 0.
+    //! the m_num_refs reference count is 0.
```
ViniciusCestarii, line 258: In "refactor: rename EventLoop::m_num_clients to m_num_refs"  d499830415405ba442f597cb06ed4c56442f774a

nit: "the m_num_refs reference count" is a bit redundant now because "refs" already means references. Could be renamed to "the m_num_refs count is 0" (unrelated but  this condition is also incomplete: loop() exits on done(), m_num_refs == 0 and m_async_fns is empty).

ViniciusCestarii (commented):
ACK d499830415405ba442f597cb06ed4c56442f774a, I agree this name is clearer.

ryanofsky (approved):
Code review ACK d499830415405ba442f597cb06ed4c56442f774a. Thanks for the rename and comment!

This was previously discussed https://github.com/bitcoin-core/libmultiprocess/pull/107#issuecomment-2299818498

>Probably the word "client" should be replaced with "use count" here (in `num_clients` `addClient` `removeClient`). The word "client" is supposed to mean "client of the event loop" but that seems unnecessarily confusing because the word client is overloaded so much.

And https://github.com/bitcoin-core/libmultiprocess/pull/107#issuecomment-2301338227

> I didn't find this that confusing to be honest, and just by looking at the names was assuming "client of the event loop".

But the word was used more places then so renaming is easier now and makes sense.

PR #303, opened 2026-07-08
type-chrono: Add CustomBuildField/CustomReadField overloads for std::chrono::time_point

ryanofsky:
Needed for bitcoin/bitcoin#10102 since bitcoin/bitcoin#34882 was merged, which uses `NodeClock::time_point` in the `CNodeStats` struct (`m_last_send`, `m_last_recv`, `m_ping_start`) returned by `Node::getNodesStats`.

### DIFF
--- a/include/mp/type-chrono.h
+++ b/include/mp/type-chrono.h
@@ -16,9 +16,9 @@ template <class Rep, class Period, typename Value, typename Output>
 void CustomBuildField(TypeList<std::chrono::duration<Rep, Period>>, Priority<1>, InvokeContext& invoke_context, Value&& value,
                       Output&& output)
 {
-    static_assert(std::numeric_limits<decltype(output.get())>::lowest() <= std::numeric_limits<Rep>::lowest(),
+    static_assert(safe_less_equal(std::numeric_limits<decltype(output.get())>::lowest(), std::numeric_limits<Rep>::lowest()),
                   "capnp type does not have enough range to hold lowest std::chrono::duration value");
-    static_assert(std::numeric_limits<decltype(output.get())>::max() >= std::numeric_limits<Rep>::max(),
+    static_assert(safe_greater_equal(std::numeric_limits<decltype(output.get())>::max(), std::numeric_limits<Rep>::max()),
                   "capnp type does not have enough range to hold highest std::chrono::duration value");
     output.set(value.count());
 }
--- a/include/mp/type-chrono.h
+++ b/include/mp/type-chrono.h
@@ -29,6 +29,29 @@ decltype(auto) CustomReadField(TypeList<std::chrono::duration<Rep, Period>>, Pri
 {
     return read_dest.construct(input.get());
 }
+
+//! Overload CustomBuildField and CustomReadField to serialize
+//! std::chrono::time_point parameters and return values as integer tick counts.
+//! The capnp field type must be an integer type with enough range to hold the
+//! time_since_epoch() count for the given Duration (e.g. Int64 for nanoseconds).
+template <class Clock, class Duration, typename Value, typename Output>
+void CustomBuildField(TypeList<std::chrono::time_point<Clock, Duration>>, Priority<1>, InvokeContext& invoke_context,
+                      Value&& value, Output&& output)
+{
+    using Rep = typename Duration::rep;
+    static_assert(safe_less_equal(std::numeric_limits<decltype(output.get())>::lowest(), std::numeric_limits<Rep>::lowest()),
+                  "capnp type does not have enough range to hold lowest std::chrono::time_point value");
+    static_assert(safe_greater_equal(std::numeric_limits<decltype(output.get())>::max(), std::numeric_limits<Rep>::max()),
+                  "capnp type does not have enough range to hold highest std::chrono::time_point value");
+    output.set(value.time_since_epoch().count());
+}
+
+template <class Clock, class Duration, typename Input, typename ReadDest>
+decltype(auto) CustomReadField(TypeList<std::chrono::time_point<Clock, Duration>>, Priority<1>,
+                               InvokeContext& invoke_context, Input&& input, ReadDest&& read_dest)
+{
+    return read_dest.construct(std::chrono::time_point<Clock, Duration>{Duration{input.get()}});
+}
 } // namespace mp
 
 #endif // MP_PROXY_TYPE_CHRONO_H
--- a/include/mp/util.h
+++ b/include/mp/util.h
@@ -33,6 +33,21 @@ namespace mp {
 
 //! Generic utility functions used by capnp code.
 
+// std::cmp_less_equal/cmp_greater_equal only accept integer types (C++20 §[utility.intcmp]).
+// These wrappers fall back to regular comparison for floating-point types.
+template <typename A, typename B>
+constexpr bool safe_less_equal(A a, B b)
+{
+    if constexpr (std::is_floating_point_v<A> || std::is_floating_point_v<B>) return a <= b;
+    else return std::cmp_less_equal(a, b);
+}
+template <typename A, typename B>
+constexpr bool safe_greater_equal(A a, B b)
+{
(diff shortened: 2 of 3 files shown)


maflcko:
lgtm ACK b37d1d71f115f6b5a0000e09c7be063994387e1a

Sjors:
> Needed by https://github.com/bitcoin/bitcoin/pull/34882

That's merged, by I assume you still need this?

maflcko:
> > Needed by [bitcoin/bitcoin#34882](https://github.com/bitcoin/bitcoin/pull/34882)
> 
> That's merged, by I assume you still need this?

There was no rebase of https://github.com/bitcoin/bitcoin/pull/10102 in 4 months, so I don't think https://github.com/bitcoin/bitcoin/pull/10102 sees the changes from 34882 yet.

ryanofsky:
Sorry PR description was written in a confusing way, should hopefully be clearer now. In order for bitcoin/bitcoin#10102 to work after bitcoin/bitcoin#34882 it needs these `ReadField` / `BuildField` overloads. It could define them itself, but it makes sense for libmultiprocess to provide them since it already provides many similar ones including ones for `std::chrono::duration`

include/mp/type-chrono.h
```diff
@@ -29,6 +29,29 @@ decltype(auto) CustomReadField(TypeList<std::chrono::duration<Rep, Period>>, Pri
...
+//! Overload CustomBuildField and CustomReadField to serialize
+//! std::chrono::time_point parameters and return values as integer tick counts.
+//! The capnp field type must be an integer type with enough range to hold the
+//! time_since_epoch() count for the given Duration (e.g. Int64 for nanoseconds).
+template <class Clock, class Duration, typename Value, typename Output>
+void CustomBuildField(TypeList<std::chrono::time_point<Clock, Duration>>, Priority<1>, InvokeContext& invoke_context,
+                      Value&& value, Output&& output)
+{
+    using Rep = typename Duration::rep;
+    static_assert(std::numeric_limits<decltype(output.get())>::lowest() <= std::numeric_limits<Rep>::lowest(),
```
ViniciusCestarii: The range checks in both CustomBuildField overloads have a signedness bug:

```cpp
static_assert(std::numeric_limits<decltype(output.get())>::lowest() <= std::numeric_limits<Rep>::lowest(), ...);
```

When the capnp field type is unsigned (e.g. UInt64) and Rep is signed (e.g. int64_t for std::chrono::nanoseconds), the usual arithmetic conversions convert Rep::lowest() (INT64_MIN) to the unsigned type before comparing, turning it into a huge positive number. So 0 <= INT64_MIN silently evaluates to true, and the assert passes for a field that can't actually represent negative tick counts (e.g. any pre-epoch system_clock::time_point).

Concretely, this means a UInt64 field paired with a signed Duration::rep compiles today, and output.set(negative_count) wraps the value into UINT64_MAX. That's silently wrong for anything that reads the field as its declared (unsigned) type — a different language's capnp bindings, a JSON dump, or a raw comparison would see 18446744073709551615 instead of -1. A C++ round trip through the same chrono type can look fine, since the wrap is bit-for-bit invertible at equal width, which is what makes this easy to miss in testing.

Fix: use std::cmp_less_equal/std::cmp_greater_equal (C++20, <utility>) instead of <=/>=, since they're designed to compare integers of differing signedness correctly:

```cpp
static_assert(std::cmp_less_equal(std::numeric_limits<decltype(output.get())>::lowest(), std::numeric_limits<Rep>::lowest()),
              "capnp type does not have enough range to hold lowest std::chrono::time_point value");
static_assert(std::cmp_greater_equal(std::numeric_limits<decltype(output.get())>::max(), std::numeric_limits<Rep>::max()),
              "capnp type does not have enough range to hold highest std::chrono::time_point value");
```

I verified this by temporarily declaring a test field as UInt64 against a signed Rep: it compiled before the fix and correctly fails to compile after.

Happy to push chrono type tests as a follow-up PR if useful.

Note: I had AI help me write this up clearly, apologies if the phrasing feels more polished or sloppy than my usual comments.

ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/303#discussion_r3551873101

Good catch! I think this makes sense and it looks like this same problem also exists other places: for durations above and also in type-number.h. I'll see if it's possible to fix them all here without breaking anything.

ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/303#discussion_r3551873101

Added a new commit to fix this problem. Using std::cmp functions directly didn't work for time_point comparisons because floating point time point types are used in some places so I had to add wrappers to handle floats.

ryanofsky (commented):
Thanks for the reviews!


Updated b37d1d71f115f6b5a0000e09c7be063994387e1a -> 978028ffd0d14e5e46b313412ef17b7ad3647e33 ([`pr/timepoint.1`](https://github.com/ryanofsky/libmultiprocess/commits/pr/timepoint.1) -> [`pr/timepoint.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/timepoint.2), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/timepoint.1..pr/timepoint.2)) switching to std::cmp functions for more reliable static asserts

maflcko:
Looks like iwyu fails?

ryanofsky (commented):
Updated 978028ffd0d14e5e46b313412ef17b7ad3647e33 -> 1e3311158138ac9d458656805e666663f8f19f40 ([`pr/timepoint.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/timepoint.2) -> [`pr/timepoint.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/timepoint.3), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/timepoint.2..pr/timepoint.3)) to fix IWYU error https://github.com/bitcoin-core/libmultiprocess/actions/runs/30503532739/job/90748227074

Rebased 1e3311158138ac9d458656805e666663f8f19f40 -> 7a72df02e2d46f953c4c97662c3e109aa3874ea9 ([`pr/timepoint.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/timepoint.3) -> [`pr/timepoint.4`](https://github.com/ryanofsky/libmultiprocess/commits/pr/timepoint.4), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/timepoint.3-rebase..pr/timepoint.4)) adding new commit and rebasing to fix incompatibility with `bool` and `std::cmp_less_equal/std::cmp_greater_equal` in libc++  https://github.com/bitcoin-core/libmultiprocess/actions/runs/30551079943/job/90899596238?pr=303

include/mp/util.h
```diff
@@ -33,6 +33,21 @@ namespace mp {
...
+constexpr bool safe_less_equal(A a, B b)
+{
+    if constexpr (std::is_floating_point_v<A> || std::is_floating_point_v<B>) return a <= b;
+    else return std::cmp_less_equal(a, b);
+}
+template <typename A, typename B>
+constexpr bool safe_greater_equal(A a, B b)
+{
+    if constexpr (std::is_floating_point_v<A> || std::is_floating_point_v<B>) return a >= b;
+    else return std::cmp_greater_equal(a, b);
```
uqlidi, line 48:
```suggestion
    else return safe_less_equal(b, a);
```

uqlidi (changes requested):
looks good but there's a small suggestion

maflcko:
lgtm ACK 7a72df02e2d46f953c4c97662c3e109aa3874ea9

ViniciusCestarii (commented):
ACK 7a72df02e2d46f953c4c97662c3e109aa3874ea9

include/mp/util.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/303#discussion_r3688595630

This does seem like a good idea, but it's a minor cleanup so not planning to make the change here. Could be worth doing if this code changes again.

PR #304, opened 2026-07-08
proxy: fix BuildList to use non-const iteration for interface types

ryanofsky:
Needed for bitcoin/bitcoin#10102 since #277 was merged.

Since #277, it is no longer possible to return `vector<unique_ptr<ExternalSigner>>` from `Node::listExternalSigners()` in bitcoin/bitcoin#10102, because the proxy server objects libmultiprocess creates to wrap the `ExternalSigner` objects need to take ownership of the objects to keep them alive, so they need to be moved out of the returned vector, which can only be done if `BuildField` is passed a non-const vector element.

This PR restores previous behavior before #277 making it possible to mutate container elements while serializing them, in cases like this where it is necessary.

### DIFF
--- a/include/mp/proxy-types.h
+++ b/include/mp/proxy-types.h
@@ -285,9 +285,28 @@ void BuildList(TypeList<LocalType>, InvokeContext& invoke_context, Output&& outp
 {
     auto list = output.init(value.size());
     size_t i = 0;
-    for (const auto& elem : value) {
-        BuildField(TypeList<LocalType>(), invoke_context, ListOutput<typename decltype(list)::Builds>(list, i), elem);
-        ++i;
+    // Iterate with an explicit iterator rather than a range-for loop so the
+    // value category of `*it` is passed through to BuildField unchanged. This
+    // matters for two reasons:
+    //
+    // - Elements must be passed as non-const so that BuildField can move out of
+    //   them, e.g. calling unique_ptr::release() to transfer ownership of an
+    //   interface pointer to the capnp server.
+    //
+    // - For proxy containers like std::vector<bool>, `*it` is a prvalue proxy
+    //   object rather than a reference. A range-for loop would bind it to a
+    //   named variable and demote it to an lvalue; passing `*it` directly
+    //   preserves the prvalue-ness.
+    //
+    // Only move out of elements when `value` itself is an rvalue container that
+    // is about to be destroyed. When it is an lvalue reference owned by the
+    // caller, pass elements as lvalues so BuildField does not move from them.
+    for (auto it = value.begin(); it != value.end(); ++it, ++i) {
+        if constexpr (std::is_lvalue_reference_v<Value&&>) {
+            BuildField(TypeList<LocalType>(), invoke_context, ListOutput<typename decltype(list)::Builds>(list, i), *it);
+        } else {
+            BuildField(TypeList<LocalType>(), invoke_context, ListOutput<typename decltype(list)::Builds>(list, i), std::move(*it));
+        }
     }
 }
 
--- a/test/mp/test/test.cpp
+++ b/test/mp/test/test.cpp
@@ -272,6 +272,16 @@ KJ_TEST("Call FooInterface methods")
     KJ_REQUIRE(data_out[0] != nullptr);
     KJ_EXPECT(*data_out[0] == *data_in[0]);
     KJ_EXPECT(!data_out[1]);
+
+    // Test returning vector<unique_ptr<interface>> from server. This exercises
+    // BuildList with interface element types, which requires non-const iteration
+    // so unique_ptr::release() can transfer ownership to the proxy server.
+    std::vector<std::unique_ptr<Bar>> bars{foo->listBars(3)};
+    KJ_REQUIRE(bars.size() == 3u);
+    for (int i = 0; i < 3; ++i) {
+        KJ_REQUIRE(bars[i] != nullptr);
+        KJ_EXPECT(bars[i]->value() == i);
+    }
 }
 
 KJ_TEST("Call IPC method after client connection is closed")
--- a/test/mp/test/test.cpp
+++ b/test/mp/test/test.cpp
@@ -269,6 +269,16 @@ KJ_TEST("Call FooInterface methods")
     KJ_REQUIRE(data_out[0] != nullptr);
     KJ_EXPECT(*data_out[0] == *data_in[0]);
     KJ_EXPECT(!data_out[1]);
+
+    // Test returning vector<unique_ptr<interface>> from server. This exercises
+    // BuildList with interface element types, which requires non-const iteration
+    // so unique_ptr::release() can transfer ownership to the proxy server.
+    std::vector<std::unique_ptr<Bar>> bars{foo->listBars(3)};
+    KJ_REQUIRE(bars.size() == 3u);
+    for (int i = 0; i < 3; ++i) {
+        KJ_REQUIRE(bars[i] != nullptr);
+        KJ_EXPECT(bars[i]->value() == i);
+    }
 }
 
 KJ_TEST("Call IPC method after client connection is closed")
--- a/include/mp/type-map.h
+++ b/include/mp/type-map.h
@@ -9,6 +9,8 @@
 #include <mp/type-pair.h>
 #include <mp/util.h>
 
+#include <map>
+
 namespace mp {
 template <typename KeyLocalType, typename ValueLocalType, typename Value, typename Output>
 void CustomBuildField(TypeList<std::map<KeyLocalType, ValueLocalType>>,
--- a/src/mp/gen.cpp
+++ b/src/mp/gen.cpp
@@ -329,6 +329,7 @@ static void Generate(kj::StringPtr src_prefix,
     cpp_client << "#include <kj/common.h>\n";
     cpp_client << "#include <map>\n";
     cpp_client << "#include <mp/proxy.h>\n";
+    cpp_client << "#include <mp/proxy-io.h>\n";
     cpp_client << "#include <mp/util.h>\n";
     cpp_client << "#include <string>\n";
     cpp_client << "#include <vector>\n";
--- a/test/mp/test/foo-types.h
+++ b/test/mp/test/foo-types.h
@@ -39,6 +39,7 @@ struct ExtendedCallback; // IWYU pragma: export
 struct FooCallback; // IWYU pragma: export
 struct FooFn; // IWYU pragma: export
 struct FooInterface; // IWYU pragma: export
+struct BarInterface; // IWYU pragma: export
 } // namespace messages
 
 template <typename Output>
--- a/test/mp/test/foo.capnp
+++ b/test/mp/test/foo.capnp
@@ -37,6 +37,7 @@ interface FooInterface $Proxy.wrap("mp::test::FooImplementation") {
     callFnAsync @18 (context :Proxy.Context) -> ();
     callIntFnAsync @21 (context :Proxy.Context, arg :Int32) -> (result :Int32);
     passDataPointers @22 (arg :List(Data)) -> (result :List(Data));
+    listBars @25 (context :Proxy.Context, n :Int32) -> (result :List(BarInterface));
 }
 
 interface FooInit $Proxy.wrap("mp::test::FooInit") {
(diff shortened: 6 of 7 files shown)


ryanofsky:
Updated a0b01d73698597293e48771cad2b4cb8e0ba16a0 -> 46deb5e6bcb853c29f7573ce5e19414929f7f551 ([`pr/buildlist.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/buildlist.2) -> [`pr/buildlist.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/buildlist.3), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/buildlist.2..pr/buildlist.3)) to fix IWYU CI failures: add `#include <mp/proxy-io.h>` to generated proxy-client files and `reserve()` call in `listCallbacks` https://github.com/bitcoin-core/libmultiprocess/actions/runs/28966350534/job/85950397301

test/mp/test/test.cpp
```diff
@@ -266,6 +258,16 @@ KJ_TEST("Call FooInterface methods")
...
     KJ_EXPECT(!data_out[1]);
+
+    // Test returning vector<unique_ptr<interface>> from server. This exercises
+    // BuildList with interface element types, which requires non-const iteration
+    // so unique_ptr::release() can transfer ownership to the proxy server.
+    std::vector<std::unique_ptr<FooCallback>> callbacks{foo->listCallbacks(3)};
+    KJ_REQUIRE(callbacks.size() == 3u);
+    for (int i = 0; i < 3; ++i) {
+        KJ_REQUIRE(callbacks[i] != nullptr);
+        KJ_EXPECT(callbacks[i]->call(0) == i);
```
Sjors: IIUC this isn't really a callback flow, so the terminology had me confused about the test. Might be better to add another (`Bar`) interface to make it more clear.

```diff
diff --git a/test/mp/test/foo-types.h b/test/mp/test/foo-types.h
index b96eabfcfb..96506a9f50 100644
--- a/test/mp/test/foo-types.h
+++ b/test/mp/test/foo-types.h
@@ -40,4 +40,5 @@ struct FooCallback; // IWYU pragma: export
 struct FooFn; // IWYU pragma: export
 struct FooInterface; // IWYU pragma: export
+struct BarInterface; // IWYU pragma: export
 } // namespace messages

diff --git a/test/mp/test/foo.capnp b/test/mp/test/foo.capnp
index edf7841c8a..0896274209 100644
--- a/test/mp/test/foo.capnp
+++ b/test/mp/test/foo.capnp
@@ -37,5 +37,5 @@ interface FooInterface $Proxy.wrap("mp::test::FooImplementation") {
     callIntFnAsync @21 (context :Proxy.Context, arg :Int32) -> (result :Int32);
     passDataPointers @22 (arg :List(Data)) -> (result :List(Data));
-    listCallbacks @24 (context :Proxy.Context, n :Int32) -> (result :List(FooCallback));
+    listBars @24 (context :Proxy.Context, n :Int32) -> (result :List(BarInterface));
 }

@@ -49,4 +49,9 @@ interface ExtendedCallback extends(FooCallback) $Proxy.wrap("mp::test::ExtendedC
 }

+interface BarInterface $Proxy.wrap("mp::test::Bar") {
+    destroy @0 (context :Proxy.Context) -> ();
+    value @1 (context :Proxy.Context) -> (result :Int32);
+}
+
 interface FooFn $Proxy.wrap("ProxyCallback<std::function<int()>>") {
     destroy @0 (context :Proxy.Context) -> ();
diff --git a/test/mp/test/foo.h b/test/mp/test/foo.h
index 17aeeb8388..5c881117ce 100644
--- a/test/mp/test/foo.h
+++ b/test/mp/test/foo.h
@@ -60,13 +60,4 @@ public:
 };

-//! Concrete FooCallback that returns a fixed value, used by listCallbacks tests.
-class SimpleCallback : public FooCallback
-{
-public:
-    explicit SimpleCallback(int value) : m_value(value) {}
-    int call(int) override { return m_value; }
-    int m_value;
-};
-
 class ExtendedCallback : public FooCallback
 {
@@ -75,4 +66,27 @@ public:
 };

+//! A second, arbitrary interface used to test returning a
+//! list of interface objects.
+class Bar
+{
+public:
+    virtual ~Bar() = default;
+    virtual int value() = 0;
+};
+
+//! Concrete Bar that returns a fixed value, used by listBars tests.
+class SimpleBar : public Bar
+{
+public:
+    explicit SimpleBar(int value) : m_value(value) {}
+    int value() override { return m_value; }
+    int m_value;
+};
+
 class FooImplementation
 {
@@ -99,9 +113,9 @@ public:
     int passFn(std::function<int()> fn) { return fn(); }
     std::vector<FooDataRef> passDataPointers(std::vector<FooDataRef> values) { return values; }
-    std::vector<std::unique_ptr<FooCallback>> listCallbacks(int n)
+    std::vector<std::unique_ptr<Bar>> listBars(int n)
     {
-        std::vector<std::unique_ptr<FooCallback>> result;
+        std::vector<std::unique_ptr<Bar>> result;
         result.reserve(n);
-        for (int i = 0; i < n; ++i) result.push_back(std::make_unique<SimpleCallback>(i));
+        for (int i = 0; i < n; ++i) result.push_back(std::make_unique<SimpleBar>(i));
         return result;
     }
diff --git a/test/mp/test/test.cpp b/test/mp/test/test.cpp
index 41ca506c62..41aa4d67fd 100644
--- a/test/mp/test/test.cpp
+++ b/test/mp/test/test.cpp
@@ -263,9 +263,9 @@ KJ_TEST("Call FooInterface methods")
     // BuildList with interface element types, which requires non-const iteration
     // so unique_ptr::release() can transfer ownership to the proxy server.
-    std::vector<std::unique_ptr<FooCallback>> callbacks{foo->listCallbacks(3)};
-    KJ_REQUIRE(callbacks.size() == 3u);
+    std::vector<std::unique_ptr<Bar>> bars{foo->listBars(3)};
+    KJ_REQUIRE(bars.size() == 3u);
     for (int i = 0; i < 3; ++i) {
-        KJ_REQUIRE(callbacks[i] != nullptr);
-        KJ_EXPECT(callbacks[i]->call(0) == i);
+        KJ_REQUIRE(bars[i] != nullptr);
+        KJ_EXPECT(bars[i]->value() == i);
     }
 }
```

include/mp/proxy-types.h
```diff
@@ -285,7 +285,7 @@ void BuildList(TypeList<LocalType>, InvokeContext& invoke_context, Output&& outp
 {
     auto list = output.init(value.size());
     size_t i = 0;
-    for (const auto& elem : value) {
+    for (auto&& elem : value) {
```
ryanofsky: In commit "proxy: fix BuildList to use non-const iteration for interface types" (a0b01d73698597293e48771cad2b4cb8e0ba16a0)

Note, while this change mostly restores behavior to what it was before #277, passing `elem` to `BuildField` below as an lvalue reference regardless of whether the container is lvalue or rvalue, it might make more sense to pass `elem` as an rvalue if the container is an rvalue, or as an lvalue if it's an lvalue with something like:

```c++
for (auto&& elem : value) {
    if constexpr (std::is_lvalue_reference_v<Value&&>) {
        BuildField(/*...*/, elem);
    } else {
        BuildField(/*...*/, std::move(elem));
    }
}
```

This could be safer because it could allow `BuildField` to distinguish based on the way the container is being used and only move from temporary containers that are about to be destroyed, not containers passed as references.

Also it might be better to replace `for (auto&& elem : value)` with `for (auto it = elem.begin(); it != elem.end(); ++it) like the previous code had since this would could also preserve rvaluedness of dereferencing the iterator, if the iterator deferences as a proxy object instead of a reference like `vector<bool>`

Which is just to say that a0b01d73698597293e48771cad2b4cb8e0ba16a0 is a minimal change restoring ability to return lists of interface pointers, but it might make sense to actually increase safety here as well.

test/mp/test/test.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/304#discussion_r3551766092

> IIUC this isn't really a callback flow, so the terminology had me confused about the test. Might be better to add another (`Bar`) interface to make it more clear.

Yeah you are right. The server is returning callback objects in the sense of "you can call me back later by using these objects" not taking callback objects that it can use to call into the client later. So your example makes more sense and its better to avoid the callback term here.

ryanofsky (commented):
Thanks for review. Looks like there are some things that should be improved here so I will make this a draft

test/mp/test/test.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/304#discussion_r3551766092

Applied suggested changes. I tried to do something a simpler adding a FooInterface method that returns a list of FooInterfaces, but it didn't work out because there's no C++ FooInterface class with virtual methods. It might make sense to add virtual methods and have separate FooInterface / FooImplementation classes which would make the test a little more realistic. But it would also require defining each method twice and overriding it, so seem better to keep current test setup.

ryanofsky (commented):
Updated 46deb5e6bcb853c29f7573ce5e19414929f7f551 -> b14b0bfaeaeffc9115af0bfc12490b233b8bb33a ([`pr/buildlist.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/buildlist.3) -> [`pr/buildlist.4`](https://github.com/ryanofsky/libmultiprocess/commits/pr/buildlist.4), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/buildlist.3..pr/buildlist.4)) updating test as suggested

Rebased b14b0bfaeaeffc9115af0bfc12490b233b8bb33a -> 5b13fe52d6a9ce1b47b20e5fafbcfe0c77db2d86 ([`pr/buildlist.4`](https://github.com/ryanofsky/libmultiprocess/commits/pr/buildlist.4) -> [`pr/buildlist.5`](https://github.com/ryanofsky/libmultiprocess/commits/pr/buildlist.5), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/buildlist.4-rebase..pr/buildlist.5)) due to silent conflict with #305 and adding more rvalueness preserving code for more safety (to avoid moving from lvalues)

Updated 5b13fe52d6a9ce1b47b20e5fafbcfe0c77db2d86 -> f9e1c8d6b518888dceb2f1b4f609f3004ff73554 ([`pr/buildlist.5`](https://github.com/ryanofsky/libmultiprocess/commits/pr/buildlist.5) -> [`pr/buildlist.6`](https://github.com/ryanofsky/libmultiprocess/commits/pr/buildlist.6), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/buildlist.5..pr/buildlist.6)) to fix IWYU error https://github.com/bitcoin-core/libmultiprocess/actions/runs/30508160651/job/90762305561?pr=304

include/mp/proxy-types.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/304#discussion_r3552087096

Latest push implements all these suggestions so rvaluedness is preserved and the code should be safer and potentially work more types. It is a little more verbose and complicated now, but I think not too bad.

include/mp/proxy-types.h
```diff
@@ -285,9 +285,28 @@ void BuildList(TypeList<LocalType>, InvokeContext& invoke_context, Output&& outp
...
+    // - For proxy containers like std::vector<bool>, `*it` is a prvalue proxy
+    //   object rather than a reference. A range-for loop would bind it to a
+    //   named variable and demote it to an lvalue; passing `*it` directly
+    //   preserves the prvalue-ness.
+    //
+    // Only move out of elements when `value` itself is an rvalue container that
+    // is about to be destroyed. When it is an lvalue reference owned by the
+    // caller, pass elements as lvalues so BuildField does not move from them.
+    for (auto it = value.begin(); it != value.end(); ++it, ++i) {
+        if constexpr (std::is_lvalue_reference_v<Value&&>) {
```
ViniciusCestarii, line 305: Currently, all callers of BuildList pass the named parameter value, which is an lvalue. Should those callers use std::forward<Value>(value) so the move path actually kicks in? Otherwise the comment about only moving out of rvalue containers describes something that can't happen with the current callers.

ViniciusCestarii (commented):
tACK d6f8588d1ab8195c72a24359d0e67f9f6c0f11b2 

Checked that the new test prevents the regression and left a non blocking question.

include/mp/proxy-types.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/304#discussion_r3787138852

Yes they should. I implemented this already in b6951636acb1c6b0bd655dfc2264f9d86f2345ff, in a [branch](https://github.com/ryanofsky/libmultiprocess/commits/pr/strictmove) I didn't push yet. I didn't include those changes in this PR because this is needed to fix the https://github.com/bitcoin/bitcoin/pull/10102 build after #277 and I didn't want to make it too complicated

PR #306, opened 2026-07-10
doc: Bump version 12 > 13

ryanofsky:
Increase version number after previous version was added to bitcoin core. Also update version history and documentation.


Sjors:
ACK 438fdd243d3987dd99d843b472a693c6d3c2923b

I only read the v12 release note, not the misc Bitcoin Core version refinements in older releases.

PR #305, opened 2026-07-09
refactor: memcpy to std::ranges::copy to work around ubsan warn

maflcko:
Using std::ranges::copy from the C++ standard library has a few benefits here:

* It has the additional benefit of being a bit more type safe and document any needed cast explicitly. (No cast is needed here)
* It has defined semantics for empty source ranges.

This should also fix https://github.com/bitcoin/bitcoin/pull/35118#discussion_r3506566815

PR #307, opened 2026-07-14
 refactor: memcpy -> std::ranges::copy 

maflcko:
Some follow-ups to https://github.com/bitcoin-core/libmultiprocess/pull/305#pullrequestreview-4689778284

### DIFF
--- a/include/mp/type-char.h
+++ b/include/mp/type-char.h
@@ -7,6 +7,10 @@
 
 #include <mp/util.h>
 
+#include <algorithm>
+#include <ranges>
+#include <stdexcept>
+
 namespace mp {
 template <typename Output, size_t size>
 void CustomBuildField(TypeList<const unsigned char*>,
@@ -16,7 +20,7 @@ void CustomBuildField(TypeList<const unsigned char*>,
     Output&& output)
 {
     auto result = output.init(size);
-    memcpy(result.begin(), value, size);
+    std::ranges::copy(value, result.begin());
 }
 
 template <size_t size, typename Input, typename ReadDest>
@@ -26,9 +30,11 @@ decltype(auto) CustomReadField(TypeList<unsigned char[size]>,
     Input&& input,
     ReadDest&& read_dest)
 {
+    auto data = input.get();
+    if (data.size() != size) throw std::range_error("unexpected field size");
+
     return read_dest.update([&](auto& value) {
-        auto data = input.get();
-        memcpy(value, data.begin(), size);
+        std::ranges::copy(input.get(), std::ranges::begin(value));
     });
 }
 } // namespace mp
--- a/include/mp/type-data.h
+++ b/include/mp/type-data.h
@@ -32,9 +32,8 @@ template <typename LocalType, typename Value, typename Output>
 void CustomBuildField(TypeList<LocalType>, Priority<2>, InvokeContext& invoke_context, Value&& value, Output&& output)
 requires (std::is_same_v<decltype(output.get()), ::capnp::Data::Builder> && IsByteSpan<LocalType>)
 {
-    auto data = std::span{value};
-    auto result = output.init(data.size());
-    std::ranges::copy(data, result.begin());
+    auto result = output.init(value.size());
+    std::ranges::copy(value, result.begin());
 }
 
 template <typename LocalType, typename Input, typename ReadDest>
--- a/include/mp/type-string.h
+++ b/include/mp/type-string.h
@@ -7,6 +7,8 @@
 
 #include <mp/util.h>
 
+#include <algorithm>
+
 namespace mp {
 template <typename Value, typename Output>
 void CustomBuildField(TypeList<std::string>,
@@ -16,7 +18,7 @@ void CustomBuildField(TypeList<std::string>,
     Output&& output)
 {
     auto result = output.init(value.size());
-    memcpy(result.begin(), value.data(), value.size());
+    std::ranges::copy(value, result.begin());
 }
 
 template <typename Input, typename ReadDest>
--- a/test/mp/test/listen_tests.cpp
+++ b/test/mp/test/listen_tests.cpp
@@ -225,7 +225,7 @@ KJ_TEST("ListenConnections enforces a local connection limit")
 {
     // With max-connections=1, the second socket can connect to the kernel
     // backlog, but ListenConnections should not accept or serve it until the
-    // first accepts clients disconnects.
+    // first accepted client disconnects.
 
     ListenSetup server(/*max_connections=*/1);
 


include/mp/type-string.h
```diff
@@ -7,6 +7,9 @@
 
 #include <mp/util.h>
 
+#include <algorithm>
+#include <ranges>
```
ViniciusCestarii: In "refactor: memcpy -> std::ranges::copy" 8226c05549b0405ebac3208751c53a03563ea745

nit: I believe this `#include <ranges>` is not necessary here

ViniciusCestarii (commented):
ACK 8226c05549b0405ebac3208751c53a03563ea745 just commented a nit

include/mp/type-string.h
maflcko: Isn't the iwyu CI supposed to catch this? It passes with either version ... 

Maybe it could make sense to fix the CI instead, so that all places are fixed and not only the ones caught in review?

ViniciusCestarii: The CI runs IWYU and it only audits ["the input .cc file and its associated .h files"](https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_globals.cc#L78) (e.g. foo.cpp also checks foo.h). type-string.h has no matching .cpp so it isn't audited.

```diff
-  set(CMAKE_CXX_INCLUDE_WHAT_YOU_USE "${IWYU_EXECUTABLE};-Xiwyu;--error")
+  set(CMAKE_CXX_INCLUDE_WHAT_YOU_USE "${IWYU_EXECUTABLE};-Xiwyu;--error;-Xiwyu;--check_also=${PROJECT_SOURCE_DIR}/include/mp/*.h")
```

With the above diff it checks the headers too and can check for type-string.h. ([Bitcoin core also uses --check_also for primitive headers](https://github.com/bitcoin/bitcoin/blob/master/ci/test/03_test_script.sh#L250)):
```
/home/vinicius/Code/my/libmultiprocess/include/mp/type-string.h should add these lines:
#include <string>     // for string
namespace mp { struct InvokeContext; }

/home/vinicius/Code/my/libmultiprocess/include/mp/type-string.h should remove these lines:
- #include <ranges>  // lines 11-11
```
It also flag a lot of what there is currently (10+ header files) and can also make false positives: note the `#include <variant>                     // for tuple` which doens't make sense for proxy-io.h:
```
/home/vinicius/Code/my/libmultiprocess/include/mp/proxy-io.h should add these lines:
#include <capnp/capability.h>          // for Capability, CallContext, Capab...
#include <capnp/common.h>              // for Void, word
#include <capnp/message.h>             // for MallocMessageBuilder, ReaderOp...
#include <capnp/rpc-twoparty.capnp.h>  // for VatId, Side, Side_9fd69ebc87b9...
#include <capnp/rpc.h>                 // for RpcSystem, makeRpcClient, make...
#include <kj/async-io.h>               // for LowLevelAsyncIoProvider, Async...
#include <kj/async-prelude.h>          // for ReadyNow
#include <kj/async.h>                  // for TaskSet, Promise, READY_NOW
#include <kj/common.h>                 // for mv, ArrayPtr, KJ_IF_MAYBE, Maybe
#include <kj/exception.h>              // for runCatchingExceptions, Exception
#include <kj/memory.h>                 // for Own, heap
#include <kj/string.h>                 // for KJ_STRINGIFY, StringPtr
#include <list>                        // for _List_iterator, list, _List_co...
#include <tuple>                       // for tuple
#include <utility>                     // for forward, move
#include <variant>                     // for tuple
#include <vector>                      // for vector
namespace mp { class Connection; }
namespace mp { class EventLoop; }
namespace mp { template <typename Interface> struct ProxyClient; }
namespace mp { template <typename Interface> struct ProxyServer; }
```

So I believe this isn't a one line fix, it would require a one-time PR to manually audit and fix and map the false positives in a mapping file and then enabling --check_also and --mapping_file with the mapping file to handle false positives on the CI. Maybe this is not worth effort.

This is just what I found poking around, if there's a better way id like to hear it.

include/mp/type-data.h
```diff
@@ -32,9 +32,8 @@ template <typename LocalType, typename Value, typename Output>
 void CustomBuildField(TypeList<LocalType>, Priority<2>, InvokeContext& invoke_context, Value&& value, Output&& output)
 requires (std::is_same_v<decltype(output.get()), ::capnp::Data::Builder> && IsByteSpan<LocalType>)
 {
-    auto data = std::span{value};
-    auto result = output.init(data.size());
-    std::ranges::copy(data, result.begin());
+    auto result = output.init(value.size());
```
ryanofsky, line 35: In commit "refactor: Directly use value in CustomBuildField" (ce865a9ba86e603e6eb69aba53f791203b80fcb2)

This looks like a it's a good change, but it's not really implementing the suggestion to drop the `result` variable. It's doing something different and dropping the `data` variable (which is good and I didn't know was possible). Maybe consider updating the commit message or dropping `result` too

include/mp/type-char.h
```diff
@@ -27,8 +30,7 @@ decltype(auto) CustomReadField(TypeList<unsigned char[size]>,
     ReadDest&& read_dest)
 {
     return read_dest.update([&](auto& value) {
-        auto data = input.get();
-        memcpy(value, data.begin(), size);
+        std::ranges::copy(input.get(), std::ranges::begin(value));
```
ryanofsky, line 37: In commit "refactor: memcpy -> std::ranges::copy" (8226c05549b0405ebac3208751c53a03563ea745)

This is ok, but it is now deciding how much data to copy based on the length of the input instead of the length of the output, so previously if there was a mismatch the code could read too many bytes, and now it could write too many bytes.

More ideally this would ensure the size matches with something like:

```c++
auto data = input.get();
if (data.size() != size) throw std::range_error("unexpected field size");
```

include/mp/type-string.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/307#discussion_r3605683715

Wow the horrors of IWYU are unending! Would review patch, but this adds to my regrets from https://github.com/bitcoin/bitcoin/pull/10575#issuecomment-307782757

ryanofsky (approved):
Code review ACK 8226c05549b0405ebac3208751c53a03563ea745. Thanks for the followup. Left some suggestions but also would be happy if the PR were merged as-is

include/mp/type-char.h
maflcko: thx, done

include/mp/type-data.h
maflcko: This is mostly for consistency with string `include/mp/type-string.h`, which also has the allocation in a separate line:

```cpp
    auto result = output.init(value.size());

include/mp/type-string.h
maflcko: thx, removed `#include <ranges>` for now

ryanofsky (approved):
Code review ACK 26452e02d75e8a2a33752ef181f7ffc96ca968f0. Thanks for the updates!

PR #310, opened 2026-07-17
test: cover immediate client disconnects for `ListenConnections`

xyzconstant:
Following on testing the reversed direction *ryanofsky* [suggested](https://github.com/bitcoin-core/libmultiprocess/pull/298#issuecomment-4887471523) in #298, this PR adds a test to cover immediate client disconnects on the server side.

Additionally, this test adds `DefaultLogHandler` and a `log_handler` parameter to `ListenSetup`, allowing individual tests to observe logs by passing a custom log handler. The new test takes advantage of this by catching and skipping `Uncaught exception in daemonized task.` logs.

NOTE: an issue surfaced on the macOS job, the `accept()` call in `ListenConnections` [fails](https://github.com/bitcoin-core/libmultiprocess/actions/runs/29545518652/job/87776862017?pr=310) for the closed connection, this is a Cap'n Proto bug as reported by *ViniciusCestarii* in https://github.com/bitcoin-core/libmultiprocess/pull/310#pullrequestreview-4723481031 and its fix is [available](https://github.com/capnproto/capnproto/commit/7df5bd078f389ded313479981bd0ae06cbcdfe1b#diff-ec577ad66535f58f6d7396ea51d3e56c0065308aa8fb02751cd6a8cfaa67252fR1358-R1372) in the v2 branch.

### DIFF
--- a/test/mp/test/listen_tests.cpp
+++ b/test/mp/test/listen_tests.cpp
@@ -129,18 +129,23 @@ public:
     std::thread thread;
 };
 
+//! Default server event loop log handler, throws so tests can assert on errors.
+void DefaultLogHandler(mp::LogMessage log)
+{
+    KJ_LOG(INFO, log.level, log.message);
+    if (log.level == mp::Log::Raise) throw std::runtime_error(log.message);
+}
+
 //! Runs a server EventLoop on its own thread, starts ListenConnections() on a
 //! UnixListener socket, and records connection/disconnection counts through
 //! EventLoop test hooks
 class ListenSetup
 {
 public:
-    explicit ListenSetup(std::optional<size_t> max_connections = std::nullopt)
-        : thread([this, max_connections] {
-              EventLoop loop("mptest-server", [](mp::LogMessage log) {
-                  KJ_LOG(INFO, log.level, log.message);
-                  if (log.level == mp::Log::Raise) throw std::runtime_error(log.message);
-              });
+    explicit ListenSetup(std::optional<size_t> max_connections = std::nullopt,
+                         mp::LogFn log_handler = DefaultLogHandler)
+        : thread([this, max_connections, log_handler] {
+              EventLoop loop("mptest-server", log_handler);
               loop.testing_hook_disconnected = [&] {
                   Lock lock(counter_mutex);
                   ++disconnected_count;
--- a/test/mp/test/listen_tests.cpp
+++ b/test/mp/test/listen_tests.cpp
@@ -8,6 +8,7 @@
 #include <mp/test/foo.capnp.proxy.h>
 
 #include <chrono>
+#include <compare>
 #include <condition_variable>
 #include <cstdlib>
 #include <cstring>
@@ -292,6 +293,43 @@ KJ_TEST("ListenConnections accepts multiple connections")
     server.WaitForConnectedCount(3);
 
     KJ_EXPECT(client3->client->add(3, 4) == 7);
+}
+
+KJ_TEST("ListenConnections handles a client that disconnects before being accepted")
+{
+    Mutex mutex;
+    bool accept_error = false;
+    ListenSetup server(std::nullopt, [&](mp::LogMessage log) {
+        // On macOS, accept() can fail if the peer closes the connection before it is accepted.
+        // The event loop then reports this as an uncaught task exception. We catch and ignore
+        // this specific error here so that the corresponding CI job does not fail.
+        //
+        // This is a Cap'n Proto bug, a fix is available in the v2 branch at: https://github.com/capnproto/capnproto/commit/7df5bd078f389ded313479981bd0ae06cbcdfe1b#diff-ec577ad66535f58f6d7396ea51d3e56c0065308aa8fb02751cd6a8cfaa67252fR1358-R1372
+        if (log.level == mp::Log::Error && log.message.find("Uncaught exception in daemonized task.") != std::string::npos) {
+            Lock lock(mutex);
+            accept_error = true;
+        }
+        DefaultLogHandler(log);
+    });
+
+    // This is racy, if the close does not happen before accept(),
+    // the connection is accepted normally.
+    int fd = server.listener.MakeConnectedSocket();
+    KJ_SYSCALL(close(fd));
+
+    // Wait for the connection to either be accepted and disconnected, or fail
+    // to be accepted and log the error above.
+    const auto deadline = std::chrono::steady_clock::now() + FAILURE_TIMEOUT;
+    auto accept_error_seen = [&] { Lock lock(mutex); return accept_error; };
+    while (!accept_error_seen() && server.DisconnectedCount() < 1 && std::chrono::steady_clock::now() < deadline) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+    }
+
+#ifndef __APPLE__
+    // No error is expected on platforms other than macOS.
+    KJ_REQUIRE(!accept_error_seen());
+#endif
+    KJ_EXPECT(server.DisconnectedCount() == (accept_error_seen() ? 0 : 1));
 }
 
 } // namespace


xyzconstant:
the macos job [failed](https://github.com/bitcoin-core/libmultiprocess/actions/runs/29545518652/job/87776862017?pr=310) with:

```
[ TEST ] listen_tests.cpp:292: ListenConnections survives a client that disconnects before being accepted
mp/proxy.cpp:48: error: Uncaught exception in daemonized task.; exception = kj/async-io-unix.c++:1365: failed: setsocketopt(IPPROTO_TCP, TCP_NODELAY): Invalid argument
stack: 1020dc7cb 1020dcaff 1020dcc60 10079556c
/Users/runner/work/libmultiprocess/libmultiprocess/test/mp/test/listen_tests.cpp:189: failed: expected matched
stack: 10078b28b 1007ac7bf
```

I came across the same issue on my machine (MacBook Pro M2) too. I'll push a fix in a moment.

include/mp/proxy-io.h
```diff
@@ -902,6 +902,12 @@ void _Listen(const std::shared_ptr<Listener>& listener, EventLoop& loop, InitImp
                 if (resume_accept) _Listen<InitInterface>(listener, loop, init);
             });
             _Listen<InitInterface>(listener, loop, init);
+        },
+        // Keep listening if a single accept() fails, so the server does not
+        // stop accepting future connections.
+        [&loop, &init, listener](const kj::Exception& e) {
+            MP_LOG(loop, Log::Info) << "IPC server: accept failed.";
```
ViniciusCestarii: In "Keep listening after a failed accept in `ListenConnections`" 38391ebc196d7ece7152a6372993ba104dceb9c6

nit: I believe `MP_LOG(loop, Log::Warning)` would fit better here and also the kj exception could be logged too

ViniciusCestarii (commented):
ACK 38391ebc196d7ece7152a6372993ba104dceb9c6 Nice finding! ~~This is nicer than~~ the old code that would just stop listening because of a transient error that KJ didn't catch but also with this change now a non-transient error will spin forever retrying an unrecoverable error which I believe could be fixed in a follow up.

Also I have dug in and I found that this is a KJ bug because it should treat transient errors like this and [it was already fixed for capnp v2](https://github.com/capnproto/capnproto/commit/7df5bd078f389ded313479981bd0ae06cbcdfe1b#diff-ec577ad66535f58f6d7396ea51d3e56c0065308aa8fb02751cd6a8cfaa67252fR1358-R1372) which isn't released yet.

~~Given that, I still think this PR is correct and the defensive retry is the right call regardless. The spin forever on permanent failure case remains a valid follow-up on top of this.~~

test/mp/test/listen_tests.cpp
```diff
@@ -289,6 +289,21 @@ KJ_TEST("ListenConnections accepts multiple connections")
...
 }
 
+KJ_TEST("ListenConnections survives a client that disconnects before being accepted")
+{
+    ListenSetup server;
+
+    // Connect and close before the server has a chance to accept() the
+    // connection.
+    int fd = server.listener.MakeConnectedSocket();
+    close(fd);
```
ryanofsky: In commit "Keep listening after a failed accept in `ListenConnections`" (38391ebc196d7ece7152a6372993ba104dceb9c6)

Can this be checked with KJ_SYSCALL?

xyzconstant:
Thanks for the review @ViniciusCestarii!

> Also I have dug in and I found that this is a KJ bug because it should treat transient errors like this and [it was already fixed for capnp v2](https://github.com/capnproto/capnproto/commit/7df5bd078f389ded313479981bd0ae06cbcdfe1b#diff-ec577ad66535f58f6d7396ea51d3e56c0065308aa8fb02751cd6a8cfaa67252fR1358-R1372) which isn't released yet.

Nice catch! Now it's clear why this wasn't failing on other platforms.

> Given that, I still think this PR is correct and the defensive retry is the right call regardless. The spin forever on permanent failure case remains a valid follow-up on top of this.

You're right on this, I was considering this too. I'm yet to figure out which exceptions we want to stop the accept loop for and which we want to retry, so a follow-up is the right call to me. If other reviewers would rather it be solved in this PR, I'm happy to tackle it here though.

include/mp/proxy-io.h
xyzconstant: Solved at d8e5395ed8e6f21fc847b65148c2ec3c34bce524 [(compare)](https://github.com/bitcoin-core/libmultiprocess/compare/92d8f86c13259cf5663adf4ffe52f187c74bf1de..d8e5395ed8e6f21fc847b65148c2ec3c34bce524)

Thanks!

test/mp/test/listen_tests.cpp
ryanofsky: In commit "Keep listening after a failed accept in `ListenConnections`" (d8e5395ed8e6f21fc847b65148c2ec3c34bce524)

Would be good to add a comment here that this close is a little racy, if the socket not closed before accepted the test could appear to crash when there is still a problem.

ryanofsky (commented):
Code review 38391ebc196d7ece7152a6372993ba104dceb9c6

Nice catch and elegant test! However, I think the suggested fix is too dangerous because retrying whenever accept fails could lead to an infinite loop if the accept failed for reasons other than this macos/capnproto bug, like file descriptors being closed, or resources being exhausted.

I think it'd be good to add the test but not the fix here, and make the test allow the `setsocketopt(IPPROTO_TCP, TCP_NODELAY): Invalid argument` error (https://github.com/bitcoin-core/libmultiprocess/actions/runs/29545518652/job/87776862017#step:4:126) that results on macos when the capnproto version is less than whatever capnproto version fixes this to document the bug. Separately we should make sure this is fixed in the capnproto version we are using in Bitcoin Core and consider adding a patch to the Bitcoin Core depends build if needed temporarily.

ViniciusCestarii:
Thinking back here, I agree with @ryanofsky. In my review I understimated the spin forever on permanent failure case and the fix truly belongs to KJ and not here because if we were to try to reliably distinguish a transient error from a non transient error here we would basically be doing what KJ was supposed to do

xyzconstant:
Thanks for the reviews, addressed feedback and force-pushed: d8e5395ed8e6f21fc847b65148c2ec3c34bce524 ->  b5b357f22a37a59dd019caebe283b92eeb3b76c6 [(compare)](https://github.com/bitcoin-core/libmultiprocess/compare/d8e5395ed8e6f21fc847b65148c2ec3c34bce524..b5b357f22a37a59dd019caebe283b92eeb3b76c6)

re: https://github.com/bitcoin-core/libmultiprocess/pull/310#pullrequestreview-4724053865

> I think it'd be good to add the test but not the fix here, and make the test allow the `setsocketopt(IPPROTO_TCP, TCP_NODELAY): Invalid argument` error (https://github.com/bitcoin-core/libmultiprocess/actions/runs/29545518652/job/87776862017#step:4:126) that results on macos when the capnproto version is less than whatever capnproto version fixes this to document the bug. Separately we should make sure this is fixed in the capnproto version we are using in Bitcoin Core and consider adding a patch to the Bitcoin Core depends build if needed temporarily.

Fair concern and I think you're right. The fix was designed to pass the test, assuming this behavior would be tolerated. So it's better we drop it, even though it means the test has a workaround for this macOS bug, which I think we could be okay with.

xyzconstant:
Just opened #319 to keep track of the macOS accept crash.

test/mp/test/listen_tests.cpp
```diff
@@ -294,6 +295,38 @@ KJ_TEST("ListenConnections accepts multiple connections")
...
+    KJ_SYSCALL(close(fd));
+
+    // Wait for the connection to either be accepted and disconnected, or fail
+    // to be accepted and log the error above.
+    const auto deadline = std::chrono::steady_clock::now() + FAILURE_TIMEOUT;
+    auto accept_error_seen = [&] { Lock lock(mutex); return accept_error; };
+    while (!accept_error_seen() && server.DisconnectedCount() < 1 && std::chrono::steady_clock::now() < deadline) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+    }
+    KJ_EXPECT(accept_error_seen() || server.DisconnectedCount() >= 1);
```
ryanofsky: In commit "test: cover immediate client disconnects for `ListenConnections`" (b5b357f22a37a59dd019caebe283b92eeb3b76c6)

Would be good to tighten this check to something like  `server.DisconnectedCount() == (accept_error_seen() ? 0 : 1)` . I think this would also clarify what behavior is actually expected.

Also would be good to require `accept_error_seen()` is false unless compiling for macos ( `#ifdef __APPLE__` )

ryanofsky (approved):
Code review ACK b5b357f22a37a59dd019caebe283b92eeb3b76c6. Thanks for the test! Left suggestions below but think this could also be merged as is.

xyzconstant:
Thanks for the review @ryanofsky!

I force-pushed changes based on your suggestion to tighten the `KJ_EXPECT(...)` check, plus the `KJ_REQUIRE()` check on `accept_error_seen()` for non-macOS builds.

b5b357f22a37a59dd019caebe283b92eeb3b76c6 -> 496fb84e69f8dbf8fd9e2d29a03b44e1c73db3d2 ([compare](https://github.com/bitcoin-core/libmultiprocess/compare/b5b357f22a37a59dd019caebe283b92eeb3b76c6..496fb84e69f8dbf8fd9e2d29a03b44e1c73db3d2))

test/mp/test/listen_tests.cpp
xyzconstant: Done at 496fb84e69f8dbf8fd9e2d29a03b44e1c73db3d2

Thanks!

test/mp/test/listen_tests.cpp
```diff
@@ -294,6 +295,43 @@ KJ_TEST("ListenConnections accepts multiple connections")
...
+KJ_TEST("ListenConnections handles a client that disconnects before being accepted")
+{
+    Mutex mutex;
+    bool accept_error = false;
+    ListenSetup server(std::nullopt, [&](mp::LogMessage log) {
+        // On macOS, accept() can fail if the peer closes the connection before it is accepted.
+        // The event loop then reports this as an uncaught task exception. We catch and ignore
+        // this specific error here so that the corresponding CI job does not fail.
+        //
+        // This is a Cap'n Proto bug, a fix is available in the v2 branch at: https://github.com/capnproto/capnproto/commit/7df5bd078f389ded313479981bd0ae06cbcdfe1b#diff-ec577ad66535f58f6d7396ea51d3e56c0065308aa8fb02751cd6a8cfaa67252fR1358-R1372
```
ryanofsky, line 307: In commit "test: cover immediate client disconnects for `ListenConnections`" (496fb84e69f8dbf8fd9e2d29a03b44e1c73db3d2)

Now that the fix is merged would be nice to simplify the comment to something like `This a Cap'n Proto bug, fixed by https://github.com/capnproto/capnproto/pull/2748`

ryanofsky (approved):
Code review ACK 496fb84e69f8dbf8fd9e2d29a03b44e1c73db3d2. Since last review just tightened the checks to only allow the error on macos

PR #311, opened 2026-07-20
bugfix: clear FD_CLOEXEC in child instead of parent before fork

Sjors:
Commit 652934fb793743ab1bf962b90c7a7d6f6681c199 in #274 _sets_ `FD_CLOEXEC` in order "to ensure sockets are not leaked if processes are spawned". However it's _cleared_ too early, before forking. This PR clears it _after_ forking, but still before exec.

The first commit adds a helper for signal-safely emitting an error message, since the second commit also needs this.

The second commit and a code comment explain why it's unsafe to clear `FD_CLOEXEC` before fork. ryanofsky also described it:

> As I understand it, this race has always existed, and was not introduced in [652934f](https://github.com/bitcoin-core/libmultiprocess/commit/652934fb793743ab1bf962b90c7a7d6f6681c199). What [652934f](https://github.com/bitcoin-core/libmultiprocess/commit/652934fb793743ab1bf962b90c7a7d6f6681c199) did was start using `FD_CLOEXEC` which narrowed the race window, without completely closing it. This followup PR fixes the race more completely, but there is still a small race between creating the socketpair and applying the cause the CLOEXEC flags.
> 
> The race happens when a `SpawnProcess` call happens at the same time as a separate `fork` call in unrelated thread not using `SpawnProcess`. Because `SpawnProcess` creates a socket pair, if a separate `fork` happens in another thread, it could inherit the socket pair file descriptors and keep them open too long if it is not looping over them and closing them like `SpawnProcess` is. As I understand it this could result in `socketpair` connections staying open even after the `SpawnProcess` parent or child have closed them, so `onDisconnect` events might not be triggered, and resources might not be freed.

A test in https://github.com/Sjors/libmultiprocess/commit/1e1ff0397bf26b69c0c4b0dfffac4106490f482c demonstrates the issue, but is not included in the PR to keep things simple.

This should not be an actual problem in the way Bitcoin Core uses libmultiprocess today, but it may be with Windows and/or multiple connection support.

### DIFF
--- a/src/mp/util.cpp
+++ b/src/mp/util.cpp
@@ -60,6 +60,18 @@ size_t MaxFd()
     }
 }
 
+//! Report an error and exit from the post-fork child of a multi-threaded
+//! process, where only async-signal-safe calls (like write and _exit) are
+//! allowed. Accepting only a reference to a char array (in practice a string
+//! literal) ensures no allocation is needed at the call site.
+template <std::size_t N>
+[[noreturn]] void ChildFail(const char (&msg)[N]) noexcept
+{
+    const ssize_t written = ::write(STDERR_FILENO, msg, N - 1);
+    (void)written;
+    _exit(126);
+}
+
 } // namespace
 
 std::string ThreadName(const char* exe_name)
@@ -147,10 +159,7 @@ std::tuple<ProcessId, SocketId> SpawnProcess(SpawnConnectInfoToArgsFn&& connect_
             (void)close(fds[1]);
             throw std::system_error(errno, std::system_category(), "close");
         }
-        static constexpr char msg[] = "SpawnProcess(child): close(fds[1]) failed\n";
-        const ssize_t writeResult = ::write(STDERR_FILENO, msg, sizeof(msg) - 1);
-        (void)writeResult;
-        _exit(126);
+        ChildFail("SpawnProcess(child): close(fds[1]) failed\n");
     }
 
     if (!pid) {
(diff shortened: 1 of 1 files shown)


ryanofsky (commented):
Code review ACK 1e0c7ff9a51aedf9ddbda13dc3d8f03428da1c87. Thanks for the fix! I think it's be good to add bugfix: to the title and make PR description describe bug more practically.

As I understand it, this race has always existed, and was not introduced in 652934fb793743ab1bf962b90c7a7d6f6681c199. What 652934fb793743ab1bf962b90c7a7d6f6681c199 did was start using `FD_CLOEXEC` which narrowed the race window, without completely closing it. This followup PR fixes the race more completely, but there is still a small race between creating the socketpair and applying the cause the CLOEXEC flags.

The race happens when a `SpawnProcess` call happens at the same time as a separate `fork` call in unrelated thread not using `SpawnProcess`. Because `SpawnProcess` creates a socket pair, if a separate `fork` happens in another thread, it could inherit the socket pair file descriptors and keep them open too long if it is not looping over them and closing them like `SpawnProcess` is. As I understand it this could result in `socketpair` connections staying open even after the `SpawnProcess` parent or child have closed them, so `onDisconnect` events might not be triggered, and resources might not be freed.

ViniciusCestarii:
ACK 1e0c7ff9a51aedf9ddbda13dc3d8f03428da1c87 verified that without this change the test https://github.com/Sjors/libmultiprocess/commit/1e1ff0397bf26b69c0c4b0dfffac4106490f482c fails and that this really tightens the race window to just between socketpair and fcntl syscalls.

xyzconstant (approved):
tACK 1e0c7ff9a51aedf9ddbda13dc3d8f03428da1c87

Built and ran tests successfully. This PR adds utility changes without affecting proxy or mpgen.

LGTM, nice changes!

Sjors:
@ryanofsky I added your explanation as part of the PR description.

PR #312, opened 2026-07-20
util: report back child errors to parent and throw

ViniciusCestarii:
Report execvp() and close() failures to the parent via a socket and throw from SpawnProcess.

Previously the parent didn't have a way of knowing whether the child process successfully execute or not without polling for it or waiting for it.

Now SpawnProcess waits for the exec() to happen before returning and before throwing it also reaps the child, preventing zombie process.

Also make post-fork child not rely on on perror(), which is not async-signal-safe.

This addresses the TODO on commit 69652f0edfa10bf8e43c9513d70193bae8ab35fc

### DIFF
--- a/src/mp/util.cpp
+++ b/src/mp/util.cpp
@@ -13,6 +13,7 @@
 #include <kj/string-tree.h>
 #include <optional>
 #include <pthread.h>
+#include <csignal>
 #include <sstream>
 #include <string>
 #include <sys/types.h>
@@ -159,6 +160,16 @@ void WriteSpawnError(int fd, const SpawnError& error)
     }
 }
 
+// Get rid of a child process the parent is abandoning because SpawnProcess is
+// about to throw, so it is not left behind as a zombie. The child may still be
+// alive, so kill it first: waiting without that could block for as long as the 
+// spawned program runs.
+void KillAndReapChild(ProcessId pid)
+{
+    (void)::kill(pid, SIGKILL);
+    while (::waitpid(pid, /*status=*/nullptr, /*options=*/0) == -1 && errno == EINTR) {}
+}
+
 } // namespace
 
 std::string ThreadName(const char* exe_name)
@@ -250,6 +261,7 @@ std::tuple<ProcessId, SocketId> SpawnProcess(SpawnConnectInfoToArgsFn&& connect_
             (void)close(fds[1]);
             (void)close(error_fds[0]);
             (void)close(error_fds[1]);
+            KillAndReapChild(pid);
             throw std::system_error(err, std::system_category(), "close");
         }
         WriteSpawnError(error_fds[0], {.which = SpawnErrorOp::CLOSE, .err = err});
@@ -291,6 +303,7 @@ std::tuple<ProcessId, SocketId> SpawnProcess(SpawnConnectInfoToArgsFn&& connect_
         const int err = errno;
         (void)close(error_fds[1]);
         (void)close(fds[1]);
+        KillAndReapChild(pid);
         throw std::system_error(err, std::system_category(), "close");
     }
 
@@ -298,6 +311,7 @@ std::tuple<ProcessId, SocketId> SpawnProcess(SpawnConnectInfoToArgsFn&& connect_
     (void)close(error_fds[1]);
     if (error) {
         (void)close(fds[1]);
+        KillAndReapChild(pid);
         throw std::system_error(error->err, std::system_category(), SpawnErrorName(error->which));
     }
     return {pid, fds[1]};
--- a/src/mp/util.cpp
+++ b/src/mp/util.cpp
@@ -11,6 +11,7 @@
 #include <kj/common.h>
 #include <kj/debug.h>
 #include <kj/string-tree.h>
+#include <optional>
 #include <pthread.h>
 #include <sstream>
 #include <string>
@@ -70,6 +71,92 @@ template <std::size_t N>
     const ssize_t written = ::write(STDERR_FILENO, msg, N - 1);
     (void)written;
     _exit(126);
+}
+
+enum class SpawnErrorOp
+{
+    CLOSE,
+    EXECVP,
+    READ,
+    FCNTL,
+};
+
+struct SpawnError {
+    SpawnErrorOp which;
+    int err;
+};
+
+const char* SpawnErrorName(SpawnErrorOp which)
+{
+    switch (which) {
+    case SpawnErrorOp::CLOSE: return "close";
+    case SpawnErrorOp::EXECVP: return "execvp";
+    case SpawnErrorOp::READ: return "read";
+    case SpawnErrorOp::FCNTL: return "fcntl";
+    }
+    return "unknown";
+}
+
+// Read the child's error report. Returns nullopt on success, or a SpawnError to
+// throw on failure. Success is a clean EOF: read() returns 0 with nothing
+// buffered because the child's write end was closed by a successful exec (via
+// FD_CLOEXEC). A fully-read struct is the failure the child reported. A read()
+// error, or an EOF partway through the struct (the child died mid-report), is
+// surfaced as a SpawnErrorOp::READ failure rather than being mistaken for
+// success. A single read() is not guaranteed to return all sizeof(SpawnError)
+// bytes (the socket is SOCK_STREAM, which has no message boundaries) and may be
+// interrupted by a signal, so loop until the whole struct is read.
+//
+// Note that a clean EOF is also what happens if the child is killed before it
+// reaches exec, so this function reports success in that case. There is no good
+// way to distinguish the two here, but callers will still see the failure when
+// they call WaitProcess and get the child's exit status.
+std::optional<SpawnError> ReadSpawnResult(int fd)
+{
+    SpawnError error{};
+    char* buf = reinterpret_cast<char*>(&error);
+    size_t remaining = sizeof(error);
+    while (remaining > 0) {
+        const ssize_t n = ::read(fd, buf, remaining);
+        if (n < 0) {
+            if (errno == EINTR) continue;
+            return SpawnError{.which = SpawnErrorOp::READ, .err = errno};
+        }
+        if (n == 0) {
+            if (remaining == sizeof(error)) return std::nullopt; // clean EOF: success
+            return SpawnError{.which = SpawnErrorOp::READ, .err = EPROTO}; // torn report
+        }
+        buf += n;
+        remaining -= static_cast<size_t>(n);
+    }
+    return error;
+}
+
+// Write the whole SpawnError to fd, retrying short writes and EINTR so the
+// parent never sees a torn struct. Runs in the post-fork child, so it must stay
+// async-signal-safe: it only calls write() and does not allocate or throw. This
+// is best-effort -- if the write cannot complete there is nothing useful the
+// child can do, so it stops and lets the caller _exit().
+void WriteSpawnError(int fd, const SpawnError& error)
+{
+    const char* buf = reinterpret_cast<const char*>(&error);
+    size_t remaining = sizeof(error);
+    while (remaining > 0) {
+        const ssize_t n = ::write(fd, buf, remaining);
+        if (n < 0) {
+            if (errno == EINTR) continue;
+            break;
+        }
+        buf += n;
+        remaining -= static_cast<size_t>(n);
+    }
+    if (remaining > 0) {
+        // The parent's read end is gone (e.g. the parent exited before the
+        // child could report), so the structured error can't be delivered.
+        // Leave a breadcrumb on stderr and exit. The exit code is irrelevant
+        // here since no live parent remains to wait on it.
+        ChildFail("SpawnProcess(child): failed and could not report error to parent\n");
+    }
 }
 
 } // namespace
@@ -132,6 +219,9 @@ std::string LogEscape(const kj::StringTree& string, size_t max_size)
 std::tuple<ProcessId, SocketId> SpawnProcess(SpawnConnectInfoToArgsFn&& connect_info_to_args)
 {
     auto fds{SocketPair()};
+    // Only used for the child to report errors back to the parent, so a one-way
+    // pipe would be sufficient, but reuse the existing SocketPair() helper.
+    auto error_fds{SocketPair()};
 
     // Evaluate the callback and build the argv array before forking.
     //
@@ -144,25 +234,35 @@ std::tuple<ProcessId, SocketId> SpawnProcess(SpawnConnectInfoToArgsFn&& connect_
 
     ProcessId pid = fork();
     if (pid == -1) {
-        throw std::system_error(errno, std::system_category(), "fork");
+        const int err = errno;
+        (void)close(fds[0]);
+        (void)close(fds[1]);
+        (void)close(error_fds[0]);
+        (void)close(error_fds[1]);
+        throw std::system_error(err, std::system_category(), "fork");
     }
     // Parent process closes the descriptor for socket 0, child closes the
     // descriptor for socket 1. On failure, the parent throws, but the child
     // must _exit(126) (post-fork child must not throw).
     if (close(fds[pid ? 0 : 1]) != 0) {
+        const int err = errno;
         if (pid) {
             (void)close(fds[1]);
-            throw std::system_error(errno, std::system_category(), "close");
+            (void)close(error_fds[0]);
+            (void)close(error_fds[1]);
+            throw std::system_error(err, std::system_category(), "close");
         }
-        ChildFail("SpawnProcess(child): close(fds[1]) failed\n");
+        WriteSpawnError(error_fds[0], {.which = SpawnErrorOp::CLOSE, .err = err});
+        _exit(126);
     }
 
     if (!pid) {
         // Child process must close all potentially open descriptors, except
-        // socket 0. Do not throw, allocate, or do non-fork-safe work here.
+        // socket 0 and the error-reporting socket 0. Do not throw, allocate, or
+        // do non-fork-safe work here.
         const int maxFd = MaxFd();
         for (int fd = 3; fd < maxFd; ++fd) {
-            if (fd != fds[0]) {
+            if (fd != fds[0] && fd != error_fds[0]) {
                 close(fd);
             }
         }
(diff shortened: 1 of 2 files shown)


src/mp/util.cpp
ryanofsky: In commit "util: report back child execvp error to parent and throw" (dee4383be1a6991adaf3da55d99e71cf853b6a09)

Could be good to extend the PR to cover this error as well and translate it to an exception in the parent. Might be good to introduce a struct like:

```c++
  struct ChildError {
      int which;  // e.g. 1=close, 2=fcntl, 3=execvp
      int err;    // errno
  };
```

and write that to the pipe, then have the parent throw

```c++
throw std::system_error(error.err, std::system_category(), name_for(error.which))
```

ryanofsky (approved):
Concept ACK dee4383be1a6991adaf3da55d99e71cf853b6a09. I plan to review more and left a suggestion below.

I think PR description might be underselling the change, because it describes the main advantage as avoiding a perror call that is not signal safe (so could hang). But I think a bigger improvement is translating the execvp failure in the child into an exception in the parent. Previously the parent didn't have a way of knowing whether the child process successfully execute or not without polling for it or waiting for it.

ViniciusCestarii:
Thanks for reviewing @ryanofsky! I updated the PR description to sell this PR more and I will address your comment.

src/mp/util.cpp
xyzconstant, line 266: If this `close()` call fails in the parent, it throws while the `error_fds` pair is still open, leaking those descriptors.

Both ends of `error_fds` should be closed in the parent before throwing, and the original `errno` should be saved in a local variable so it isn’t overwritten by subsequent `close()` calls.

```diff
     if (close(fds[pid ? 0 : 1]) != 0) {
+        const int error = errno;
         if (pid) {
             (void)close(fds[1]);
-            throw std::system_error(errno, std::system_category(), "close");
+            (void)close(error_fds[0]);
+            (void)close(error_fds[1]);
+            throw std::system_error(error, std::system_category(), "close");
         }
```

src/mp/util.cpp
```diff
@@ -155,22 +156,45 @@ std::tuple<ProcessId, SocketId> SpawnProcess(SpawnConnectInfoToArgsFn&& connect_
...
+        const ssize_t writeResult = ::write(error_fds[0], &error, sizeof(error));
+        (void)writeResult;
         _exit(127);
     }
+
+    // Close the parent's copy of the child's write end.
+    if (close(error_fds[0]) != 0) {
+        (void)close(fds[1]);
+        throw std::system_error(errno, std::system_category(), "close");
+    }
```
xyzconstant, line 308: `error_fds[1]` is leaking here too, close it (plus the same stale-errno fix as above).

```diff
     if (close(error_fds[0]) != 0) {
+        const int error = errno;
+        (void)close(error_fds[1]);
         (void)close(fds[1]);
-        throw std::system_error(errno, std::system_category(), "close");
+        throw std::system_error(error, std::system_category(), "close");
     }
```

src/mp/util.cpp
xyzconstant, line 254: There are also more descriptor leaks here that were already present. Since this PR adds an additional pair it doubles the issue. It might be a good idea to close all four of them while we're at it.

```diff
     if (pid == -1) {
+        const int error = errno;
+        (void)close(fds[0]);
+        (void)close(fds[1]);
+        (void)close(error_fds[0]);
+        (void)close(error_fds[1]);
-        throw std::system_error(errno, std::system_category(), "fork");
+        throw std::system_error(error, std::system_category(), "fork");
     }
```

src/mp/util.cpp
```diff
@@ -155,22 +156,45 @@ std::tuple<ProcessId, SocketId> SpawnProcess(SpawnConnectInfoToArgsFn&& connect_
...
+    if (readResult < 0) {
+        (void)close(fds[1]);
+        throw std::system_error(errno, std::system_category(), "read");
+    }
+    if (readResult > 0 && error) {
+        (void)close(fds[1]);
+        // Wait it here to avoid leaking a zombie process and then throw.
+        (void)::waitpid(pid, nullptr, 0);
+        throw std::system_error(error, std::system_category(), "execvp");
+    }
```
xyzconstant: It might be worth extracting this into a small helper, e.g.

```cpp
//! Read the child's exec result from the error socket, then close it.
//! Returns 0 on success, otherwise an errno.
int ReadChildErrno(int fd);
```

So `SpawnProcess` has a single close/waitpid/throw site instead of repeating cleanup in each error branch.

xyzconstant (commented):
Concept ACK dee4383be1a6991adaf3da55d99e71cf853b6a09

Left some comments about descriptor leaks on the error paths. Planning to review more once you update on ryanofsky's feedback.

Sjors:
Note that this change happens very close to the code touched by #311, but does seem orthogonal at first glance.

src/mp/util.cpp
ViniciusCestarii: Nice catch! Fixed this on d4177062b3f4e7113aab9d40d994eabe711cb371

src/mp/util.cpp
ViniciusCestarii: Nice! Fixed on d4177062b3f4e7113aab9d40d994eabe711cb371

src/mp/util.cpp
ViniciusCestarii: Nice catch! Fixed on d4177062b3f4e7113aab9d40d994eabe711cb371

src/mp/util.cpp
ViniciusCestarii: Yes I agree that it is better to extract this. I have added a function to extract it on a761acaa0897d8e5326996d58315d3d50e0f387c (note that I updated the code to use SpawnError)

src/mp/util.cpp
ViniciusCestarii: I agree, I have added ChildError to cover execv and close errors on d4177062b3f4e7113aab9d40d994eabe711cb371 and then added a helper function to read the error on a761acaa0897d8e5326996d58315d3d50e0f387c and renamed it to SpawnError to cover the parent error on reading from socket.

ViniciusCestarii:
Thanks @xyzconstant for reviewing! I have force-pushed d4177062b3f4e7113aab9d40d994eabe711cb371 and pushed a761acaa0897d8e5326996d58315d3d50e0f387c and updated PR title and description.

I have addresed:

- @xyzconstant comment about leaking fds and throwing wrong errno and @ryanofsky comment for reporting a struct ChildError to cover execv and close errors on d4177062b3f4e7113aab9d40d994eabe711cb371. 
- @xyzconstant comment that this PR could benefit with a function for reading the error from child so I added ReadSpawnResult on a761acaa0897d8e5326996d58315d3d50e0f387c and also renamed ChildError to SpawnError to cover the parent error on reading from socket.

I believe this now conflicts with #311 but it should be a simple rebase for whichever PR merges second.

I also noticed that there are several throws after fork that could benefit from reaping the child to prevent zombie process, but I believe this could be a follow up.

src/mp/util.cpp
```diff
@@ -117,9 +117,30 @@ std::string LogEscape(const kj::StringTree& string, size_t max_size)
     return result;
 }
 
+enum class ChildErrorOp
```
ryanofsky: In commit "util: report back child error to parent and throw" (d4177062b3f4e7113aab9d40d994eabe711cb371)

It would be good to move this to anonymous namespace above below MakeArgv and MaxFd so functions here don't get exported as symbols and so this code can be skipped in the same `#ifndef WIN32` on windows in #231

src/mp/util.cpp
```diff
@@ -117,26 +117,41 @@ std::string LogEscape(const kj::StringTree& string, size_t max_size)
...
-    case ChildErrorOp::CLOSE: return "close";
-    case ChildErrorOp::EXECVP: return "execvp";
+    case SpawnErrorOp::CLOSE: return "close";
+    case SpawnErrorOp::EXECVP: return "execvp";
+    case SpawnErrorOp::READ: return "read";
     }
     return "unknown";
 }
 
+SpawnError ReadSpawnResult(int fd)
```
ryanofsky: In commit "util: extract ReadSpawnResult and report read() failures as SpawnError" (a761acaa0897d8e5326996d58315d3d50e0f387c)

Changes here looks good, but seems like this commit should be squashed in the first commit.

ryanofsky: In commit "util: extract ReadSpawnResult and report read() failures as SpawnError" (a761acaa0897d8e5326996d58315d3d50e0f387c)

Might want to handle short reads here and also short/interrupted writes. Claude suggests:

```c++
// Read the child's error report. Returns nullopt on success, or a SpawnError to
// throw on failure. Success is a clean EOF: read() returns 0 with nothing
// buffered because the child's write end was closed by a successful exec (via
// FD_CLOEXEC). A fully-read struct is the failure the child reported. A read()
// error, or an EOF partway through the struct (the child died mid-report), is
// surfaced as a SpawnErrorOp::READ failure rather than being mistaken for
// success. A single read() is not guaranteed to return all sizeof(SpawnError)
// bytes (the socket is SOCK_STREAM, which has no message boundaries) and may be
// interrupted by a signal, so loop until the whole struct is read.
std::optional<SpawnError> ReadSpawnResult(int fd)
{
    SpawnError error{};
    char* buf = reinterpret_cast<char*>(&error);
    size_t remaining = sizeof(error);
    while (remaining > 0) {
        const ssize_t n = ::read(fd, buf, remaining);
        if (n < 0) {
            if (errno == EINTR) continue;
            return SpawnError{.which = SpawnErrorOp::READ, .err = errno};
        }
        if (n == 0) {
            if (remaining == sizeof(error)) return std::nullopt; // clean EOF: success
            return SpawnError{.which = SpawnErrorOp::READ, .err = EPROTO}; // torn report
        }
        buf += n;
        remaining -= static_cast<size_t>(n);
    }
    return error;
}

// Write the whole SpawnError to fd, retrying short writes and EINTR so the
// parent never sees a torn struct. Runs in the post-fork child, so it must stay
// async-signal-safe: it only calls write() and does not allocate or throw. This
// is best-effort -- if the write cannot complete there is nothing useful the
// child can do, so it stops and lets the caller _exit().
void WriteSpawnError(int fd, const SpawnError& error)
{
    const char* buf = reinterpret_cast<const char*>(&error);
    size_t remaining = sizeof(error);
    while (remaining > 0) {
        const ssize_t n = ::write(fd, buf, remaining);
        if (n < 0) {
            if (errno == EINTR) continue;
            break;
        }
        buf += n;
        remaining -= static_cast<size_t>(n);
    }
    if (remaining > 0) {
        // The parent's read end is gone (e.g. the parent exited before the
        // child could report), so the structured error can't be delivered.
        // Leave a breadcrumb on stderr and exit. The exit code is irrelevant
        // here since no live parent remains to wait on it.
        ChildFail("SpawnProcess(child): failed and could not report error to parent\n");
    }
}
```

Note this uses ChildFail from #311 (which will probably be merged first so should be available)

src/mp/util.cpp
```diff
@@ -117,9 +117,30 @@ std::string LogEscape(const kj::StringTree& string, size_t max_size)
...
+    case ChildErrorOp::CLOSE: return "close";
+    case ChildErrorOp::EXECVP: return "execvp";
+    }
+    return "unknown";
+}
+
 std::tuple<ProcessId, SocketId> SpawnProcess(SpawnConnectInfoToArgsFn&& connect_info_to_args)
 {
     auto fds{SocketPair()};
+    auto error_fds{SocketPair()};
```
ryanofsky, line 235: In commit "util: report back child error to parent and throw" (d4177062b3f4e7113aab9d40d994eabe711cb371)

I think current approach is good and probably best to use a socketpair instead of a pipe, but might be worth a comment that this is only used for child to send errors to the parent so a one-way pipe would be sufficient

ryanofsky (approved):
Code review ACK a761acaa0897d8e5326996d58315d3d50e0f387c. Nice changes that should substantially improve error reporting.

I left some suggestions below but they are not important, this should already be an improvement over current code. Also I think #311 will be merged first and this will need to be rebased, but the two pr's should complement each other and not conflict very much.

Would note that even with #311 there is a still a race condition since CLOEXEC flag is not applied right away and even with this PR there are still errors not handled like connect_info_to_args or KJ_SYSCALL syscall throwing which could cause leaks, so more improvements could be made as followups.

The posix code is definitely getting more complicated as a result of these error handling improvements (especially compared to the windows code) but it seems like there might not be a way to avoid this with current Bitcoin core code which is not applying CLOEXEC flags, so file descriptors will leak into child processes. This prevents using posix_spawn instead of fork/exec.

ViniciusCestarii:
Thanks again for the reviews! Forced-push eecda07ea4ae8f7bbb592bdbf00a19d9c43458ca squashing into a single commit and basing these changes on #311. I also addressed @ryanofsky comments.

src/mp/util.cpp
ViniciusCestarii: Nice, done on eecda07ea4ae8f7bbb592bdbf00a19d9c43458ca.

src/mp/util.cpp
ViniciusCestarii: This is nicer, added on eecda07ea4ae8f7bbb592bdbf00a19d9c43458ca.

ViniciusCestarii: Squashed on eecda07ea4ae8f7bbb592bdbf00a19d9c43458ca.

src/mp/util.cpp
ViniciusCestarii: Nice, done on eecda07ea4ae8f7bbb592bdbf00a19d9c43458ca.

src/mp/util.cpp
```diff
@@ -174,17 +269,32 @@ std::tuple<ProcessId, SocketId> SpawnProcess(SpawnConnectInfoToArgsFn&& connect_
...
+        (void)close(error_fds[1]);
+        (void)close(fds[1]);
+        throw std::system_error(err, std::system_category(), "close");
+    }
+
+    const std::optional<SpawnError> error{ReadSpawnResult(error_fds[1])};
+    (void)close(error_fds[1]);
+    if (error) {
+        (void)close(fds[1]);
+        throw std::system_error(error->err, std::system_category(), SpawnErrorName(error->which));
```
ryanofsky, line 315: In commit "util: report back child error to parent and throw" (eecda07ea4ae8f7bbb592bdbf00a19d9c43458ca)

In the places there this function is throwing after calling fork() and it returns a child pid (lines 248, 289, and 296) it looks like this code will leak the pid and leave behind a zombie process. I think it would make sense to call waitpid these places, with the pid and maybe with NOHANG.

src/mp/util.cpp
```diff
@@ -72,6 +73,87 @@ template <std::size_t N>
...
+
+// Read the child's error report. Returns nullopt on success, or a SpawnError to
+// throw on failure. Success is a clean EOF: read() returns 0 with nothing
+// buffered because the child's write end was closed by a successful exec (via
+// FD_CLOEXEC). A fully-read struct is the failure the child reported. A read()
+// error, or an EOF partway through the struct (the child died mid-report), is
+// surfaced as a SpawnErrorOp::READ failure rather than being mistaken for
+// success. A single read() is not guaranteed to return all sizeof(SpawnError)
+// bytes (the socket is SOCK_STREAM, which has no message boundaries) and may be
+// interrupted by a signal, so loop until the whole struct is read.
```
ryanofsky, line 109: In commit "util: report back child error to parent and throw" (eecda07ea4ae8f7bbb592bdbf00a19d9c43458ca)

Would be good to note that an EOF can also happen if the child is killed before running exec, and this will return success in that case. There isn't really a good way for this function to avoid this but callers should see the failure when they call WaitProcess

ryanofsky (approved):
Code review ACK eecda07ea4ae8f7bbb592bdbf00a19d9c43458ca. Left minor suggests below but this looks good as-is.

Might want to note in commit messages or PR description that this does change SpawnProcess behavior a little bit beyond error reporting since it now waits for the exec() to happen before returning instead of just the fork() so it might be a little slower (not that it should matter in practice)

src/mp/util.cpp
ViniciusCestarii: Good catch. Fixed it in a separate commit: 3f05b11624ce02e35b52e7451b2c7092b9b48a7c

I went with SIGKILL followed by a blocking wait rather than NOHANG because at some places the child may still be alive and racing toward exec, so NOHANG would usually return 0 and leave the zombie behind. After a SIGKILL the wait is guaranteed to reap and returns immediately.

src/mp/util.cpp
ViniciusCestarii: Nice catch. I added a note in 4a56c1837a781fbd859c4a3ef91fdd9e213b42a7

ViniciusCestarii:
Thanks for the review! Forced push 3f05b11624ce02e35b52e7451b2c7092b9b48a7c addressing https://github.com/bitcoin-core/libmultiprocess/pull/312#discussion_r3706130092 adding a note on function ReadSpawnResult and https://github.com/bitcoin-core/libmultiprocess/pull/312#discussion_r3705995397 by adding the function KillAndReapChild and calling it before throwing.

Updated PR description with: SpawnProcess waits for the exec() to happen before returning and before throwing it also reaps the child, preventing zombie process.

xyzconstant:
LGTM. ACK 3f05b11624ce02e35b52e7451b2c7092b9b48a7c

enirox001 (commented):
Concept ACK

src/mp/util.cpp
```diff
@@ -174,17 +274,32 @@ std::tuple<ProcessId, SocketId> SpawnProcess(SpawnConnectInfoToArgsFn&& connect_
         // fcntl is async-signal-safe.
         const int fds0_flags = fcntl(fds[0], F_GETFD);
         if (fds0_flags == -1 || fcntl(fds[0], F_SETFD, fds0_flags & ~FD_CLOEXEC) == -1) {
-            ChildFail("SpawnProcess(child): clearing FD_CLOEXEC failed\n");
+            const int err = errno;
+            WriteSpawnError(error_fds[0], {.which = SpawnErrorOp::FCNTL, .err = err});
+            _exit(126);
         }
 
         execvp(argv[0], argv.data());
```
ryanofsky, line 282: In commit "util: report back child error to parent and throw" (4a56c1837a781fbd859c4a3ef91fdd9e213b42a7)

Might be good to comment above that execvp line, that if execvp succeeds, this will automatically close `error_fds[0]` (because it is FD_CLOEXEC), sending EOF (success) to the parent process.

I didn't actually realize until now that this whole approach depends on FD_CLOEXEC and couldn't work without it.

src/mp/util.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/312#discussion_r3705995397

Make sense sense. IIUC if NOHANG approach was used it might appear to work but could still be racy and leave behind zombie processes.

src/mp/util.cpp
```diff
@@ -144,25 +234,35 @@ std::tuple<ProcessId, SocketId> SpawnProcess(SpawnConnectInfoToArgsFn&& connect_
 
     ProcessId pid = fork();
     if (pid == -1) {
-        throw std::system_error(errno, std::system_category(), "fork");
+        const int err = errno;
+        (void)close(fds[0]);
+        (void)close(fds[1]);
```
ryanofsky, line 239: In commit "util: report back child error to parent and throw" (4a56c1837a781fbd859c4a3ef91fdd9e213b42a7)

Just noting since I didn't see it noted elsewhere that these two close calls fix a leak that would have happened previously if fork() failed.

ryanofsky (approved):
Code review ACK 3f05b11624ce02e35b52e7451b2c7092b9b48a7c. Looks great and makes the libmultiprocess library more competent at spawning processes, actually providing specific error information if this fails

PR #314, opened 2026-07-21
Export compile_commands.json in standalone builds (by default) 

Sjors:
This makes the default build configuration play nicely with code editors. E.g. with Zed, looking at `spawn_argv` in #231:

Before:
before

After:
after

One downside of this approach is that it only just works(tm) on the `build` directory.

There are several alternative approaches:

1. Update documentation to recommend:

```
cmake -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
ln -sf build/compile_commands.json .
```

2. Have cmake add a symlink to the source dir
3. Add a (dev) cmake preset

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -1,0 +1,6 @@
+# Point clangd at the compile database in the build directory suggested by
+# doc/install.md. clangd only searches ancestor directories by default, so
+# without this a symlink to compile_commands.json would be needed in the
+# source directory.
+CompileFlags:
+  CompilationDatabase: build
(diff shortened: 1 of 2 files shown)


ryanofsky:
Code review fab9081bc3a1ef37c521528b466df7d5d1c9df8c.

I think we should do something to make this less cumbersome, but I'm not sure about this approach, especially since it hardcodes the name of the `build` directory. The simplest thing to do would be to suggest running `cmake -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=ON && ln -sf build/compile_commands.json .` in documentation to set this up.

But I've also been thinking it would be good to have a clearer separation in the cmake build between essential features that can't be provided externally and convenience features and maybe just enable the convenience features in a dev mode present like Bitcoin core. Sketch of this idea is: https://github.com/bitcoin-core/libmultiprocess/compare/master...ryanofsky:libmultiprocess:pr/extras ([branch]( https://github.com/ryanofsky/libmultiprocess/commits/pr/extras)) which would automatically set up this editor support if dev mode is enabled while avoiding a hardcoded .clangd file, and simplifying the main cmake file instead of complicating it.

Sjors:
Closing in favor of your suggested approach. I won't have time in the next two weeks to implement that though.

PR #315, opened 2026-07-22
Fix startup race in example

xyzconstant:
Since #274, connection setup has been split into two calls: `MakeStream` followed by `ConnectStream`. Between these calls, the event loop's reference count can temporarily drop to zero, allowing `EventLoop::loop()` to exit (and the EventLoop to be destroyed) before `ConnectStream` posts its work. This crashes mpexample on startup.

Fix this by keeping an `EventLoopRef` alive in main() so the loop stays running while it is in use. This same pattern is already used in Bitcoin Core (`m_loop_ref` member of `CapnpProtocol`).

A second commit documents the reference-counted EventLoop lifetime in the `EventLoopRef` class comment and in `doc/usage.md`.

NOTE: On macOS, mpexample was crashing intermittently on startup (sometimes failing the m_post_fn == nullptr assert in the EventLoop destructor, sometimes with a mutex lock failure), depending on timing. With this change, the crashes are gone.

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -13,6 +13,7 @@
 #include <kj/async.h>
 #include <kj/common.h>
 #include <memory>
+#include <mp/proxy.h>
 #include <mp/proxy-io.h>
 #include <mp/util.h>
 #include <stdexcept>
@@ -55,6 +56,7 @@ int main(int argc, char** argv)
         loop.loop();
     });
     mp::EventLoop* loop = promise.get_future().get();
+    mp::EventLoopRef loop_ref{*loop};
 
     auto [printer_init, printer_pid] = Spawn(*loop, argv[0], "mpprinter");
     auto [calc_init, calc_pid] = Spawn(*loop, argv[0], "mpcalculator");
@@ -71,6 +73,7 @@ int main(int argc, char** argv)
     mp::WaitProcess(calc_pid);
     printer_init.reset();
     mp::WaitProcess(printer_pid);
+    loop_ref.reset();
     loop_thread.join();
     std::cout << "Bye!" << std::endl;
     return 0;
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -12,6 +12,12 @@ The `*.capnp` data definition files are consumed by the _libmultiprocess_ code g
 
 The `ProxyServer` objects help translate IPC requests from a socket to method calls on a local object. The `ProxyServer` objects are just used internally by the `mp::ServeStream(loop, socket, wrapped_object)` and `mp::ListenConnections(loop, socket, wrapped_object[, max_connections])` functions, and aren't exposed externally. The `ProxyClient` classes are exposed, and returned from the `mp::ConnectStream(loop, socket)` function and meant to be used directly. The classes implement methods described in `.capnp` definitions, and whenever any method is called, a request with the method arguments is sent over the associated IPC connection, and the corresponding `wrapped_object` method on the other end of the connection is called, with the `ProxyClient` method blocking until it returns and forwarding back any return value to the `ProxyClient` method caller.
 
+## Event loop
+
+Making connections and IPC calls requires an `mp::EventLoop` instance running in its own dedicated thread. The event loop is reference-counted, and `EventLoop::loop()` returns when the last `mp::EventLoopRef` is released. Therefore, any code that needs to use the loop across multiple calls should maintain its own `EventLoopRef`. Connection objects and proxy objects automatically hold a reference for as long as they exist.
+
+For a typical setup, see [example.cpp](../example/example.cpp).
+
 ## Example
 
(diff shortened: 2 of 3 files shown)


xyzconstant:
failed CI job seems unrelated

allibmb1997-collab (approved):

allibmb1997-collab (approved):

ViniciusCestarii (commented):
tACK f2fabf668c4bfb1c55fd41f993b7a4bd754a8653 nice catch! I was able to reproduce on linux and it also intermittently fails without this fix. I liked the doc explanation too.

nit: a more surgical fix for against against master could look like:

```diff
+#include "mp/proxy.h"
 #include <init.capnp.h>
 #include <init.capnp.proxy.h>
 
int main(int argc, char** argv)
         loop.loop();
     });
     mp::EventLoop* loop = promise.get_future().get();
+    mp::EventLoopRef loop_ref{*loop};
 
     auto [printer_init, printer_pid] = Spawn(*loop, argv[0], "mpprinter");
     auto [calc_init, calc_pid] = Spawn(*loop, argv[0], "mpcalculator");
int main(int argc, char** argv)
     mp::WaitProcess(calc_pid);
     printer_init.reset();
     mp::WaitProcess(printer_pid);
+    loop_ref.reset();
     loop_thread.join();
     std::cout << "Bye!" << std::endl;
     return 0;
```

example/example.cpp
```diff
@@ -24,14 +25,15 @@
 
 namespace fs = std::filesystem;
 
-static auto Spawn(mp::EventLoop& loop, const std::string& process_argv0, const std::string& new_exe_name)
+static auto Spawn(const mp::EventLoopRef& loop_ref, const std::string& process_argv0, const std::string& new_exe_name)
```
ryanofsky: In commit "Fix startup race in mpexample" (f433a3f5a19db6c190615ceb60ba9f962ed49a29)

I don't think it actually makes sense to change this function and pass in a loop ref object when this isn't going to actually change the reference count. Would be better to revert changes to this function.

Similarly I think most of the changes below are not needed. Would suggest a simpler fix:

```diff
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -55,6 +55,7 @@ int main(int argc, char** argv)
         loop.loop();
     });
     mp::EventLoop* loop = promise.get_future().get();
+    mp::EventLoopRef loop_ref{*loop};
 
     auto [printer_init, printer_pid] = Spawn(*loop, argv[0], "mpprinter");
     auto [calc_init, calc_pid] = Spawn(*loop, argv[0], "mpcalculator");
@@ -71,6 +72,7 @@ int main(int argc, char** argv)
     mp::WaitProcess(calc_pid);
     printer_init.reset();
     mp::WaitProcess(printer_pid);
+    loop_ref.reset();
     loop_thread.join();
     std::cout << "Bye!" << std::endl;
     return 0;
```

example/example.cpp
```diff
@@ -48,16 +50,16 @@ int main(int argc, char** argv)
         return 1;
     }
 
-    std::promise<mp::EventLoop*> promise;
+    std::promise<mp::EventLoopRef> promise;
     std::thread loop_thread([&] {
         mp::EventLoop loop("mpexample", LogPrint);
-        promise.set_value(&loop);
+        promise.set_value(mp::EventLoopRef(loop));
```
ryanofsky: In commit "Fix startup race in mpexample" (f433a3f5a19db6c190615ceb60ba9f962ed49a29)

Suggested an alternative in diff above, and I think creating an EventLoopRef here and then moving it into another EventLoopRef variable is too complicated and not necessary. It should only be necessary to increment the event loop usage count before using the event loop, and doesn't have to be done when creating it.

ryanofsky (approved):
Code review ACK 265302bc418136e072e37aa5de70ec7791679076. Nice catch and this fix looks correct, but it seems a little more complicated than it needs to be and I suggested a simpler version below.

I was initially confused how #274 causes this bug, but it happens specifically because of the `loop.sync` call in [`MakeStream`](https://github.com/bitcoin-core/libmultiprocess/blob/3f221b5bfd7ee0e7972e3c5ed4bb7ee86e457f6d/src/mp/proxy.cpp#L48). Each time `loop.sync` is called it causes the event loop to spin and check the reference count and potentially exit. The sync call in `MakeStream` is actually not necessary in practice, but was added as [precaution](https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3596039776) in case the the wrapFd call was changed to update shared state. But because creating the stream object and connecting used to happen in one sync call and now happens in two calls, adding an extra reference to the event loop is necessary.

xyzconstant:
Thanks for the reviews!

The same diff has been suggested twice, so I just applied it in the latest push (36c6c63520e2f23a67c21c21598adc1e2672b9cf).

With #323 now merged, CI jobs should be green.

example/example.cpp
xyzconstant: Done at 3a997e113cf781222022d2a61ce15e05efd43cdd

Thanks!

ryanofsky (approved):
Code review ACK 36c6c63520e2f23a67c21c21598adc1e2672b9cf. Thanks for the fix, documentation, and simplification!

PR #316, opened 2026-07-22
cmake: Fix stale codegen when mpgen binary changes

ViniciusCestarii:
Add MPGEN_BINARY to DEPENDS so target_capnp_sources() also rebuilds when mpgen binary changes. This prevents stale code gen when mpgen binary changes but capnp files don't.

I noticed this after seeing IWYU errors while building the examples, which were caused by stale generated code that no longer matched what the current mpgen produces.


hebasto (approved):
ACK eea9c64f6e1682f7ad7c4a0e7ffbcc8c61619b21.

From CMake [docs](https://cmake.org/cmake/help/latest/command/add_custom_command.html):
> List target names with the `DEPENDS` option to add ... file-level dependencies.

xyzconstant:
ACK eea9c64f6e1682f7ad7c4a0e7ffbcc8c61619b21

>I noticed this after seeing IWYU errors while building the examples, which was caused by stale generated code that no longer matched what the current mpgen produces.

Nice catch! I experienced this too, but I ignored it thinking there was something wrong with my machine.

ryanofsky (approved):
Code review ACK eea9c64f6e1682f7ad7c4a0e7ffbcc8c61619b21. Nice fix which should make testing generator changes more convenient.

PR #317, opened 2026-07-22
ipc: Fix mpgen capnp tool path for vcpkg/Windows builds

ryanofsky:
It seems that in the vcpkg capnproto package, the `capnp` code generator can't be found in `<capnp_PREFIX>/bin/capnp` but is in another path that needs to be read from the cmake configuration, so this change adds cmake code to handle that.

Unfortunately the previous code using `<capnp_PREFIX>/bin` can't be removed because the llbcapnp-dev 1.0.1 package in Ubuntu Noble is buggy and specifies the wrong executable path in its cmake information. (If this problem is fixed, the old `capnp_PREFIX` code could be dropped.)

This change is needed to get MSVC builds working in https://github.com/bitcoin/bitcoin/pull/32387

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -209,6 +209,16 @@ target_link_libraries(mpgen PRIVATE CapnProto::capnp-rpc)
 target_link_libraries(mpgen PRIVATE CapnProto::capnpc)
 target_link_libraries(mpgen PRIVATE CapnProto::kj)
 target_link_libraries(mpgen PRIVATE Threads::Threads)
+# Use CAPNP_EXECUTABLE/CAPNPC_CXX_EXECUTABLE cmake variables set by
+# CapnProtoConfig.cmake rather than reading the CapnProto::capnp_tool
+# IMPORTED_LOCATION directly. On Debian/Ubuntu, IMPORTED_LOCATION is broken
+# by a multiarch cmake packaging bug (see
+# https://github.com/bitcoin-core/libmultiprocess/issues/328), but
+# CapnProtoConfig.cmake works around it by hardcoding the correct paths. On
+# other platforms the variables hold generator expressions that
+# target_compile_definitions evaluates correctly at generation time.
+target_compile_definitions(mpgen PRIVATE CAPNP_EXECUTABLE=\"${CAPNP_EXECUTABLE}\")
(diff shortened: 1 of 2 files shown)


hebasto (commented):
I reviewed 049dae17b5d0b59858e6ead1dec40e5e778f44d7 and played with the code on Windows using the following `vcpkg.json` file:
```json
{
  "$comment": "The builtin-baseline corresponds to 2025.08.27 Release",
  "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg.schema.json",
  "builtin-baseline": "120deac3062162151622ca4860575a33844ba10b",
  "dependencies": [
    "capnproto"
  ]
}
```

The code works. However, I believe the build system could do better and always define `CAPNP_EXECUTABLE` and `CAPNPC_CXX_EXECUTABLE` to verified paths.
See this branch: https://github.com/hebasto/libmultiprocess/commits/pr317/0731.alt/
Additionally, the code is organised into a function, and unnecessary quotes are removed.

Feel free to pick any part of that branch.

ryanofsky:
Thanks for the review! I dug into the debian bug a little more and think I found a simpler workaround. The details of the bug are in https://github.com/bitcoin-core/libmultiprocess/issues/328. I will push an update to https://github.com/bitcoin/bitcoin/pull/32387 including this workaround so we can be sure it still works for MSVC.

---


Updated 049dae17b5d0b59858e6ead1dec40e5e778f44d7 -> a26a08496b8bd9c43999e3636f9f53ec9022d124 ([`pr/win-cmake.1`](https://github.com/ryanofsky/libmultiprocess/commits/pr/win-cmake.1) -> [`pr/win-cmake.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/win-cmake.2), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/win-cmake.1..pr/win-cmake.2)) switching to simpler workaround for the debian packaging bug

hebasto (approved):
ACK a26a08496b8bd9c43999e3636f9f53ec9022d124, I've verified `CAPNP_EXECUTABLE` and `CAPNPC_CXX_EXECUTABLE` values on Ubuntu 24.04, Fedora 44 and Windows 11.

PR #318, opened 2026-07-22
workaround: MinGW thread_local use-after-free in ThreadContext

ryanofsky:
Work around mingw bug `thread_local` bug https://sourceforge.net/p/mingw-w64/bugs/527/ that causes failures in windows "test cross-built" CI jobs in https://github.com/bitcoin/bitcoin/pull/32387 by skipping `thread_local` destructors in mingw.

The workaround will result in resource leaks in mingw builds, that might be noticeable if a lot of threads are created and destroyed, but should not be a significant problem in practice.

This change is just a refactoring in other builds (including MSVC) not affected by this bug.

### DIFF
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -929,12 +929,7 @@ void ListenConnections(EventLoop& loop, SocketId fd, InitImpl& init, std::option
     });
 }
 
-extern thread_local ThreadContext g_thread_context; // NOLINT(bitcoin-nontrivial-threadlocal)
-// Silence nonstandard bitcoin tidy error "Variable with non-trivial destructor
-// cannot be thread_local" which should not be a problem on modern platforms, and
-// could lead to a small memory leak at worst on older ones.
-
-//! Return the current thread's ThreadContext.
+//! Return the current thread's ThreadContext, creating it on first use.
 //!
 //! Why per-thread state is needed at all: libmultiprocess has no control over
 //! which threads the C++ application uses to call ProxyClient methods after
@@ -951,7 +946,60 @@ extern thread_local ThreadContext g_thread_context; // NOLINT(bitcoin-nontrivial
 //! are freed. A thread_local object is the C++ mechanism that provides both
 //! of these: per-thread storage plus a destructor that runs at thread exit
 //! (the C equivalent would be a pthread key destructor). This is why
-//! ThreadContext is thread_local and why its destructor is nontrivial.
+//! ThreadContext is thread_local and why its destructor is nontrivial —
+//! which is what makes the MinGW bug below bite here.
+//!
+//! It is normally an ordinary function-local thread_local object destroyed at
+//! thread exit, except on MinGW, where it is a lazily-created heap object
+//! that is deliberately leaked (held by a trivially-destructible thread_local
+//! pointer, so no destructor is registered at thread exit at all). Note that
+//! on MinGW this skips more than freeing memory: releasing the remote Thread
+//! capabilities at client-thread exit is what lets the server destroy that
+//! thread's dedicated server threads, so with the workaround, server threads
+//! belonging to exited client threads are only freed when the connection
+//! closes (via the disconnect cleanup callbacks registered in SetThread /
+//! ProxyClient<Thread>::m_disconnect_cb).
+//!
+//! The MinGW workaround exists because MinGW-w64's emutls implementation can
+//! free the heap storage backing thread_local variables before the C++
+//! destructors registered via __cxa_thread_atexit run at thread exit. Both
+//! cleanups are implemented as winpthreads pthread key destructors (run by
+//! _pthread_cleanup_dest in winpthreads src/thread.c), and pthread key
+//! destructor order is unspecified, so gcc's emutls key destructor
+//! (emutls_destroy in libgcc/emutls.c, which frees each thread_local
+//! object's storage and the per-thread array) can run before the key
+//! destructor that invokes the C++ thread_local destructors. When that
+//! happens, any thread_local object with a nontrivial destructor is
+//! destroyed after its own memory was already freed. For this ThreadContext
+//! struct that meant ~ThreadContext walking the request_threads /
+//! callback_threads std::map trees through freed memory and double-freeing
+//! their nodes, corrupting the heap.
+//!
+//! This was observed in Bitcoin Core Windows CI as intermittent
+//! STATUS_HEAP_CORRUPTION (0xC0000374) exit code 3221226356 crashes of both
+//! bitcoin-cli (its EventLoop::loop thread exiting at IPC teardown) and
+//! bitcoin-node (per-request server threads created by
+//! ProxyServer<ThreadMap>::makeThread() exiting after client disconnects) in
+//! x86_64-w64-mingw32 (msvcrt) cross builds, but not in
+//! x86_64-w64-mingw32ucrt builds — presumably pthread-key creation-order
+//! luck, since the underlying bug is CRT-independent and unsafe on all
+//! mingw-posix toolchains. Under gdb the crash appears as SIGSEGV in
+//! ~ThreadContext inside _pthread_cleanup_dest with the object memory
+//! containing the 0xfeeefeee "freed heap" debug fill pattern, matching the
+//! signature described in the upstream reports below. (Without a debugger
+//! there is no fill pattern, so the destructor walks stale-but-intact
+//! pointers and double-frees them, which is why the corruption was
+//! intermittent and often detected later, e.g. in kj::AsyncIoContext
+//! teardown, rather than at the faulting site.)
+//!
+//! Known upstream reports of this mingw-w64 bug:
+//! - https://sourceforge.net/p/mingw-w64/bugs/527/ "thread_local stl object's destructor causing crash"
+//! - https://sourceforge.net/p/mingw-w64/bugs/727/ "Crash during destruction of thread_local objects"
+//! - https://sourceforge.net/p/mingw-w64/bugs/445/ "thread_local destructors broken with posix threading"
+//! - https://sourceforge.net/p/mingw-w64/bugs/859/ "thread_local destructors not called at thread exit"
+//! - https://github.com/msys2/MINGW-packages/issues/2519 "gcc thread_local destructor cause use after free"
+//! - https://www.mail-archive.com/mingw-w64-public@lists.sourceforge.net/msg17975.html
+//!   (proposed fix discussion: "fix __cxa_thread_atexit destructors on GCC")
 ThreadContext& CurrentThread();
 
 } // namespace mp
--- a/src/mp/proxy.cpp
+++ b/src/mp/proxy.cpp
@@ -40,11 +40,36 @@
 
 namespace mp {
 
-thread_local ThreadContext g_thread_context; // NOLINT(bitcoin-nontrivial-threadlocal)
-
 ThreadContext& CurrentThread()
 {
-    return g_thread_context;
+#ifdef __MINGW32__
+    // On MinGW only, deliberately leak a heap object instead of using a plain
+    // thread_local variable, because MinGW-w64's emutls implementation can
+    // free the storage backing thread_local variables before C++ destructors
+    // registered by __cxa_thread_atexit run at thread exit (pthread key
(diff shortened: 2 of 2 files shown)


ryanofsky:
Updated 1243dd3309692938d6c71473da180265fe26e8d0 -> 1500701d98a72067874cb7e5b72a9c40b03fe722 ([`pr/win-tls.1`](https://github.com/ryanofsky/libmultiprocess/commits/pr/win-tls.1) -> [`pr/win-tls.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/win-tls.2), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/win-tls.1..pr/win-tls.2)) renaming `GThreadContext` to `CurrentThread`

xyzconstant:
In commit 1500701 _"workaround: MinGW thread_local use-after-free in ThreadContext"_

nit: The commit body still mentions `GThreadContext()` instead of `CurrentThread`

xyzconstant:
Code review ACK 1500701

Compiled and ran tests on macOS. Tried to reproduce the heap corruption with a `mingw-w64` cross-build run under Wine following some of the instructions in the bug report links but didn't have any success (maybe luck?).

My understanding: MinGW's machinery doesn't guarantee a safe cleanup order for `thread_local` variables. The idea is to skip the `__cxa_thread_atexit` call entirely (which registers `~ThreadContext`, a non-trivial destructor to run at thread exit) so its scheduled destructor doesn't run on already-freed memory and thereby avoid heap corruption. The changes in this PR implement that by modifying the `g_thread_context` variable (now returned by `CurrentThread()`) from a value of type `ThreadContext` to a pointer of type `ThreadContext *` on MinGW. This results in only the `emutls` cleanup running. This frees the pointer's storage at destruction time, and the `context` object is, as a result, leaked.

The cost of this leakage is that exiting client threads won't release the `Thread` capability, which means the server threads aren't closed until the connection is closed (closing the connection destroys those OS threads).

Other target builds remain unaffected. The documentation is well-written, and it's clear to any developer why this workaround exists. This PR looks good to me.

include/mp/proxy-io.h
```diff
@@ -929,10 +929,78 @@ void ListenConnections(EventLoop& loop, SocketId fd, InitImpl& init, std::option
...
+//! object's storage and the per-thread array) can run before the key
+//! destructor that invokes the C++ thread_local destructors. When that
+//! happens, any thread_local object with a nontrivial destructor is
+//! destroyed after its own memory was already freed. For this ThreadContext
+//! struct that meant ~ThreadContext walking the request_threads /
+//! callback_threads std::map trees through freed memory and double-freeing
+//! their nodes, corrupting the heap.
+//!
+//! This was observed in Bitcoin Core Windows CI as intermittent
+//! STATUS_HEAP_CORRUPTION (0xC0000374) exit code 3221226356 crashes of both
```
hebasto, line 979: Is this specific to MSVCRT, or does it occur with UCRT as well?

Sjors:
Wrapping `g_thread_context` in `CurrentThread()`, and then dropping it entirely, is nice in general. Maybe make it a prep refactor commit(s), so the windows change is more focussed. See [sjors/2026/07/mingw](https://github.com/Sjors/libmultiprocess/commits/2026/07/mingw/).

My agent also couldn't reproduce the original issue on Linux with Wine, initially not on Windows 11 either, but I'll keep trying.

I didn't verify the Windows-specific rationale, but the change itself seems simple enough is well documented.

Sjors:
ACK 1500701d98a72067874cb7e5b72a9c40b03fe722

It figured it out: https://gist.github.com/Sjors/4b70c60a5d2989bf2ba4adb408d2554c

maflcko:
Seems fine, but given that this was already fixed in https://github.com/mingw-w64/mingw-w64/commit/8e06daa36dfcea4bb491acf4b350658f40738f02  / 13.0.0 , I wonder if it would be easier to just build that version (or later).

This is with the background that the Ubuntu/Debian packages are already broken and possibly unmaintained, so we may want to consider self-building anyway? Ref: https://github.com/bitcoin/bitcoin/pull/33593#issuecomment-3748883932

Either self-building directly, or using something like https://github.com/0xB10C/bitcoind-gunix/blob/v31.1/nix/win64/toolchain.nix ? Edit: mingw in guix was bumped to v13 in https://github.com/bitcoin/bitcoin/commit/31eb46f054f8cbd62b3ac44fd98a61f6dda34879

include/mp/proxy-io.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/318#discussion_r3658249746

> Is this specific to MSVCRT, or does it occur with UCRT as well?

I don't think it has anything do to with the C runtime, although agent noted it did not see this error under UCRT builds for some reason (see "since the underlying bug is CRT-independent" comment below)

This error is pretty easy to reproduce by disabling the `#ifdef __MINGW32__` workaround in https://github.com/bitcoin/bitcoin/pull/32387 and seeing the cross-compiled CI job in `interface_ipc_cli.py` and `interface_ipc_mining.py` tests. You can see the ci errors looking at `win-work` branch pushes from Jul 22 in  `https://github.com/ryanofsky/bitcoin/actions like https://github.com/ryanofsky/bitcoin/actions/runs/29931590196/job/88964021272

ryanofsky (commented):
re: https://github.com/bitcoin-core/libmultiprocess/pull/318#issuecomment-5108942086

> mingw in guix was bumped to v13 in [bitcoin/bitcoin@31eb46f](https://github.com/bitcoin/bitcoin/commit/31eb46f054f8cbd62b3ac44fd98a61f6dda34879)

Thanks for these details. Would have to look into this more but would be nice if this is already fixed in guix builds and the workaround is unnecessary there.

I think we still might want this code change (maybe narrowed to depend on mingw version or whether the fix is present) if it is easier than self-building mingw in CI. A motivation for this change is getting
`interface_ipc_cli.py` and `interface_ipc_mining.py` tests to pass in https://github.com/bitcoin/bitcoin/pull/32387 cross-compiled CI jobs

---


Rebased 1500701d98a72067874cb7e5b72a9c40b03fe722 -> 608d09a8fec7f65cbe3a2c4b2afec69dfca32c15 ([`pr/win-tls.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/win-tls.2) -> [`pr/win-tls.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/win-tls.3), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/win-tls.2-rebase..pr/win-tls.3)) due to silent conflicts with #323

maflcko:
> Thanks for these details. Would have to look into this more but would be nice if this is already fixed in guix builds and the workaround is unnecessary there.

It is just a guess that this is already fixed, since I don't have Windows to test. One could check by using `ubuntu:resolute` in the CI config, maybe?

The general idea is that we are having problems cross-compiling with the apt system packages, so using guix or nix instead could help with that, and possibly also help with making the CI closer to the release bins, and possibly even align the cross-compile docs to produce bins identical to the ones from contrib/guix.

So trying to use a mingw from nix (possibly in combination with a apt system gcc, or so) would be interesting and could possibly fix a bunch of issues (or introduce more issues, heh).

maflcko:
> I think we still might want this code change (maybe narrowed to depend on mingw version or whether the fix is present) if it is easier than self-building mingw in CI. A motivation for this change is getting
> `interface_ipc_cli.py` and `interface_ipc_mining.py` tests to pass in [bitcoin/bitcoin#32387](https://github.com/bitcoin/bitcoin/pull/32387) cross-compiled CI jobs

If you want, you can merge https://github.com/bitcoin/bitcoin/pull/35877 (mingw 13) into 32387, to see if it passes CI.

ryanofsky:
re:  https://github.com/bitcoin-core/libmultiprocess/pull/318#issuecomment-5180729878

> If you want, you can merge [bitcoin/bitcoin#35877](https://github.com/bitcoin/bitcoin/pull/35877) (mingw 13) into 32387, to see if it passes CI.

Thanks I confirmed bitcoin/bitcoin#35877 works and fixes the CI jobs. I'll convert this PR to a draft assuming https://github.com/bitcoin/bitcoin/pull/35877 can be merged and the workaround leaking memory with mingw is no longer necessary.

I still think some change to libmultiprocess is needed to trigger an error if it is built with older versions of mingw that would be unstable without the bugfix. So I will keep this PR open, but it should be less important now

Separately it might also make sense to disable the [nontrivial-threadlocal.h](https://github.com/bitcoin/bitcoin/blob/master/contrib/devtools/bitcoin-tidy/nontrivial-threadlocal) check at some point if the underlying mingw bug is fixed.

#### Test 1: Reverting current workaround

I tested this with 3 pushes. The first push reverted the workaround in this PR and showed the cross-built msvcrt and ucrt jobs failing:

https://github.com/ryanofsky/bitcoin/commits/pr/ipc-win.27-revert-mingw
https://github.com/bitcoin/bitcoin/actions/runs/30943797734/job/92119010339?pr=32387
https://github.com/bitcoin/bitcoin/actions/runs/30943797734/job/92119010280?pr=32387

 It was actually a little surprising to see the ucrt job failing since it didn't look like UCRT jobs were failing [previously](https://github.com/bitcoin-core/libmultiprocess/pull/318#discussion_r3684492070), but it does make sense for them to fail.

The msvcrt job showed both `interface_ipc_cli.py` and `interface_ipc_mining.py` failing while the ucrt job showed only `interface_ipc_cli.py` failing. In the mining test the node failed with 3221226356 exit code which is the the heap corruption error described by Claude. And in the cli tests `bitcoin-cli` failed with the same 3221226356 exit code.

#### Test 2: Applying fix

The next push added cherry-picked commits from bitcoin/bitcoin#35877 but these failed to compile with `error: ‘pthread_self’ was not declared in this scope` caused by `HAVE_PTHREAD_GETNAME_NP` check being true for the nix build of mingw when it was false in the debian build.

https://github.com/ryanofsky/bitcoin/commits/pr/pr/ipc-win.27-nix-mingw
https://github.com/bitcoin/bitcoin/actions/runs/30952272732/job/92139206608?pr=32387
https://github.com/bitcoin/bitcoin/actions/runs/30952272732/job/92139206619?pr=32387

#### Test 3: Verifying fix

The last push fixed the compile error and confirmed the fixes with all jobs now passing:

https://github.com/ryanofsky/bitcoin/commits/pr/pr/ipc-win.27-nix-mingw2
https://github.com/bitcoin/bitcoin/actions/runs/30959207156/job/92159561789?pr=32387
https://github.com/bitcoin/bitcoin/actions/runs/30959207156/job/92159561813?pr=32387

maflcko:
> Separately it might also make sense to disable the [nontrivial-threadlocal.h](https://github.com/bitcoin/bitcoin/blob/master/contrib/devtools/bitcoin-tidy/nontrivial-threadlocal) check at some point if the underlying mingw bug is fixed.

Maybe, but see https://github.com/bitcoin/bitcoin/pull/30095#discussion_r1608289449

ryanofsky:
Closing the PR, since the use-after-free bug is fixed in mingw-w64 13.0.0, which includes https://github.com/mingw-w64/mingw-w64/commit/8e06daa36dfcea4bb491acf4b350658f40738f02 and our binaries are using since https://github.com/bitcoin/bitcoin/commit/31eb46f054f8cbd62b3ac44fd98a61f6dda34879. Marco's change https://github.com/bitcoin/bitcoin/commit/fa8762da626a592585f936b9398a100e3adfd923 additionally fixes the problem in CI.

Someone building from source with an older mingw could still hit this, so it could be reasonable to add a check like `#if defined(__MINGW32__) && __MINGW64_VERSION_MAJOR < 13` with a `#warning` or `#error` to suggest an upgrade.

The bug might also still occur if libstdc++ is natively compiled instead of cross compiled. A natively compiled version could bypass the fix in https://github.com/mingw-w64/mingw-w64/commit/8e06daa36dfcea4bb491acf4b350658f40738f02 because it would detect the `__cxa_thread_atexit` function provided by  `mingw-w64-crt` and use that instead of pthreads to invoke destructors and still call them after storage was freed.

The bug can also be avoided by using GCC 16, which supports native thread-local storage on Windows (msys2/MINGW-packages#29272), or Clang.

PR #330, opened 2026-08-02
ci: Compile with minimum supported g++ in olddeps

maflcko:
Looks like there is no CI coverage for the minimum supported g++ version?

### DIFF
--- a/ci/configs/olddeps.bash
+++ b/ci/configs/olddeps.bash
@@ -1,5 +1,5 @@
 CI_DESC="CI job using old Cap'n Proto and cmake versions"
 CI_DIR=build-olddeps
 export CXXFLAGS="-Werror -Wall -Wextra -Wpedantic -Wno-unused-parameter -Wno-error=array-bounds"
-NIX_ARGS=(--argstr capnprotoVersion "0.9.2" --argstr cmakeVersion "3.12.4")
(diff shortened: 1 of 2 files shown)


hebasto:
> ci: Compile with minimum supported g++-11 in olddeps

It [seems](https://github.com/bitcoin-core/libmultiprocess/pull/321#discussion_r3699662514) g++10 is the minimum supported.

ryanofsky (approved):
Code review ACK edf6343562450eec0a212d10d96fb9e39192933f. Would maybe change title to "Compile with older GCC in olddeps" because it's not clear what the real minimum is. (Or if it is clear it would be nice to document).

Also this is only testing libmultiprocess build with the older GCC, so using older GCC for capnproto as well could be more realistic, but this PR is probably the simplest change

maflcko:
> Would maybe change title to "Compile with older GCC in olddeps" because it's not clear what the real minimum is. (Or if it is clear it would be nice to document).

Happy to do this, but not sure how. Should I cherry-pick https://github.com/bitcoin/bitcoin/pull/34580  into this repo ?



> Also this is only testing libmultiprocess build with the older GCC, so using older GCC for capnproto as well could be more realistic, but this PR is probably the simplest change

This seems like something for upstream to do. Recall they already have `Linux-old (g++-9)` config.

PR #331, opened 2026-08-02
Remove code for Cap'n Proto versions before 0.9

hebasto:
The removed code has been dead since db4f9a3d7395d435c484f5735a787f0798080c52, which bumped the minimum required Cap'n Proto version to 0.9.


maflcko:
lgtm ACK 788f17a850942448f84e543aeefda3ac6c577327

shell.nix
```diff
@@ -17,10 +17,6 @@ let
     };
   };
   capnprotoHashes = {
-    "0.7.0" = "sha256-Y/7dUOQPDHjniuKNRw3j8dG1NI9f/aRWpf8V0WzV9k8=";
-    "0.7.1" = "sha256-3cBpVmpvCXyqPUXDp12vCFCk32ZXWpkdOliNH37UwWE=";
-    "0.8.0" = "sha256-rfiqN83begjJ9eYjtr21/tk1GJBjmeVfa3C3dZBJ93w=";
-    "0.8.1" = "sha256-OZqNVYdyszro5rIe+w6YN00g6y8U/1b8dKYc214q/2o=";
```
ryanofsky, line 23: In commit "Remove code for Cap'n Proto versions before 0.9" (788f17a850942448f84e543aeefda3ac6c577327)

Other changes in this commit all make sense, but I do think it'd be a little nicer to keep these hashes as they are convenient for testing. The dependency on 0.9.0 is weak and only due to a `getFd` call which I think we may want to drop anyway, because capnproto has long had better mechanisms for communicating across threads than writing to file descriptors: https://github.com/bitcoin-core/libmultiprocess/pull/274#discussion_r3602426663

ryanofsky (approved):
Code review ACK 788f17a850942448f84e543aeefda3ac6c577327. It's especially nice to clean up the `cmake/compat_config.cmake` file

PR #332, opened 2026-08-04
ci: add -Wextra-semi to default config

fanquake:
Follow-up to https://github.com/bitcoin-core/libmultiprocess/commit/bd508311b5617bd8d31b875d8781392f2a6ad7be.

### DIFF
--- a/ci/configs/default.bash
+++ b/ci/configs/default.bash
@@ -1,5 +1,5 @@
 CI_DESC="CI job using default libraries and tools, and running IWYU"
 CI_DIR=build-default
-export CXXFLAGS="-Werror -Wall -Wextra -Wpedantic -Wno-unused-parameter"
+export CXXFLAGS="-Werror -Wall -Wextra -Wpedantic -Wno-unused-parameter -Wextra-semi"
 CMAKE_ARGS=(-DMP_ENABLE_IWYU=ON)
 BUILD_ARGS=(-k)


test/mp/test/test.cpp
```diff
@@ -569,12 +569,12 @@ KJ_TEST("Call async IPC method without thread or pool errors correctly")
...
                     error_thrown = true;
                     KJ_EXPECT(std::string_view{e.getDescription().cStr()}.find(
-                        "no thread specified and no pool configured") != std::string_view::npos);
+                        "no thread specified and no pool configured") != std::string_view::npos)
                     done.set_value();
                 }));
     });
     done.get_future().get();
-    KJ_EXPECT(error_thrown);
+    KJ_EXPECT(error_thrown)
```
maflcko: hmm, for macros, I generally prefer if they *enforce* a trailing semicolon.

Though, it may be too late to fix this now?

maflcko (commented):
Seems fine, but I am -0

fanquake:
Yea. I don't have too much of an opinion, but it wasn't clear to me why a PR labelled "CI" (#321) was also making random source code changes (bd508311b5617bd8d31b875d8781392f2a6ad7be) (although it seems like that PR was generally confusing (https://github.com/bitcoin-core/libmultiprocess/pull/321#issuecomment-5171132506)), so thought we could at least make this consistent.

test/mp/test/test.cpp
hebasto: Then perhaps omit `-Wextra-semi-stmt`?

maflcko:
I think 321 was dropping it after a function def, this one is dropping it after statements.

hebasto:
> ... random source code changes ([bd50831](https://github.com/bitcoin-core/libmultiprocess/commit/bd508311b5617bd8d31b875d8781392f2a6ad7be))...

See https://github.com/bitcoin-core/libmultiprocess/pull/321#discussion_r3699662514.

fanquake:
> See https://github.com/bitcoin-core/libmultiprocess/pull/321#discussion_r3699662514.

Thanks. It would be good to write commit messages that contain information for other contributors (which then persists over time), so they don't have to search through GitHub comments to try and find context/explanations.

fanquake:
Changed to just add `-Wextra-semi`, to prevent some re-introduction.

hebasto (approved):
ACK 49834b2609e3e4dca1b9cba36ee152552e221039.

ci/configs/default.bash
```diff
@@ -1,5 +1,5 @@
 CI_DESC="CI job using default libraries and tools, and running IWYU"
 CI_DIR=build-default
-export CXXFLAGS="-Werror -Wall -Wextra -Wpedantic -Wno-unused-parameter"
+export CXXFLAGS="-Werror -Wall -Wextra -Wpedantic -Wno-unused-parameter -Wextra-semi"
```
ryanofsky, line 3: In commit "ci: add -Wextra-semi to default config" (49834b2609e3e4dca1b9cba36ee152552e221039)

This could be good to add to `llvm.bash` as well in case gcc/clang behaviors are different

ryanofsky (approved):
Code review ACK 49834b2609e3e4dca1b9cba36ee152552e221039. Tested locally and confirmed this catches both extra semicolons with bd508311b5617bd8d31b875d8781392f2a6ad7be reverted

ci/configs/default.bash
fanquake: Done in #341.

PR #334, opened 2026-08-07
ci: Set CMAKE_BUILD_PARALLEL_LEVEL to enable parallelism by default

ryanofsky:
This benefits `default` and `olddeps` CI jobs which use make and specify `-j` in BUILD_ARGS.

Jobs which specify explicit `-j` options should be unaffected and ninja jobs should be unaffected because ninja uses the same default internally.


maflcko:
lgtm ACK ce51d737255df9e420d5a3dc09e1615b5d12f2ff

hebasto (approved):
ACK ce51d737255df9e420d5a3dc09e1615b5d12f2ff.

PR #335, opened 2026-08-07
proxy-io.h: Add Connection `disconnect` and `waitDrained` methods

ryanofsky:
**Note: This is based on #361.** Initial commits should be reviewed in that PR.

---

Add `Connection` class `disconnect` and `waitDrained` methods to provide more flexibility when forcibly disconnecting from remote clients or servers.

Without these methods, the only way to forcibly close IPC connections is to delete `Connection` objects. This works but is not ideal because once a `Connection` object is gone, it is difficult to track state still associated with the connection, particularly:

- `ProxyServer` objects that may still be alive because they are executing asynchronous requests made before the disconnect. Without a way to track these objects, there is no generic way to wait for requests to finish existing after disconnecting. So individual IPC interfaces like the Bitcoin mining interface would need to implement custom synchronization to avoid race conditions during shutdown. Followup PR https://github.com/bitcoin/bitcoin/pull/35932 builds on this PR, calling the new `waitDrained` method introduced here to avoid IPC mining crashes on Bitcoin core shutdown without needing to change the mining code. A unit test is added here simulating these mining crashes.

- `ProxyClient` objects that contain pointers to `Connection` objects. Currently `ProxyClient` object need to register cleanup handlers with `Connection` objects to deal with Connections being deleted, which consumes memory and complicates `ProxyClient` shutdown logic. After this change, a followup PR will drop the cleanup handlers so `Connection` objects no longer need to track lists of `ProxyClient` objects associated with them. This is implemented in #336.


xyzconstant:
Concept ACK

enirox001:
CI seems upset?

```
/home/runner/work/libmultiprocess/libmultiprocess/src/mp/proxy.cpp should add these lines:
#include <capnp/rpc-twoparty.h>  // for TwoPartyVatNetwork
```

ryanofsky:
Updated 39ed2ca8f068a065403afa89bbf1d14f75773720 -> a40189f5bbed89f09e5d63ce8377dc90574983f6 ([`pr/keepconn.1`](https://github.com/ryanofsky/libmultiprocess/commits/pr/keepconn.1) -> [`pr/keepconn.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/keepconn.2), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/keepconn.1..pr/keepconn.2)) fixing iwyu errors and olddeps failure due to incompatibility with old capnproto versions which lack `kj:::TaskSet::clear` method https://github.com/bitcoin-core/libmultiprocess/actions/runs/31189931644/job/92903876132?pr=335

Added 1 commits a40189f5bbed89f09e5d63ce8377dc90574983f6 -> 11929f1b9b39aba8ac636e66bd727689bb2b7a95 ([`pr/keepconn.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/keepconn.2) -> [`pr/keepconn.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/keepconn.3), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/keepconn.2...pr/keepconn.3)) to fix pre-existing ~ThreadContext() bug exposed by combination of new test in this PR and the onDisconnect handler added in #298 commit bb473690c97ceed78a482d6a97b480cec4a63190 https://github.com/bitcoin-core/libmultiprocess/actions/runs/31662352370/job/94329590873?pr=335

Updated 11929f1b9b39aba8ac636e66bd727689bb2b7a95 -> 901a090da03ed6b04d46a4040ed16fdeb238e7fd ([`pr/keepconn.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/keepconn.3) -> [`pr/keepconn.4`](https://github.com/ryanofsky/libmultiprocess/commits/pr/keepconn.4), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/keepconn.3..pr/keepconn.4)) to fix iwyu error https://github.com/bitcoin-core/libmultiprocess/actions/runs/31666872134/job/94343269972?pr=335

src/mp/proxy.cpp
```diff
@@ -113,6 +114,25 @@ Connection::~Connection() noexcept(false)
...
     assert(std::this_thread::get_id() == m_loop->m_thread_id);
+    disconnect();
+}
+
+void Connection::disconnect()
+{
+    // Disconnecting triggers I/O and tears down capnp state, so it must run on
+    // the event loop thread, like the destructor.
+    assert(std::this_thread::get_id() == m_loop->m_thread_id);
+    if (m_disconnected) return;
```
enirox001: In commit https://github.com/bitcoin-core/libmultiprocess/pull/335/changes/39cc757fba71871b154d915b02cf3967fe73fe0d: _ipc: add Connection::disconnect() separating teardown from destruction_

`disconnect()` sets `m_disconnected = true`; later on we clean everything up, so I am unsure, but if there was a scenario where one of the cleanups threw, it would not complete the rest. This might not be a problem, but another call to disconnect() would be a no-op. 

I do not think all the operations after this can cause this to throw and lead to this, but `shutdownWrite()` might if it throws an exception other than the ones mentioned. 

A simple fix is to set the `m_disconnected = true` only after all teardown that must run has completed.

```diff --git a/src/mp/proxy.cpp b/src/mp/proxy.cpp
index 0aaa58a..8b9f458 100644
--- a/src/mp/proxy.cpp
+++ b/src/mp/proxy.cpp
@@ -124,7 +124,6 @@ void Connection::disconnect()
     // the event loop thread, like the destructor.
     assert(std::this_thread::get_id() == m_loop->m_thread_id);
     if (m_disconnected) return;
-    m_disconnected = true;

     // Cancel pending onDisconnect handlers first. Severing the connection
     // below completes m_network.onDisconnect() promises, and the registered
@@ -253,6 +252,8 @@ void Connection::disconnect()
     // stream.
     m_network.reset();
     m_stream = nullptr;
+
+    m_disconnected = true;
 }

 void Connection::waitDrained()
```

or a better solution that make sure the the cleanup happens even if `shutdownWrite` fails?

include/mp/proxy-io.h
```diff
@@ -473,7 +486,7 @@ class Connection
         // handler fires, do not call the function f right away, instead add it
         // to the EventLoop TaskSet to avoid "Promise callback destroyed itself"
         // error in the typical case where f deletes this Connection object.
-        m_on_disconnect.add(m_network.onDisconnect().then(
+        m_on_disconnect->add(m_network->onDisconnect().then(
```
enirox001: In https://github.com/bitcoin-core/libmultiprocess/pull/335/changes/39cc757fba71871b154d915b02cf3967fe73fe0d: _ipc: add Connection::disconnect() separating teardown from destruction_

Before this PR, when a remote side disconnected, libmultiprocess had callbacks that would eventually remove the `Connection`. It does not necessarily call the remove operation immediately; it can schedule it into another task set. The new `disconnect()` wants different behaviour. such that it will disconnect and then call `waitDrained` later on. So it tries to reset the `m_on_disconnect` callbacks. 

But if the callback has already progressed one step further before reset happens, this violates the goal of this new system.

In https://github.com/bitcoin-core/libmultiprocess/pull/336/changes/aa49a11c028f92aafa3867a59f9eb52cf9bdde6d this is made to use a `weak_ptr`, but I wonder if we should move those changes to this pr instead? Or rather, a small cancellation guard could be added to this PR such that it keeps the existing changes focused while preventing the potential regression. 

A minimal change adding a weak cancelation token that has moved into the event loop queue.

```diff --git a/include/mp/proxy-io.h b/include/mp/proxy-io.h
index 1f77b26..d817eb6 100644
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -576,8 +576,18 @@ public:
         // handler fires, do not call the function f right away, instead add it
         // to the EventLoop TaskSet to avoid "Promise callback destroyed itself"
         // error in the typical case where f deletes this Connection object.
+        const std::weak_ptr<void> guard{m_on_disconnect_guard};
         m_on_disconnect->add(m_network->onDisconnect().then(
-            [f = std::forward<F>(f), this]() mutable { m_loop->m_task_set->add(kj::evalLater(kj::mv(f))); }));
+            [f = std::forward<F>(f), guard, this]() mutable {
+                m_loop->m_task_set->add(kj::evalLater(
+                    [f = kj::mv(f), guard]() mutable {
+                        // The connection-owned TaskSet may have already handed
+                        // this callback to the event-loop TaskSet by the time
+                        // disconnect() cancels it. Only run it if the
+                        // connection has not been disconnected in between.
+                        if (guard.lock()) f();
+                    }));
+            }));
     }

     EventLoopRef m_loop;
@@ -587,6 +597,10 @@ public:
     //! disconnections, if the connection is closed locally first by deleting
     //! this Connection object.
     std::optional<kj::TaskSet> m_on_disconnect{std::in_place, m_error_handler};
+    //! Lifetime token checked by onDisconnect handlers after they are handed
+    //! off to the EventLoop TaskSet. Reset by disconnect() so a handler already
+    //! queued there cannot run after local teardown.
+    std::shared_ptr<void> m_on_disconnect_guard{std::make_shared<char>()};
     //! Wrapped in std::optional so disconnect() can destroy it (and m_stream
     //! below) to sever the transport while this object stays alive. Closing
     //! the stream is what makes the peer observe the disconnect: it reads EOF
diff --git a/src/mp/proxy.cpp b/src/mp/proxy.cpp
index 0aaa58a..06063f8 100644
--- a/src/mp/proxy.cpp
+++ b/src/mp/proxy.cpp
@@ -133,6 +133,7 @@ void Connection::disconnect()
     // harmful when disconnect() is called separately by code that keeps using
     // the object afterwards (e.g. code waiting for in-flight calls to finish
     // before destroying it).
+    m_on_disconnect_guard.reset();
     m_on_disconnect.reset();

     // Try to cancel any calls that may be executing.
```
This change closes the gap where resetting `m_on_disconnect` was too late because the callback had already moved into the EventLoop task set

enirox001: In https://github.com/bitcoin-core/libmultiprocess/pull/335/changes/39cc757fba71871b154d915b02cf3967fe73fe0d:  _ipc: add Connection::disconnect() separating teardown from destruction_

The listener now keeps a counter of the active connections added in https://github.com/bitcoin-core/libmultiprocess/pull/269/changes/39a10ce8958ef9d49d0ec82acd3a6d506d5c8fee. When it is full, it stops accepting new connections, and when a client disconnects, a callback decreases the counter, and the listener can start accepting again. 

But when the server calls `disconnect()` it cancels that callback. The connection closes, but the counter does not change, so the listener might think it is full and never accept another connection

Added this change to so that the listener count can be updated for every disconnect, while automatic deletion happens only for remote disconnects

```diff --git a/include/mp/proxy-io.h b/include/mp/proxy-io.h
index 1f77b26..30627ec 100644
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -1016,10 +1016,12 @@ void _Serve(EventLoop& loop, kj::Own<kj::AsyncIoStream>&& stream, InitImpl& init
     auto it = loop.m_incoming_connections.begin();
     MP_LOG(loop, Log::Info) << "IPC server: socket connected.";
     if (loop.testing_hook_connected) loop.testing_hook_connected();
-    it->onDisconnect([&loop, it, on_disconnect = std::forward<OnDisconnect>(on_disconnect)]() mutable {
+    it->addSyncCleanup([on_disconnect = std::forward<OnDisconnect>(on_disconnect)]() mutable {
+        on_disconnect();
+    });
+    it->onDisconnect([&loop, it]() mutable {
         MP_LOG(loop, Log::Info) << "IPC server: socket disconnected.";
         loop.m_incoming_connections.erase(it);
-        on_disconnect();
         if (loop.testing_hook_disconnected) loop.testing_hook_disconnected();
     });
 }
```
This test could also be added to verify the above behaviour

```diff --git a/test/mp/test/listen_tests.cpp b/test/mp/test/listen_tests.cpp
index a9d4dca..240af3f 100644
--- a/test/mp/test/listen_tests.cpp
+++ b/test/mp/test/listen_tests.cpp
@@ -265,6 +265,29 @@ KJ_TEST("ListenConnections enforces a local connection limit")
     KJ_EXPECT(client3->client->add(3, 4) == 7);
 }

+KJ_TEST("ListenConnections resumes after a local disconnect")
+{
+    ListenSetup server(/*max_connections=*/1);
+
+    auto client1 = std::make_unique<ClientSetup>(server.listener.MakeConnectedSocket());
+    server.WaitForConnectedCount(1);
+    KJ_EXPECT(client1->client->add(1, 2) == 3);
+
+    auto client2 = std::make_unique<ClientSetup>(server.listener.MakeConnectedSocket());
+    (**server.m_loop_ref).sync([] {});
+    KJ_EXPECT(server.ConnectedCount() == 1);
+
+    EventLoop& loop{**server.m_loop_ref};
+    loop.sync([&] {
+        KJ_REQUIRE(loop.m_incoming_connections.size() == 1);
+        loop.m_incoming_connections.front().disconnect();
+        loop.m_incoming_connections.pop_front();
+    });
+
+    server.WaitForConnectedCount(2);
+    KJ_EXPECT(client2->client->add(2, 3) == 5);
+}
+
 KJ_TEST("ListenConnections accepts multiple connections")
 {
     // With max-connections=2, two clients should be accepted and usable at the
```

include/mp/proxy-io.h
```diff
@@ -442,22 +442,35 @@ class Connection
 public:
     Connection(EventLoop& loop, kj::Own<kj::AsyncIoStream>&& stream_)
         : m_loop(loop), m_stream(kj::mv(stream_)),
-          m_network(*m_stream, ::capnp::rpc::twoparty::Side::CLIENT, ::capnp::ReaderOptions()),
-          m_rpc_system(::capnp::makeRpcClient(m_network)) {}
+          m_network(std::in_place, *m_stream, ::capnp::rpc::twoparty::Side::CLIENT, ::capnp::ReaderOptions()),
```
enirox001, line 538: In commit https://github.com/bitcoin-core/libmultiprocess/pull/335/changes/39cc757fba71871b154d915b02cf3967fe73fe0d: _ipc: add Connection::disconnect() separating teardown from destruction_

Connection can now remain alive after it has been disconnected, but the class does not define which methods are safe to call afterwards. 

Previously, disconnection meant destroying the whole object, but now the connection object still exists. This is needed so callers can use methods such as `waitDrained`, but other methods still behave as if the connection is active.

Could some documentation, assertion, or runtime check be helpful for this?

src/mp/proxy.cpp
```diff
@@ -254,6 +254,15 @@ void Connection::disconnect()
     m_stream = nullptr;
 }
 
+void Connection::waitDrained()
+{
+    // Blocking the event loop thread here would deadlock: in-flight call
+    // bodies sync() back to the event loop to deliver their results, and
+    // server objects are destroyed on the event loop thread.
+    assert(std::this_thread::get_id() != m_loop->m_thread_id);
```
enirox001: In https://github.com/bitcoin-core/libmultiprocess/pull/335/changes/631d8d9d439e22ea782d469c5ce4c75e2ee3d58c: _ipc: add Connection::waitDrained() to wait for in-flight server calls_

The documentation for the `waitDrained` method in `proxy-io.h` says it is meant to be called after `disconnect()` call, but does nothing to enforce it, i think we can assert that the `disconnect` method has been called before calling `waitDrained` as such

```diff --git a/src/mp/proxy.cpp b/src/mp/proxy.cpp
index 0aaa58a..6a318b0 100644
--- a/src/mp/proxy.cpp
+++ b/src/mp/proxy.cpp
@@ -261,6 +261,7 @@ void Connection::waitDrained()
     // bodies sync() back to the event loop to deliver their results, and
     // server objects are destroyed on the event loop thread.
     assert(std::this_thread::get_id() != m_loop->m_thread_id);
+    assert(m_disconnected);
     m_server_objects->wait();
 }
```

include/mp/proxy-io.h
```diff
@@ -471,6 +537,26 @@ class Connection
...
+    //! not garbage collected, so the count would not drain. Must NOT be called
+    //! from the event loop thread: in-flight call bodies need the event loop
+    //! to deliver their results before their server objects are destroyed, so
+    //! blocking the loop here would deadlock.
+    //!
+    //! This lets shutdown code ensure no IPC call body is still executing (and
+    //! dereferencing application state that is about to be freed) after
+    //! incoming connections are disconnected. See Ipc::disconnectIncoming and
+    //! https://github.com/bitcoin/bitcoin/issues/35845.
+    void waitDrained();
```
enirox001: In https://github.com/bitcoin-core/libmultiprocess/pull/335/changes/631d8d9d439e22ea782d469c5ce4c75e2ee3d58c: _ipc: add Connection::waitDrained() to wait for in-flight server calls_

nit: 
The name `waitDrained` could be clearer if it is called `waitServerCallsDrained`? or at least document its exact scope a bit more clearly

test/mp/test/test.cpp
```diff
@@ -425,6 +425,77 @@ KJ_TEST("Calling async IPC method, with server disconnect after cleanup")
...
+
+    // The FooInterface server object is the connection's only counted server
+    // object, and its call body is executing.
+    KJ_EXPECT(connection->pendingServerObjects() == 1);
+
+    // Disconnect. This cancels the call's promise (the client above sees the
+    // disconnect error), but the body is still blocked on the worker thread,
+    // so its server object must still be alive.
+    foo->m_context.loop->sync([&] { connection->disconnect(); });
+    KJ_EXPECT(connection->pendingServerObjects() == 1);
```
enirox001, line 595: In commit https://github.com/bitcoin-core/libmultiprocess/pull/335/changes/092d1db8fe7531e220c90bab0c4a78c83bdd978c: _test: cover draining in-flight server call after disconnect_

I think using an exact count of one is a bit brittle; the test only needs to establish that something remains in flight. This would be less implementation-specific

```diff --git a/test/mp/test/test.cpp b/test/mp/test/test.cpp
index 5bccb86..da47fde 100644
--- a/test/mp/test/test.cpp
+++ b/test/mp/test/test.cpp
@@ -463,13 +463,13 @@ KJ_TEST("Waiting for in-flight server call to finish after disconnect")

     // The FooInterface server object is the connection's only counted server
     // object, and its call body is executing.
-    KJ_EXPECT(connection->pendingServerObjects() == 1);
+    KJ_EXPECT(connection->pendingServerObjects() > 0);

     // Disconnect. This cancels the call's promise (the client above sees the
     // disconnect error), but the body is still blocked on the worker thread,
     // so its server object must still be alive.
     foo->m_context.loop->sync([&] { connection->disconnect(); });
-    KJ_EXPECT(connection->pendingServerObjects() == 1);
+    KJ_EXPECT(connection->pendingServerObjects() > 0);

     // A drain must block while the body runs and return only once it
     // finishes, which is what Ipc::disconnectIncoming relies on during
```

test/mp/test/test.cpp
```diff
@@ -425,6 +425,77 @@ KJ_TEST("Calling async IPC method, with server disconnect after cleanup")
...
+    // finishes, which is what Ipc::disconnectIncoming relies on during
+    // shutdown.
+    std::atomic<bool> drained{false};
+    std::thread drain_thread([&] {
+        connection->waitDrained();
+        drained = true;
+    });
+
+    // The body is still blocked, so waitDrained() must not have returned.
+    std::this_thread::sleep_for(std::chrono::milliseconds(20));
```
enirox001, line 607: In commit https://github.com/bitcoin-core/libmultiprocess/pull/335/changes/092d1db8fe7531e220c90bab0c4a78c83bdd978c: _test: cover draining in-flight server call after disconnect_

The 20 ms check is a bit too scheduler dependent, if the drain thread has not been scheduled during that interval, drained remains false even if `waitDrained()` is broken and would return immediately causing false pass.

We could add a hook here that runs only when `ServerObject::wait()` sees a non zero count and is about to wait

```diff --git a/include/mp/proxy-io.h b/include/mp/proxy-io.h
index 1f77b26..ab506b6 100644
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -494,12 +494,14 @@ struct ServerObjectTracker
     void wait()
     {
         Lock lock(m_mutex);
+        if (m_count != 0 && testing_hook_wait) testing_hook_wait();
         m_cv.wait(lock.m_lock, [this]() MP_REQUIRES(m_mutex) { return m_count == 0; });
     }

     mutable Mutex m_mutex;
     std::condition_variable m_cv;
     size_t m_count MP_GUARDED_BY(m_mutex){0};
+    std::function<void()> testing_hook_wait;
 };

 //! Object holding network & rpc state associated with either an incoming server
diff --git a/test/mp/test/test.cpp b/test/mp/test/test.cpp
index 5bccb86..eec52f6 100644
--- a/test/mp/test/test.cpp
+++ b/test/mp/test/test.cpp
@@ -474,13 +474,17 @@ KJ_TEST("Waiting for in-flight server call to finish after disconnect")
     // A drain must block while the body runs and return only once it
     // finishes, which is what Ipc::disconnectIncoming relies on during
     // shutdown.
+    std::promise<void> drain_waiting;
+    connection->m_server_objects->testing_hook_wait = [&] { drain_waiting.set_value(); };
     std::atomic<bool> drained{false};
     std::thread drain_thread([&] {
         connection->waitDrained();
         drained = true;
     });

-    // The body is still blocked, so waitDrained() must not have returned.
+    // Wait until waitDrained() has observed the live server object and is
+    // about to block, then verify it does not return while the body is blocked.
+    drain_waiting.get_future().get();
     std::this_thread::sleep_for(std::chrono::milliseconds(20));
     KJ_EXPECT(!drained);
```

The test will then wait for that hook before starting the 20ms check. This ensures the drain thread has entered `wait` and observed pending work. This substantially reduces the possibility of a false positive

src/mp/proxy.cpp
```diff
@@ -478,13 +487,57 @@ ProxyClient<Thread>::~ProxyClient()
...
 }
 
+ThreadContext::~ThreadContext()
+{
+    // Destroy the thread client maps entry by entry: remove each entry from
+    // its map while holding Waiter::m_mutex, since event loop threads
+    // concurrently remove entries when connections are broken (see SetThread
+    // cleanup function), then destroy the removed ProxyClient<Thread> with the
+    // mutex released, since its destructor needs to lock EventLoop::m_mutex
+    // and Waiter::m_mutex must not be held when EventLoop::m_mutex is
```
enirox001, line 505: In commit https://github.com/bitcoin-core/libmultiprocess/pull/335/changes/901a090da03ed6b04d46a4040ed16fdeb238e7fd: _Fix thread map teardown race causing use-after-free on disconnect_

The `Waiter` documentation says 
```
//! This mutex can be held at the same time as
//! EventLoop::m_mutex as long as Waiter::mutex is locked first and
//! EventLoop::m_mutex is locked second.
```

But the new commit says 
```
//! Waiter::m_mutex must not be held when EventLoop::m_mutex is
//! acquired
```

It also says releasing the waiter mutex avoids locking the Waiter mutex before the EventLoop mutex, these rules cannot both be correct. 

I think an actual order should be identified and updated here

enirox001 (commented):
Code Review https://github.com/bitcoin-core/libmultiprocess/pull/335/changes/901a090da03ed6b04d46a4040ed16fdeb238e7fd

Separating connection teardown from destruction and providing a server-call drain functioanlity is a good addition. The overall approach makes sense. I intend to review this more

I think the commit messages and code documentation are a bit too verbose. The explanations are nice to have, but it overexplains quite often, which ultimately makes it a bit harder to understand. Would suggest some revisions to the commit messages and the documentation to increase clarity

In commit https://github.com/bitcoin-core/libmultiprocess/pull/335/changes/39cc757fba71871b154d915b02cf3967fe73fe0d: _ipc: add Connection::disconnect() separating teardown from destruction_ 

I also think this is not exactly a behavior-neutral change; the commit message itself says `Two details are new`: as we now explicitly cancel `m_on_disconnect` handlers before severing the connection, and explicitly release `m_thread_pool` and `m_thread_map` during `disconnect()` rather than relying on member destruction.

The `m_on_disconnect` change is especially not something I would call behavior-neutral, as now we have to proactively cancel because `Connection` remains alive after the transport is severed and is no longer a consequence of destruction teardown. So even though the externally observable behaviour might seem unchanged, the lifetime and cancellation behaviour has changed, and I think that distinction matters

So the text saying

```
“This is a behavior-neutral refactor: the same steps run in the same order on destruction.”
```

is a bit misleading i think?

Also, in commit https://github.com/bitcoin-core/libmultiprocess/pull/335/changes/a40189f5bbed89f09e5d63ce8377dc90574983f6, there does not seem to be a clear commit title and description here; they are together

Left a few more suggestions and nits below

include/mp/proxy-io.h
```diff
@@ -22,6 +22,7 @@
 #include <map>
 #include <memory>
 #include <optional>
+#include <ranges>
```
ryanofsky: >

src/mp/proxy.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/335#discussion_r3821782150

I think I want to drop the `m_disconnected` variable and just treat `m_network` being nullopt the same as `m_disconnected` being true, which I think should be equivalent to your suggestions.

It's also true that cleanup functions throwing is not something that this library handles very well generally, and could handle better in many cases.

include/mp/proxy-io.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/335#discussion_r3821831654

Hmm, this is an interesting finding. But if there is a bug here, it seems like a pre-existing one, not something caused by this change or made worse by it.

You're saying if a remote disconnect happens first, and the `m_network->onDisconnect()` callback executes, but the `kj::evalLater` callback inside it does **not** execute yet, and if within that interval, the local process decided to delete the connection, then the connection could be deleted twice.

This does seem like it might be possible, and I'd want to look into it a little more and write a test. I'd still be inclined to save a fix for a different PR, and I believe as you pointed out #336 might fix this.

ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/335#discussion_r3821865611

Good catch and nice test!

ryanofsky (commented):
Thanks for the review! Great catches and suggestions. Just left some quick feedback below to make sure I didn't miss anything

include/mp/proxy-io.h
enirox001: This is not exactly the point I intended to pass across; it was more 

- remote callback queued
- local code calls `disconnect()`, intending to keep the connection alive
- local code retains the connection pointer for `waitDrained`
- queued callback erases and destroys the `Connection`
- shutdown code uses the dangling connection pointer

It is not a 
- local code deletes the connections
- queued callback deletes it again

But I think it could be possible for the connection object to be deleted twice, once by the object and another time by a queued callback _(which is similar to the original concern i had)_ and yes, this would be a pre-existing issue

I added a test to verify this _(to an extent)_. First of all, I added a hook to be called before an `onDisconnect` callback is queued on the event loop

```diff --git a/include/mp/proxy-io.h b/include/mp/proxy-io.h
index 1f77b26..100bd10 100644
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -378,6 +378,9 @@ public:

     //! Hook called on the event loop thread when a client has disconnected.
     std::function<void()> testing_hook_disconnected;
+
+    //! Hook called before an onDisconnect callback is queued on the event loop.
+    std::function<void()> testing_hook_before_on_disconnect_queued;
 };

 //! Single element task queue used to handle recursive capnp calls. (If the
@@ -577,7 +580,12 @@ public:
         // to the EventLoop TaskSet to avoid "Promise callback destroyed itself"
         // error in the typical case where f deletes this Connection object.
         m_on_disconnect->add(m_network->onDisconnect().then(
-            [f = std::forward<F>(f), this]() mutable { m_loop->m_task_set->add(kj::evalLater(kj::mv(f))); }));
+            [f = std::forward<F>(f), this]() mutable {
+                if (m_loop->testing_hook_before_on_disconnect_queued) {
+                    m_loop->testing_hook_before_on_disconnect_queued();
+                }
+                m_loop->m_task_set->add(kj::evalLater(kj::mv(f)));
+            }));
     }

     EventLoopRef m_loop;
```

and then wrote a test that cancels an already queued onDisconnect callback

```diff --git a/test/mp/test/test.cpp b/test/mp/test/test.cpp
index 5bccb86..4fd906c 100644
--- a/test/mp/test/test.cpp
+++ b/test/mp/test/test.cpp
@@ -291,6 +291,42 @@ KJ_TEST("Calling IPC method after server connection is closed")
     EXPECT_EXCEPTION(foo->add(1, 2), "IPC client method call interrupted by disconnect.");
 }

+KJ_TEST("Destroying a connection cancels an already queued onDisconnect callback")
+{
+    std::promise<bool> result;
+    std::thread loop_thread{[&] {
+        EventLoop loop("mptest", [](mp::LogMessage) {});
+        auto pipe = loop.m_io_context.provider->newTwoWayPipe();
+        auto server_connection =
+            std::make_unique<Connection>(loop, kj::mv(pipe.ends[0]), [&](Connection& connection) {
+                return capnp::Capability::Client(kj::heap<ProxyServer<messages::FooInterface>>(
+                    std::make_shared<FooImplementation>(), connection));
+            });
+        auto client_connection = std::make_unique<Connection>(loop, kj::mv(pipe.ends[1]));
+        auto client = client_connection->m_rpc_system->bootstrap(ServerVatId().vat_id).castAs<messages::FooInterface>();
+        bool callback_ran{false};
+
+        server_connection->onDisconnect([&] { callback_ran = true; });
+        loop.testing_hook_before_on_disconnect_queued = [&] {
+            loop.m_task_set->add(kj::evalLater([&] {
+                server_connection.reset();
+                loop.m_task_set->add(kj::evalLater([&] {
+                    client = nullptr;
+                    client_connection.reset();
+                    result.set_value(callback_ran);
+                }));
+            }));
+        };
+
+        loop.m_task_set->add(kj::evalLater([&] { client_connection->disconnect(); }));
+        loop.loop();
+    }};
+
+    const bool callback_ran{result.get_future().get()};
+    loop_thread.join();
+    KJ_EXPECT(!callback_ran);
+}
+
 KJ_TEST("Calling IPC method and disconnecting during the call")
 {
     TestSetup setup{/*client_owns_connection=*/false}
```
This test fails

This shows that the onDisconnect callback can execute after its owning `Connection` has already been destroyed, and if this happens to try to delete the Connection object, it could lead to undefined behavior

xyzconstant:
Code review 901a090

I agree with @enirox001 that some comments and commit messages are quite confusing. Some are written like a story, e.g., "Previously..." clauses that add little value to the code. I had to ignore them because reading the code itself was simpler for me to understand the changes.

Planning to review again once there are more updates.

include/mp/proxy-io.h
ryanofsky: >

include/mp/proxy-io.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/335#discussion_r3821865611

> Added this change to so that the listener count can be updated for every disconnect, while automatic deletion happens only for remote disconnects

Thanks for the bug report, and fix, and test. This is a separate, preexisting bug so I made a new PR #361 to address it. Your changes are in bb21177965e363b32878f8258cc2bc1f9a8661f4 there. This bug isn't a practical problem for bitcoin core because it does not disconnect IPC clients except when it is shutting down. But it could a problem for other IPC servers using this code. The problem is also not new to this PR. Even though this PR is adding a disconnect method which makes it possible to disconnect clients *without* deleting the Connection objects, it was always possible to disconnect clients by deleting the Connection objects.

include/mp/proxy-io.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/335#discussion_r3821893669

> Connection can now remain alive after it has been disconnected, but the class does not define which methods are safe to call afterwards.

Added some documentation that after calling disconnect method methods that perform i/o won't work. In general, I would like to make methods safe to call and avoid having unnecessary restrictions.

src/mp/proxy.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/335#discussion_r3822939172

> i think we can assert that the disconnect method has been called before calling waitDrained as such

Was there a specific scenario that made this assert seem useful? It should be fine to call waitDrained regardless of whether the disconnect method was called. It would seem useful to do that if you want to detect when there's a disconnect and server objects are no longer in use, and don't care whether the disconnect was initiated locally or remotely.

include/mp/proxy-io.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/335#discussion_r3823207838

>  The name waitDrained could be clearer if it is called waitServerCallsDrained? or at least document its exact scope a bit more clearly

This is a difficult method to name and I think it might be better to focus on improving documentation if something is unclear. The specific problem with `waitServerCallsDrained` is this doesn't **just** wait for calls to finish it also waits for objects to be released.

Fundamentally this method is meant to be useful for waiting until it is safe to free resources associated with the connection, and I think it makes sense to name it after its purpose instead after how it happens to be implemented at the moment. It could easily be implemented other ways such as by counting calls directly instead of on relying on PassField for mp.Context parameters using `thisCap`.

I did make a number of documentation updates here and the main place semantics of object counting are explained is the `ServerObjectTracker` documentation comment.

src/mp/proxy.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/335#discussion_r3825336664

> I think I want to drop the `m_disconnected` variable and just treat `m_network` being nullopt the same as `m_disconnected` being true, which I think should be equivalent to your suggestions.

I did drop this extra variable in latest push, but didn't look into the exception safety yet. As mentioned previously there are many other places in the library where unexpected exceptions from callbacks will cause problems. I do want revisit and see if there's an improvement that can be made here but would want to keep scope limited and not get into fixing preexisting problems because that could really increase the size of this change.

include/mp/proxy-io.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/335#discussion_r3829464067

>* shutdown code uses the dangling connection pointer

Thanks yes that makes sense. I was assuming the only thing shutdown code would be doing with the pointer would be deleting it. But of course if it did something else with the pointer (like call `waitDrained`) the symptom of the bug would be a use-after-free and not a double delete.

Either way, this is a preexisting bug, so I adopted your test and fix and used them in #361 commit bc98767dadb4d67b7cb6c87ca86696a7f888d715.

ryanofsky:
Thanks for the reviews! I implemented fixes for the two preexisting bugs that were pointed out here in a new PR #361, that this PR is now based on so it would makes sense to review that PR first. I've partially addressed some other comments here as well but am still working on things.

Rebased 901a090da03ed6b04d46a4040ed16fdeb238e7fd -> 33ab2153d47fd673db1a6b61bb064e262f9241ba ([`pr/keepconn.4`](https://github.com/ryanofsky/libmultiprocess/commits/pr/keepconn.4) -> [`pr/keepconn.5`](https://github.com/ryanofsky/libmultiprocess/commits/pr/keepconn.5), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/keepconn.4-rebase..pr/keepconn.5)) based on #361 implementing most review feedback

example/calculator.cpp
```diff
@@ -10,7 +10,6 @@
 #include <fstream>
 #include <functional>
 #include <iostream>
-#include <kj/async.h>
```
ryanofsky: >

test/mp/test/test.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/335#discussion_r3824414437

> The 20 ms check is a bit too scheduler dependent, if the drain thread has not been scheduled during that interval, drained remains false even if `waitDrained()` is broken and would return immediately causing false pass.

Thanks! Applied patch in latest push

test/mp/test/test.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/335#discussion_r3824359770

> I think using an exact count of one is a bit brittle; the test only needs to establish that something remains in flight. This would be less implementation-specific

It is true test could be a little more brittle, but I think checking actual counts makes the test easier to understand and makes the code match the comments. But I would agree if there was a change that did cause these checks to break, that would be evidence these are too brittle, and would be good to make checks less strict at that point.

src/mp/proxy.cpp
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/335#discussion_r3825043796

> I think an actual order should be identified and updated here

Nice catch. This is actually not the first comment that got the order backwards so I added a new commit to base PR #361 to improve all the documentation about lock order. (It's part of #361 not this PR because I also moved the "Fix thread map teardown race causing" commit you referenced to #361.)

ryanofsky (commented):
Updated 33ab2153d47fd673db1a6b61bb064e262f9241ba -> 40cfb9332a1a47e4bc55f03aeb87787285f89975 ([`pr/keepconn.5`](https://github.com/ryanofsky/libmultiprocess/commits/pr/keepconn.5) -> [`pr/keepconn.6`](https://github.com/ryanofsky/libmultiprocess/commits/pr/keepconn.6), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/keepconn.5..pr/keepconn.6)) improving comments and naming and moving more changes to base PR #361 to simplify commits here.

re:  https://github.com/bitcoin-core/libmultiprocess/pull/335#pullrequestreview-4982982109

> I also think this is not exactly a behavior-neutral change

Thanks, clarified commit message to say this is a behavior-neutral change for existing callers not calling the new methods. For callers that do specifically call the new disconnect() method, there are some differences from destroying the Connection object.

PR #338, opened 2026-08-07
test: check ReadList return value

ryanofsky:
This is a followup to https://github.com/bitcoin-core/libmultiprocess/pull/285#discussion_r3509475689, adding a test to verify the `ReadList` return value and make sure it is compatible with `ReadDestTemp`


ViniciusCestarii (commented):
ACK 62f25af06c39636d5e09d85753d3e958f264eec1

PR #342, opened 2026-08-12
Allow request cancellation for wrapped C++ methods

xyzconstant:
Currently, libmultiprocess requests are blocking, with no way to cancel them from either the client or server side. Downstream code works around this by pairing each blocking method with a dedicated `interruptX()` method whose only purpose is to wake it up. For example, bitcoin/bitcoin#33676 introduced `BlockTemplate::interruptWait()` specifically to wake an in-progress `BlockTemplate::waitNext()` call.

Cap'n Proto provides a useful [cancellation mechanism](https://capnproto.org/cxxrpc.html#cancellation) that libmultiprocess can use to support cancellation from both non-libmultiprocess and libmultiprocess clients. When a promise is dropped, it sends a cancellation request. The server may either [cancel immediately](https://github.com/capnproto/capnproto/blob/master/c++/src/capnp/rpc.capnp#L575-L580) or ignore the signal.

This PR implements approach 4 from bitcoin/bitcoin#33575, adding cancellation support on both sides:

- Allow `ProxyClient` method calls to be canceled from another thread. `ClientInvokeContext` provides two hooks an application overload can set: `set_canceler`, which receives the `kj::Canceler` that `clientInvoke` wraps the request promise in, and `handle_error`, which receives the `kj::Exception` from a failed request and decides what the caller sees. The second hook is how a cancel parameter turns the `DISCONNECTED` that a `kj::Canceler` raises into `InterruptException`.
 - Allow wrapped server methods to register a callback that runs when the request is canceled, whether because the promise was dropped or the connection was interrupted.

The PR also adds a new capnp annotation, `$Proxy.extraParam`, which declares C++-only parameters that are not sent through RPC. These parameters are handled by `CustomBuildExtraParam` on the client side and `CustomReadExtraParam` on the server side. Paired with the cancellation mechanism, this feature lets applications define their own cancel parameter types. A reference implementation lives under `test/mp/test/foo-types.h` and `test/mp/test/foo.h`, used in tests.

For example, given the following capnp schema method:

```capnp
waitNext @0 (context :Proxy.Context) -> (result :Template) $Proxy.extraParam("cancel") $Cxx.allowCancellation;
```

and this C++ method with an application-defined cancel parameter:

```c++
using CancelFn = std::function<void()>;
class CancelGuard;   // unregisters the callback when destroyed
using CancelArg = std::function<CancelGuard(CancelFn)>;

virtual std::unique_ptr<Template> waitNext(CancelArg cancel) = 0;
```

The caller (client) passes a callback that receives the function that cancels the call:

```c++
CancelFn cancel_fn;

// ... blocks until the result arrives or another thread runs cancel_fn()
auto tmpl = client->waitNext([&](CancelFn fn) {
    cancel_fn = std::move(fn);
    return CancelGuard{};
});
```

The callee (server) registers a callback that interrupts its wait and retains the returned guard until the wait ends:

```c++
std::unique_ptr<Template> waitNext(CancelArg cancel) override
{
    CancelGuard guard;
    if (cancel) guard = cancel([this] { m_cv.notify_all(); });

    // ... wait on m_cv, checking for cancellation
}
```

Detecting a dropped promise on the server requires the `$Cxx.allowCancellation` annotation on the method, file, or interface. Without it, Cap'n Proto runs the abandoned call to completion. The annotation requires Cap'n Proto 1.0 (see the "Breaking change" section in https://capnproto.org/news/2023-07-28-capnproto-1.0.html), which this PR also sets as the minimum supported version.

---

NOTE: bitcoin/bitcoin#36097 builds on this PR by adding cancel arguments to the blocking mining methods (`waitTipChanged`, `createNewBlock`, `BlockTemplate::waitNext`) and deprecating the `interrupt()`/`interruptWait()` workarounds.


include/mp/proxy-types.h
```diff
@@ -534,12 +534,70 @@ ClientParam<Accessor, Types...> MakeClientParam(Types&&... values)
...
+        ReadResults(ClientParam*) {}
+    };
+
+    std::tuple<Types&&...> m_values;
+};
+
+template <typename LocalType, typename Value>
+void MaybeBuildExtraParam(TypeList<LocalType> param, ClientInvokeContext& invoke_context, Value&& value)
+{
+    if constexpr (requires { CustomBuildExtraParam(param, invoke_context, std::forward<Value>(value)); }) {
```
ryanofsky: In commit "Edited init.cpp to include a check that -datadir exists" (ed2c01405f5a5b913c4fd64397394559051e041a)

Can we change this to a static_assert like MaybeReadExtraParam below instead of discarding the argument passed by the client?

I think this would be safer because it's easy to define a CustomBuildExtraParam function with a slightly incorrect signature that could get silently skipped. And I think in practice for real use-cases, it should make sense to define CustomBuildExtraParam and CustomReadExtraParam functions in pairs, and easy to leave implementations empty if one of them is not needed.

include/mp/proxy-types.h
```diff
@@ -534,12 +534,70 @@ ClientParam<Accessor, Types...> MakeClientParam(Types&&... values)
...
+        {
+        }
+        ReadResults(ClientParam*) {}
+    };
+
+    std::tuple<Types&&...> m_values;
+};
+
+template <typename LocalType, typename Value>
+void MaybeBuildExtraParam(TypeList<LocalType> param, ClientInvokeContext& invoke_context, Value&& value)
```
ryanofsky: In commit "Allow wrapped C++ methods to take parameters that are not sent over RPC" (88ea9fa0a791fa20f3bc21b85093aeab3d34c09c)

These functions (unlike MaybeReadField and MaybeBuildField) are only called one place, and it would seem simpler and less confusing to just inline them where they are used.

src/mp/gen.cpp
```diff
@@ -542,6 +547,42 @@ static void Generate(kj::StringPtr src_prefix,
...
+                    if (is_construct || is_destroy) {
+                        throw std::runtime_error(method_prefix +
+                                                 ": $Proxy.extraParam is not supported on construct and destroy methods, which have no "
+                                                 "corresponding C++ method");
+                    }
+                    int annotations_count{0};
+                    for (const auto annotation : method.getProto().getAnnotations()) {
+                        if (annotation.getId() == EXTRA_PARAM_ANNOTATION_ID) ++annotations_count;
+                    }
+                    if (annotations_count > 1) {
```
ryanofsky: In commit "Allow wrapped C++ methods to take parameters that are not sent over RPC" (88ea9fa0a791fa20f3bc21b85093aeab3d34c09c)

Is there a specific reason this limitation was added? It seems like multiple extraparams could work without any extra support, so unclear why they are not allowed. Similar for the other checks, it doesn't really seem like it should be a problem for these parameters not to have names for the names not to be unique. The names seem like they can only exist for descriptive purposes, and there's not really anything meaningful we could do with them, so there doesn't seem to be a reason to require them or restrict them.

test/mp/test/test.cpp
```diff
@@ -267,6 +267,11 @@ KJ_TEST("Call FooInterface methods")
 
     KJ_EXPECT(foo->passFn([]{ return 10; }) == 10);
 
+    // The `CustomReadExtraParam` overload in `foo-types.h` builds the
+    // server-side value, hardcoded to 1. As a result this always returns
+    // arg + 1 regardless of the value passed for extra.
+    KJ_EXPECT(foo->passExtra(1, 999) == 2);
```
ryanofsky, line 276: In commit "Allow wrapped C++ methods to take parameters that are not sent over RPC" (88ea9fa0a791fa20f3bc21b85093aeab3d34c09c)

Test could potentially verify that CustomBuildExtraParam is called and 999 value is received (`testing_hook_misc` from 914dc839f279269d11bd575f61dc1c09c5c2008e could work well for this)

ryanofsky (approved):
Code review 0d7a2ffebf1611e4513fc4b620541d35a51ff8ed. Starting to go through this, looks very good so far!

include/mp/proxy-types.h
xyzconstant: Done at 1eaa50151680e7ebb20532c95c22c25f4593df59

include/mp/proxy-types.h
xyzconstant: Thanks! done at 1eaa50151680e7ebb20532c95c22c25f4593df59

src/mp/gen.cpp
xyzconstant: The only reason was to ensure that only 1 extra parameter was allowed, and yes you're right. This works without extra support, so this restriction no longer makes sense. Thanks!

Done at 1eaa50151680e7ebb20532c95c22c25f4593df59

test/mp/test/test.cpp
xyzconstant: Nice, I added the misc hook at 1eaa50151680e7ebb20532c95c22c25f4593df59

src/mp/gen.cpp
```diff
@@ -559,9 +588,10 @@ static void Generate(kj::StringPtr src_prefix,
                 for (const auto& field : fields.fields) {
                     if (field.skip) continue;
 
-                    const auto& f = field.param_is_set ? field.param : field.result;
-                    auto field_name = f.getProto().getName();
-                    add_accessor(field_name);
+                    const bool extra{field.extra_name.size() > 0};
+                    const auto field_name =
+                        extra ? field.extra_name : (field.param_is_set ? field.param : field.result).getProto().getName();
```
ViniciusCestarii: In "Allow wrapped C++ methods to take parameters that are not sent over RPC" 1eaa50151680e7ebb20532c95c22c25f4593df59

nit: nested ternaries here are hard to read. Maybe:

```cpp
    kj::StringPtr field_name;
    if (extra) {
        field_name = field.extra_name;
    } else {
        field_name = (field.param_is_set ? field.param : field.result).getProto().getName();
        add_accessor(field_name);
    }
```

which also folds in the `if (!extra) add_accessor(...)` on the next line.

src/mp/gen.cpp
```diff
@@ -613,6 +649,12 @@ static void Generate(kj::StringPtr src_prefix,
 
                 client << "    using M" << method_ordinal << " = ProxyClientMethodTraits<" << method_prefix
                        << "Params>;\n";
+                if (has_extra) {
+                    client << "    static_assert(M" << method_ordinal << "::Params::size == " << argc
+                           << ", \"C++ method " << proxied_class_type << "::" << proxied_method_name
+                           << " should have " << argc << " parameters to match capnp method " << method_prefix
+                           << ".\");\n";
+                }
```
ViniciusCestarii: In "Allow wrapped C++ methods to take parameters that are not sent over RPC" 1eaa50151680e7ebb20532c95c22c25f4593df59

nit: not sure if this is helpful. Arity mismatches are already compile errors without it, and since static_assert doesn't stop further diagnostics it only prepends a nicer first line rather than replacing the noise. It also fires only for methods with $Proxy.extraParam, while an ordinary method with the same mismatch gets nothing.

enirox001 (commented):
Concept ACK

IIUC, this allows one thread to cancel an IPC call that another thread is blocked waiting for, without closing the connection. The server method can also register a callback to respond to cancellation.

This seems complementary to #335 as thesse both provide mechanisms in libmultiprocess that reduce the need for custom interruption and shutdown handling in Bitcoin Core. That seems like a useful direction.

include/mp/proxy-types.h
```diff
@@ -550,9 +599,13 @@ struct ServerCall
...
         return TryFinally(
             [&]() -> decltype(auto) {
-                return ProxyServerMethodTraits<
-                    typename decltype(server_context.call_context.getParams())::Reads
-                >::invoke(server_context, std::forward<Args>(args)...);
+                return std::apply(
+                    [&](RemoveCvRef<Extra>&... extra_args) -> decltype(auto) {
+                        return ProxyServerMethodTraits<
+                            typename decltype(server_context.call_context.getParams())::Reads
+                        >::invoke(server_context, std::forward<Args>(args)..., extra_args...);
```
enirox001: In https://github.com/bitcoin-core/libmultiprocess/pull/342/changes/1eaa50151680e7ebb20532c95c22c25f4593df59: _Allow wrapped C++ methods to take parameters that are not sent over RPC_

nit: 
Here, the extra argument passed by value here gets copied; I assume the problems with this have been considered since it was made to use an `std::move` in cfdcb45

But I think the move should be introduced in this commit instead

src/mp/gen.cpp
```diff
@@ -559,9 +588,10 @@ static void Generate(kj::StringPtr src_prefix,
                 for (const auto& field : fields.fields) {
                     if (field.skip) continue;
 
-                    const auto& f = field.param_is_set ? field.param : field.result;
-                    auto field_name = f.getProto().getName();
-                    add_accessor(field_name);
+                    const bool extra{field.extra_name.size() > 0};
```
enirox001: In https://github.com/bitcoin-core/libmultiprocess/pull/342/changes/1eaa50151680e7ebb20532c95c22c25f4593df59: _Allow wrapped C++ methods to take parameters that are not sent over RPC_

We can improve how the generator identifies an extra parameter; currently it uses a non-empty name as a marker. Still, if there is an extra entry that is empty, `$Proxy.extraParam("")` that entry has no real capnp field, but the generator treats it as an ordinary field, because its name is empty; it then attempts to obtain a name from the nonexistent field. 

The generator also exits successfully, but produces invalid c++
```
MakeClientParam<Accessor<empty_fields::, FIELD_OUT>>(M0::Fwd<1>())
```

I think we should be rejecting empty names with a clear error

include/mp/proxy-io.h
```diff
@@ -58,11 +58,11 @@ struct ServerInvokeContext : InvokeContext
     //! results structs if the request is canceled while the worker thread is
     //! reading params (`call_context.getParams()`) or writing results
     //! (`call_context.getResults()`).
-    Lock* cancel_lock{nullptr};
+    Lock* request_lock{nullptr};
```
enirox001, line 74: In commit https://github.com/bitcoin-core/libmultiprocess/pull/342/changes/7f3199674a225215f7412470e52628461e909ef5: _proxy: rename `cancel_lock` and `cancel_mutex` to `request_lock` and `request_mutex`_

This is a good change to use the naming `request_lock`, the previous name would have been confusing, as it could suggest that it controls whether cancellation happens. The new change better describes what it protects, which is the request data.

nit: there is a comment that I think you missed;

```diff --git a/include/mp/type-context.h b/include/mp/type-context.h
index 63b3fdb..c1ca04f 100644
--- a/include/mp/type-context.h
+++ b/include/mp/type-context.h
@@ -154,7 +154,7 @@ auto PassField(Priority<1>, TypeList<>, ServerContext& server_context, const Fn&
         // makes another IPC call), so avoid modifying the map.
         const bool erase_thread{inserted};
         KJ_DEFER(
-            // Release the cancel lock before calling loop->sync and
+            // Release the request lock before calling loop->sync and
             // waiting for the event loop thread, because if a
             // cancellation happened, it needs to run the on_cancel
             // callback above. It's safe to release request_lock at
```

include/mp/proxy-types.h
```diff
@@ -842,6 +845,10 @@ void clientInvoke(ProxyClient& proxy_client, const GetRequest& get_request, Fiel
             }).attach(kj::mv(request_canceler)));
     });
 
+    if (invoke_context && invoke_context->cancel_receiver) {
```
ViniciusCestarii: In "Allow request cancellation for wrapped C++ methods" cfdcb45fbc616719ff4b312f66f76c6b7e27ce28

nit: I found "cancel_receiver" a confusing name because it makes you think it receives cancellation but it is only a registration hook. A better name could be "register_cancel"

ViniciusCestarii (commented):
Concept ACK. Commented 3 nits.

I also used claude to review and it found that ServerCall::invoke cancellation is currently racy:

"The server unlocks request_lock for the whole method body and only reacquires it afterwards to clear cancel_fn. In the gap between the method returning (its frame, and anything the callback captured from it, destroyed) and that reacquire, the event loop can take the mutex, see cancel_fn still set, and invoke a callback whose captures are gone."

The race window seems pretty narrow but possible.

<details>
<summary>
Diff widening this race window with sleep and a test demonstrating the race
</summary>

```
diff --git a/include/mp/proxy-types.h b/include/mp/proxy-types.h
index 2172326..25edf50 100644
--- a/include/mp/proxy-types.h
+++ b/include/mp/proxy-types.h
@@ -7,8 +7,10 @@
 
 #include <mp/proxy-io.h>
 
+#include <chrono>
 #include <exception>
 #include <optional>
+#include <thread>
 #include <set>
 #include <typeindex>
 #include <vector>
@@ -608,6 +610,10 @@ struct ServerCall
                     extra);
             },
             [&] {
+                // TEMPORARY REPRO HACK: widen the existing window between the
+                // wrapped method returning (its locals already destroyed) and
+                // the worker thread reacquiring request_lock.
+                if (server_context.request_lock) std::this_thread::sleep_for(std::chrono::milliseconds{500});
                 if (server_context.request_lock) server_context.request_lock->m_lock.lock();
                 // The method returned, so destroy the callback it registered
                 // through its cancellation argument, if any.
index 619930e..fefd141 100644
--- a/test/mp/test/test.cpp
+++ b/test/mp/test/test.cpp
@@ -816,6 +816,55 @@ KJ_TEST("Cancel an in-flight IPC call")
     KJ_EXPECT(foo->add(1, 2) == 3);
 }
 
+KJ_TEST("Cancel racing method return is not a use-after-free")
+{
+    TestSetup setup;
+    ProxyClient<messages::FooInterface>* foo = setup.client.get();
+    foo->initThreadMap();
+    std::promise<void> registered;
+    std::atomic<bool> callback_ran{false};
+    std::atomic<bool> ran_after_return{false};
+
+    // Register a cancel callback, then return immediately instead of waiting
+    // for cancellation. This targets the window in proxy-types.h between the
+    // wrapped method returning (its frame, and anything the callback captured
+    // from it, destroyed) and request_lock being reacquired to clear
+    // cancel_fn. The callback holds a weak_ptr to a frame-owned canary
+    // instead of a dangling reference, so it can observe that it was invoked
+    // after the frame died without actually corrupting memory.
+    setup.server->m_impl->m_cancel_fn = [&](CancelArg cancel) {
+        auto canary = std::make_shared<int>(42);
+        cancel([weak = std::weak_ptr<int>{canary}, &callback_ran, &ran_after_return] {
+            callback_ran = true;
+            if (weak.expired()) ran_after_return = true;
+        });
+        registered.set_value();
+    };
+
+    std::promise<CancelFn> cancel_fn;
+    std::thread canceler([&] {
+        CancelFn fire{cancel_fn.get_future().get()};
+        registered.get_future().wait();
+        // Let the method return before canceling, to land in the window.
+        std::this_thread::sleep_for(std::chrono::milliseconds{100});
+        fire();
+    });
+    bool interrupted = false;
+    try {
+        foo->callCancelFnAsync([&](CancelFn fn) { cancel_fn.set_value(std::move(fn)); });
+    } catch (const InterruptException&) {
+        interrupted = true;
+    }
+    canceler.join();
+    KJ_EXPECT(interrupted);
+    KJ_EXPECT(callback_ran);
+    // Fails today: the callback is invoked after the method's frame is gone.
+    KJ_EXPECT(!ran_after_return);
+
+    // Connection should be unaffected.
+    KJ_EXPECT(foo->add(1, 2) == 3);
+}
+
 KJ_TEST("Dropping the client promise cancels an executing method")
 {
     TestSetup setup;
```
</details>

A possible fix is having registration return an RAII handle that clears cancel_fn under the mutex.

include/mp/proxy-types.h
```diff
@@ -603,12 +603,15 @@ struct ServerCall
...
-                        >::invoke(server_context, std::forward<Args>(args)..., extra_args...);
+                        >::invoke(server_context, std::forward<Args>(args)..., std::move(extra_args)...);
                     },
                     extra);
             },
             [&] {
                 if (server_context.request_lock) server_context.request_lock->m_lock.lock();
+                // The method returned, so destroy the callback it registered
+                // through its cancellation argument, if any.
+                server_context.cancel_fn = nullptr;
```
enirox001, line 617: In https://github.com/bitcoin-core/libmultiprocess/pull/342/changes/cfdcb45fbc616719ff4b312f66f76c6b7e27ce28: _proxy: support cancellation extra parameters_

Could cancellation run the server callback after the local variables it references have been destroyed?

The callback is cleared only after the wrapped method returns, but the method’s local variables are destroyed before that. If cancellation arrives in between, the callback could access destroyed state—for example, a local mutex or condition variable.

I reproduced this with a test that pauses during method cleanup and triggers cancellation after the local state is destroyed. The callback still runs.

```diff --git a/test/mp/test/test.cpp b/test/mp/test/test.cpp
index 619930e..5ad02a0 100644
--- a/test/mp/test/test.cpp
+++ b/test/mp/test/test.cpp
@@ -816,6 +816,85 @@ KJ_TEST("Cancel an in-flight IPC call")
     KJ_EXPECT(foo->add(1, 2) == 3);
 }

+KJ_TEST("Cancellation callback does not outlive server method locals")
+{
+    TestSetup setup;
+    auto* foo = setup.client.get();
+    foo->initThreadMap();
+    constexpr std::chrono::seconds timeout{30};
+    std::promise<void> locals_destroyed;
+    auto locals_destroyed_future = locals_destroyed.get_future();
+    std::promise<void> cancellation_attempted;
+    auto cancellation_attempted_future = cancellation_attempted.get_future();
+    std::atomic<bool> callback_after_destruction{false};
+
+    setup.server->m_impl->m_cancel_fn = [&](CancelArg cancel) {
+        // This guard runs after local_state is destroyed, but before the
+        // method returns to ServerCall's callback cleanup. Hold that window
+        // open so cancellation reliably arrives during normal scope exit.
+        KJ_DEFER(
+            locals_destroyed.set_value();
+            KJ_EXPECT(cancellation_attempted_future.wait_for(timeout) == std::future_status::ready);
+        );
+        );
+            KJ_EXPECT(cancellation_attempted_future.wait_for(timeout) == std::future_status::ready);
+        );
+        auto local_state = std::make_shared<int>(42);
+        cancel([state = std::weak_ptr<int>{local_state}, &callback_after_destruction] {
+            // Observe the lifetime safely, instead of dereferencing a dangling
+            // reference to a local mutex/condition variable as real code might.
+            callback_after_destruction = state.expired();
+        });
+        // Return normally; do not wait for cancellation in the method body.
+    };
+
+    std::promise<CancelFn> cancel_fn;
+    auto cancel_future = cancel_fn.get_future();
+    std::thread canceler([&] {
+        CancelFn fire = cancel_future.get();
+        KJ_EXPECT(locals_destroyed_future.wait_for(timeout) == std::future_status::ready);
+        fire();
+        cancellation_attempted.set_value();
+    });
+    bool interrupted{false};
+    try {
+        foo->callCancelFnAsync([&](CancelFn fn) { cancel_fn.set_value(std::move(fn)); });
+    } catch (const InterruptException&) {
+        interrupted = true;
+    }
+    canceler.join();
+    KJ_EXPECT(interrupted);
+    KJ_EXPECT(!callback_after_destruction.load());
+    KJ_EXPECT(foo->add(1, 2) == 3);
+}
+
```

Should we provide a way to unregister the callback before the captured local variables go out of scope?

include/mp/proxy-types.h
enirox001: In https://github.com/bitcoin-core/libmultiprocess/pull/342/changes/cfdcb45fbc616719ff4b312f66f76c6b7e27ce28: _proxy: support cancellation extra parameters_

From the previous commit bf8a8b1, every ipc call now creates a `ClientCancelState` allocates a `RequestCanceler` and wraps the response promise even when the caller does not want cancellation. I think with this  commit that connects `cancel_receiver`, we could have a way where this would happen only when a cancellation receiver is present. This would reduce the allocations for ordinary calls

CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ endif()
 include("cmake/compat_find.cmake")
 
 find_package(Threads REQUIRED)
-find_package(CapnProto 0.9 NO_MODULE)
+find_package(CapnProto 1.0 NO_MODULE)
```
enirox001, line 16: In https://github.com/bitcoin-core/libmultiprocess/pull/342/changes/527a099743e4a42391850165a994e6c3aa9347f5: _build: require Cap'n Proto 1.0_

nit: There is a text here that needs to be updated i think

```diff --git a/CMakeLists.txt b/CMakeLists.txt
index f2199cb..a6a8a0f 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@ find_package(CapnProto 1.0 NO_MODULE)
    message(FATAL_ERROR
      "Cap'n Proto is required but was not found.\n"
      "To resolve, choose one of the following:\n"
-     "  - Install Cap'n Proto (version 1.0+ recommended)\n"
+     "  - Install Cap'n Proto (version 1.0+ required)\n"
      "  - For Bitcoin Core compilation build with -DENABLE_IPC=OFF to disable multiprocess support\n"
    )
  endif()
```

src/mp/gen.cpp
xyzconstant: Nice suggestion, updated in c14d0d51eaa6f241d8921eb6b39937e851925faf.

src/mp/gen.cpp
xyzconstant: Nice catch! I agree this is kind of redundant. Dropped this check in c14d0d51eaa6f241d8921eb6b39937e851925faf. Thanks!

include/mp/proxy-types.h
xyzconstant: Updated it to `pass_cancel_fn` in 298806f48c05809f62dd920e2b6aef39ef402bb1

include/mp/proxy-io.h
xyzconstant: Thanks! Done in 2e5e66d7b3949aad55f19ac11eef17561eaec7f6

src/mp/gen.cpp
xyzconstant: Thanks for spotting it. A check for this existed at 88ea9fa0a791fa20f3bc21b85093aeab3d34c09c and threw an error if no name was provided. I removed it after [Ryan's feedback (comment)](https://github.com/bitcoin-core/libmultiprocess/pull/342#discussion_r3898538702) since there's no real reason to require a name. But as you noted, mpgen was still relying on the name's length being > 0, and an empty extraParam fell through the normal capnp path.

Fixed in 780120c02de4eb14aee63b4675ca2d43641f68ef. It adds a new `extra` flag to the Field struct, sets it when it detects the annotation, and replaces the name-length check to initialize the local `extra` variable. The empty name turns into `unnamed_extra_paramN`.

include/mp/proxy-types.h
xyzconstant: Done at 780120c02de4eb14aee63b4675ca2d43641f68ef

include/mp/proxy-types.h
xyzconstant: @ViniciusCestarii reported the same issue and suggested an RAII guard that unregisters the cancel callback before the wrapped method's locals go out of scope.

Following @ryanofsky's suggestion in https://github.com/bitcoin-core/libmultiprocess/pull/342#issuecomment-5587566024, it's a bit of a redesign. In the latest pushes, `type-cancel.h` was moved to `test/`, and now the source code changes are smaller than in your last review.

include/mp/proxy-types.h
xyzconstant: Thanks! Dropped this, and now `clientInvoke` checks whether the new field `invoke_context.set_canceler` is set before constructing a `kj::Canceler` object and attaching it to the request promise.

CMakeLists.txt
xyzconstant: This makes more sense, thanks. Done in 45514afee9e1c0adfa5312ef14875ea12634d0b2.

PR #344, opened 2026-08-13
test: listen_tests and connect_tests follow-ups

xyzconstant:
Addresses review suggestions left (all of them made by ryanofsky) in the now-merged PRs #298 and #310.

These are non-critical test cleanups (naming, simplification, comments, etc.) with zero changes to library code.

### DIFF
--- a/test/mp/test/common.h
+++ b/test/mp/test/common.h
@@ -1,0 +1,27 @@
+// Copyright (c) The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef MP_TEST_COMMON_H
+#define MP_TEST_COMMON_H
+
+#include <kj/debug.h>
+#include <mp/proxy-io.h>
+
+#include <stdexcept>
+
+namespace mp {
+namespace test {
+
+//! Default event loop log handler used by tests. Logs all messages and throws
+//! on errors so calling code can assert on them.
+inline void DefaultLogHandler(LogMessage log)
+{
+    KJ_LOG(INFO, log.level, log.message);
+    if (log.level == Log::Raise) throw std::runtime_error(log.message);
+}
+
+} // namespace test
+} // namespace mp
+
+#endif // MP_TEST_COMMON_H
--- a/test/mp/test/connect_tests.cpp
+++ b/test/mp/test/connect_tests.cpp
@@ -79,7 +79,6 @@ KJ_TEST("ConnectStream connects to a socket serving a valid init interface")
 
     init.reset();
     server_thread.join();
-    KJ_EXPECT(true);
 }
 
 KJ_TEST("ConnectStream throws when the socket is already disconnected")
--- a/test/mp/test/connect_tests.cpp
+++ b/test/mp/test/connect_tests.cpp
@@ -166,14 +166,12 @@ KJ_TEST("ConnectStream throws when the socket disconnects after receiving data")
     try {
         auto init = ConnectStream<messages::FooInit>(*setup.m_loop, MakeStream(*setup.m_loop, client_fd));
 
-        if (server_thread.joinable()) server_thread.join();
         KJ_EXPECT(false);
     } catch (const std::runtime_error& e) {
-        if (server_thread.joinable()) server_thread.join();
-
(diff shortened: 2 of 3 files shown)


test/mp/test/common.h
```diff
@@ -0,0 +1,27 @@
...
+#include <stdexcept>
+
+namespace mp {
+namespace test {
+
+//! Default event loop log handler used by tests. Logs all messages and throws
+//! on errors so calling code can assert on them.
+inline void DefaultLogHandler(LogMessage log)
+{
+    KJ_LOG(INFO, log.level, log.message);
```
ryanofsky, line 20: In commit "test: share DefaultLogHandler between test files" (038d33eb31eb0d5742a27caeb0c5b244f0f794a1)

Looks like [test.cpp](https://github.com/bitcoin-core/libmultiprocess/blob/8dab0d4bdebd91cfaea28d891b56e4d4eb91bfc9/test/mp/test/test.cpp#L92-L95
) is another place this could be used. It would also be nice to take the comment there about showing log output with `mptest --verbose`

test/mp/test/connect_tests.cpp
```diff
@@ -38,17 +38,17 @@ constexpr auto FAILURE_TIMEOUT = std::chrono::seconds{30};
 class TestSetup
 {
 public:
-    mp::EventLoop* m_loop;
-    std::optional<mp::EventLoopRef> m_loop_ref;
+    EventLoop* m_loop;
```
ryanofsky, line 41: In commit "test: drop mp:: prefixes in connect tests" (44bc4630bc18bb03977a3616bf979fb2b3192f54)

Note: looks like there are a few remaining mp:: prefixes that could be dropped in other test files, but reasonable to limit commit to this file

test/mp/test/connect_tests.cpp
```diff
@@ -186,14 +182,10 @@ KJ_TEST("ConnectStream throws when a connection accepted from a listener disconn
         int connection_fd = accept(server_fd, nullptr, nullptr);
 
         if (connection_fd >= 0) {
-            ssize_t bytes_received =
-                recv(connection_fd, buf, sizeof(buf), 0);
-
-            if (bytes_received > 0) {
-                close(connection_fd);
-            }
+            recv(connection_fd, buf, sizeof(buf), 0);
```
ryanofsky, line 185: In commit "test: close sockets unconditionally and check errors with KJ_SYSCALL" (b9c36c617518c0026f3b8c35014415b7437f6b8a)

Would be nice to capture errors here with `KJ_SYSCALL(recv(server_fd, buf, sizeof(buf), 0));` or `KJ_SYSCALL(bytes_received = recv(server_fd, buf, sizeof(buf), 0));` or `KJ_SYSCALL(bytes_received = recv(...), "receiving client handshake");`

test/mp/test/connect_tests.cpp
```diff
@@ -186,14 +182,10 @@ KJ_TEST("ConnectStream throws when a connection accepted from a listener disconn
         int connection_fd = accept(server_fd, nullptr, nullptr);
 
         if (connection_fd >= 0) {
```
ryanofsky, line 184: In commit "test: close sockets unconditionally and check errors with KJ_SYSCALL" (b9c36c617518c0026f3b8c35014415b7437f6b8a)

Seems probably best to fail if accept fails:

```c++
int connection_fd;
KJ_SYSCALL(connection_fd = accept(server_fd, nullptr, nullptr));
```

ryanofsky (approved):
Code review ACK c39c7850c663d1915ae2478dc53f2c7b690504db. Thanks for the followup!

test/mp/test/connect_tests.cpp
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) The Bitcoin Core developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
+#include "common.h"
 #include "unixlistener.h"
```
ryanofsky, line 5: In commit "test: share DefaultLogHandler between test files" (038d33eb31eb0d5742a27caeb0c5b244f0f794a1)

For consistency would be good to use full include paths `<mp/test/common.h>` We should probably have a linter check for this

PR #347, opened 2026-08-16
refactor: Replace EventLoop::post() with sync() taking kj::FunctionParam

ViniciusCestarii:
EventLoop::post() and the sync() wrapper around it did the same thing, so this merges them into a single sync() method. And also takes the callback as a [kj::FunctionParam](https://github.com/capnproto/capnproto/blob/v1.5.0/c%2B%2B/src/kj/function.h#L91) instead of a kj::Function: FunctionParam stores only a pointer to the caller's callable rather than taking ownership of it, so the callable no longer needs to be copyable or movable, and each call avoids a heap allocation.

This is safe because sync() does not return until the event loop thread has finished running the callback and cleared m_post_fn(now named m_sync_fn), so the pointer is never used after the callable goes out of scope.

### DIFF
--- a/include/mp/proxy-io.h
+++ b/include/mp/proxy-io.h
@@ -301,7 +301,7 @@ public:
     std::thread m_async_thread;
 
     //! Callback function to run on event loop thread during sync() call.
-    kj::FunctionParam<void()>* m_post_fn MP_GUARDED_BY(m_mutex) = nullptr;
+    kj::FunctionParam<void()>* m_sync_fn MP_GUARDED_BY(m_mutex) = nullptr;
 
     //! Callback functions to run on async thread.
     std::optional<CleanupList> m_async_fns MP_GUARDED_BY(m_mutex);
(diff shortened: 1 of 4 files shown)


include/mp/proxy-io.h
```diff
@@ -267,16 +267,14 @@ class EventLoop
...
-    {
-        post(std::forward<Callable>(callable));
-    }
+    //!
+    //! The function is accepted as a kj::FunctionParam, which only stores a
+    //! pointer to the caller's callable instead of taking ownership of it. This
+    //! avoids a heap allocation per call, and is safe because this call is
+    //! synchronous, so the callable is guaranteed not to go out of scope before
+    //! the event loop thread is done running it. It also means the callable
+    //! does not need to be copyable or movable.
```
ryanofsky: In commit "Fix segfault when creating new wallet" (d3800d59d557388b813513ef16f6b840467821d4)

I think this comment would be great to have in the commit message, but would suggest removing it from the code because it is mostly just repeating capnproto API documentation. This seems like a straightforward use of the FunctionParam type as documented. It also seems potentially confusing for callers for the method documentation to be describing why the method definition is safe. It would make more sense to write that in the implementation if necessary.

ryanofsky (approved):
Code review ACK 2b23e78b35fe0208a782db17e49b04c4f03aa088. Nice find and simplification!

include/mp/proxy-io.h
ViniciusCestarii: True, thanks. Removed from the code and added as message of commit 2330fbe81e255996710e3e258cb6a5ea94cfb250

PR #350, opened 2026-08-20
cmake: add type-unordered-set.h and version.h to public headers

ryanofsky:
These are useful for building bitcoin core with WITH_EXTERNAL_LIBMULTIPROCESS and having the install target install the complete set of public headers more generally.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -174,9 +174,11 @@ set(MP_PUBLIC_HEADERS
   include/mp/type-struct.h
   include/mp/type-threadmap.h
   include/mp/type-tuple.h
+  include/mp/type-unordered-set.h
   include/mp/type-vector.h
   include/mp/type-void.h
-  include/mp/util.h)
+  include/mp/util.h
+  include/mp/version.h)
 add_library(multiprocess STATIC
   ${MP_PROXY_SRCS}
   ${MP_PUBLIC_HEADERS}


CMakeLists.txt
```diff
@@ -174,9 +174,11 @@ set(MP_PUBLIC_HEADERS
   include/mp/type-struct.h
   include/mp/type-threadmap.h
   include/mp/type-tuple.h
+  include/mp/type-unordered-set.h
   include/mp/type-vector.h
   include/mp/type-void.h
-  include/mp/util.h)
+  include/mp/util.h
+  include/mp/version.h)
```
hebasto, line 181: Agreed. In the Bitcoin Core repo:
```console
$ git grep "mp/version.h" -- ":(exclude)src/ipc/libmultiprocess/"
src/ipc/util.h:#include <mp/version.h>
```

CMakeLists.txt
```diff
@@ -174,9 +174,11 @@ set(MP_PUBLIC_HEADERS
   include/mp/type-struct.h
   include/mp/type-threadmap.h
   include/mp/type-tuple.h
+  include/mp/type-unordered-set.h
```
hebasto, line 177: This header doesn't seem to be used in Bitcoin Core outside of `src/ipc/libmultiprocess/`:
```console
$ git grep "mp/type-unordered-set.h" -- ":(exclude)src/ipc/libmultiprocess/" | wc -l
0
```

However, it's better to treat it like the other `mp/type-*.h` headers. This line effectively amends https://github.com/bitcoin-core/libmultiprocess/pull/277.

hebasto (approved):
ACK 2448d282ccdbbe739d097463c070a1a89dc21cd2.

PR #352, opened 2026-08-20
ci: add cmake debug output

ryanofsky:
Show detailed debug information if cmake fails in CI

### DIFF
--- a/ci/scripts/ci.sh
+++ b/ci/scripts/ci.sh
@@ -42,7 +42,22 @@ mkdir -p "$CI_DIR"
 cd "$CI_DIR"
 export CMAKE_BUILD_PARALLEL_LEVEL="$(nproc)"
 git --no-pager log -1 || true
-cmake "$src_dir" "${CMAKE_ARGS[@]+"${CMAKE_ARGS[@]}"}"
+cmake_args=("${CMAKE_ARGS[@]+"${CMAKE_ARGS[@]}"}")
+if ! cmake "$src_dir" "${cmake_args[@]}"; then
+  # If cmake failed, try it again with debug options.
+  # Could add --trace / --trace-expand here too but they are very verbose.
+  cmake_args+=(--debug-output --debug-trycompile)
+  if ver_ge "$cmake_ver" "3.16"; then cmake_args+=(--log-level=DEBUG); fi
+  if ver_ge "$cmake_ver" "3.17"; then cmake_args+=(--debug-find); fi
+  cmake "$src_dir" "${cmake_args[@]}" || : "cmake exited with $?"
+  if ver_ge "$cmake_ver" "3.26"; then
+    cat CMakeFiles/CMakeConfigureLog.yaml || true
+  else
+    cat CMakeFiles/CMakeError.log CMakeFiles/CMakeOutput.log || true
+  fi
+  find . -ls || true
+  false
+fi
 if ver_ge "$cmake_ver" "3.15"; then
   cmake --build . -t "${BUILD_TARGETS[@]}" -- "${BUILD_ARGS[@]+"${BUILD_ARGS[@]}"}"
 else


ci/scripts/ci.sh
```diff
@@ -25,7 +25,16 @@ src_dir=$PWD
 mkdir -p "$CI_DIR"
 cd "$CI_DIR"
 export CMAKE_BUILD_PARALLEL_LEVEL="$(nproc)"
-cmake "$src_dir" "${CMAKE_ARGS[@]+"${CMAKE_ARGS[@]}"}"
+cmake_args=("${CMAKE_ARGS[@]+"${CMAKE_ARGS[@]}"}")
+if ! cmake "$src_dir" "${cmake_args[@]}"; then
+  # If cmake failed, try it again with debug options.
+  # Could add --trace / --trace-expand here too but they are very verbose.
+  cmake_args+=(--debug-find --debug-output --debug-trycompile --log-level=DEBUG)
+  cmake "$src_dir" "${cmake_args[@]}" || : "cmake exited with $?"
```
maflcko, line 52:
```suggestion
  cmake "$src_dir" "${cmake_args[@]}" || echo "cmake exited with $?"
```

nit: I know this is duplicate, but this way non-Bash people won't have to confirm this works:

```bash
# set -o xtrace
false || : "command exited with $?"
set +o xtrace
+ false
+ : 'command exited with 1'
+ set +o xtrace

**Review by maflcko — COMMENTED** [review]
lgtm ACK 6d719cf7ccb1cd6d958898cc11db528a334b9d39

File: ci/scripts/ci.sh [hunk 2]
```diff
@@ -25,7 +25,16 @@ src_dir=$PWD
 mkdir -p "$CI_DIR"
 cd "$CI_DIR"
 export CMAKE_BUILD_PARALLEL_LEVEL="$(nproc)"
-cmake "$src_dir" "${CMAKE_ARGS[@]+"${CMAKE_ARGS[@]}"}"
+cmake_args=("${CMAKE_ARGS[@]+"${CMAKE_ARGS[@]}"}")
+if ! cmake "$src_dir" "${cmake_args[@]}"; then
+  # If cmake failed, try it again with debug options.
+  # Could add --trace / --trace-expand here too but they are very verbose.
+  cmake_args+=(--debug-find --debug-output --debug-trycompile --log-level=DEBUG)
```
ViniciusCestarii: Since this project supports cmake 3.12+ it would be good to guard these flags against a minimum version otherwise the command will just fail for these lower versions.

if ver_ge "$cmake_ver" "3.16"; then cmake_args+=(--log-level=DEBUG); fi
if ver_ge "$cmake_ver" "3.17"; then cmake_args+=(--debug-find); fi

ViniciusCestarii (commented):
ACK 6d719cf7ccb1cd6d958898cc11db528a334b9d39

Looks good, just added a comment about cmake support for the flags used.

ci/scripts/ci.sh
```diff
@@ -25,7 +25,16 @@ src_dir=$PWD
...
+  cmake "$src_dir" "${cmake_args[@]}" || : "cmake exited with $?"
+  cat CMakeFiles/CMakeConfigureLog.yaml || true
```
hebasto: The `CMakeFiles/CMakeConfigureLog.yaml` is [available](https://cmake.org/cmake/help/latest/release/3.26.html) only in CMake >=3.26.

ci/scripts/ci.sh
hebasto, line 52: The second `cmake` invocation is reusing `CMakeCache.txt`, skipping the checks whose results are cached. Is this the desired behaviour?

ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/352#discussion_r3827880236

From my perspective, the suggested change makes output more verbose, and I don't know what problem it would be solving. I'd definitely change this if it looked misleading, but it seems pretty obvious this is trying to show an error message. I also don't know why someone who didn't know bash would worry about this. It should take only few seconds to look up how this works if anyone is worried.

ci/scripts/ci.sh
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/352#discussion_r3838449747

> The `CMakeFiles/CMakeConfigureLog.yaml` is [available](https://cmake.org/cmake/help/latest/release/3.26.html) only in CMake >=3.26.

Thanks, this good to know. The `|| true` should let the script continue if the file isn't created for any other reason as well. But it could be a good idea to add a ver_ge check like the ones ViniciusCestarii suggested above

ci/scripts/ci.sh
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/352/changes#r3830510005

> Since this project supports cmake 3.12+ it would be good to guard these flags against a minimum version otherwise the command will just fail for these lower versions.

Agree these would be nice changes and I'd be happy to review a followup adding them.

ci/scripts/ci.sh
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/352#discussion_r3838465485

> The second `cmake` invocation is reusing `CMakeCache.txt`, skipping the checks whose results are cached. Is this the desired behaviour?

I think there are tradeoffs. I'd presume more likely than not if if results were cached they were probably successful results, so the current change lets the script be faster and simpler and show strictly more information than it did previously. It could be a good idea to delete the cache or build directory though, and I'd happy review if someone wanted to implement this followup.

ci/scripts/ci.sh
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/352#discussion_r3830510005

> Since this project supports cmake 3.12+ it would be good to guard these flags against a minimum version otherwise the command will just fail for these lower versions.

Thanks! Added this change in latest rebase

ci/scripts/ci.sh
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/352#discussion_r3838449747

> The `CMakeFiles/CMakeConfigureLog.yaml` is [available](https://cmake.org/cmake/help/latest/release/3.26.html) only in CMake >=3.26.

Thanks! Added guard in latest rebase

ci/scripts/ci.sh
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/352#discussion_r3838465485

> The second cmake invocation is reusing CMakeCache.txt, skipping the checks whose results are cached.

Decided not to implement a change here because erasing CMakeCache.txt could make failures harder to debug locally and would not be respecting the `CI_CLEAN` option. I think it could be reasonable to erase the cache (or entire build directory) after a failure when `CI_CLEAN` is true and to set `CI_CLEAN` in CI jobs, but erasing things by default when these scripts are run locally seems unsafe and inconvenient, and value of extra debug output that would be provided seems low.

ryanofsky (commented):
Rebased 6d719cf7ccb1cd6d958898cc11db528a334b9d39 -> fbd50d2f5dd2ed5311ba457ea9de57322da9b0b2 ([`pr/cmake-debug.1`](https://github.com/ryanofsky/libmultiprocess/commits/pr/cmake-debug.1) -> [`pr/cmake-debug.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/cmake-debug.2), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/cmake-debug.1-rebase..pr/cmake-debug.2)) due to conflict with #212, also implementing review suggestions.

re: https://github.com/bitcoin-core/libmultiprocess/pull/352#issuecomment-5426559226

> Seeing Claude being added as co-author here in the force-push, makes me a bit [flabbergasted](https://github.com/bitcoin/bitcoin/pull/35794#issue-4970398782), but I guess this is the thing now.

Sorry about that and I sympathize but I instantly muted 35794 as soon as I saw the title because I thought it would be too distracting for me, so I am not caught up right now. I do plan to read the discussion at some point and would like to talk about it in person. Appreciate you reviewing and approving despite this.

ci/scripts/ci.sh
```diff
@@ -42,7 +42,18 @@ mkdir -p "$CI_DIR"
...
-cmake "$src_dir" "${CMAKE_ARGS[@]+"${CMAKE_ARGS[@]}"}"
+cmake_args=("${CMAKE_ARGS[@]+"${CMAKE_ARGS[@]}"}")
+if ! cmake "$src_dir" "${cmake_args[@]}"; then
+  # If cmake failed, try it again with debug options.
+  # Could add --trace / --trace-expand here too but they are very verbose.
+  cmake_args+=(--debug-output --debug-trycompile)
+  if ver_ge "$cmake_ver" "3.16"; then cmake_args+=(--log-level=DEBUG); fi
+  if ver_ge "$cmake_ver" "3.17"; then cmake_args+=(--debug-find); fi
+  cmake "$src_dir" "${cmake_args[@]}" || : "cmake exited with $?"
+  if ver_ge "$cmake_ver" "3.26"; then cat CMakeFiles/CMakeConfigureLog.yaml || true; fi
```
hebasto, line 53:
```suggestion
  if ver_ge "$cmake_ver" "3.26"; then
    cat CMakeFiles/CMakeConfigureLog.yaml || true
  else
    cat CMakeFiles/CMakeError.log CMakeFiles/CMakeOutput.log || true
  fi
```

hebasto (approved):
ACK fbd50d2f5dd2ed5311ba457ea9de57322da9b0b2.

ryanofsky (commented):
Updated fbd50d2f5dd2ed5311ba457ea9de57322da9b0b2 -> a7ff9d5da2290c1edc08460262c909f31cff938a ([`pr/cmake-debug.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/cmake-debug.2) -> [`pr/cmake-debug.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/cmake-debug.3), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/cmake-debug.2..pr/cmake-debug.3)) applying CMakeError.log / CMakeOutput.log suggestion

hebasto (approved):
re-ACK a7ff9d5da2290c1edc08460262c909f31cff938a.

PR #353, opened 2026-08-20
ci: add cmake --parallel build option

ryanofsky:
Drop nproc call and inconsistent -j setting.

### DIFF
--- a/ci/scripts/ci.sh
+++ b/ci/scripts/ci.sh
@@ -24,15 +24,14 @@ ver_ge() { [ "$(printf '%s\n' "$2" "$1" | sort -V | head -n1)" = "$2" ]; }
 src_dir=$PWD
 mkdir -p "$CI_DIR"
 cd "$CI_DIR"
-export CMAKE_BUILD_PARALLEL_LEVEL="$(nproc)"
 cmake "$src_dir" "${CMAKE_ARGS[@]+"${CMAKE_ARGS[@]}"}"
 if ver_ge "$cmake_ver" "3.15"; then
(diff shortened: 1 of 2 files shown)


ci/scripts/ci.sh
```diff
@@ -24,15 +24,14 @@ ver_ge() { [ "$(printf '%s\n' "$2" "$1" | sort -V | head -n1)" = "$2" ]; }
 src_dir=$PWD
 mkdir -p "$CI_DIR"
 cd "$CI_DIR"
-export CMAKE_BUILD_PARALLEL_LEVEL="$(nproc)"
 cmake "$src_dir" "${CMAKE_ARGS[@]+"${CMAKE_ARGS[@]}"}"
 if ver_ge "$cmake_ver" "3.15"; then
-  cmake --build . -t "${BUILD_TARGETS[@]}" -- "${BUILD_ARGS[@]+"${BUILD_ARGS[@]}"}"
+  cmake --build . --parallel -t "${BUILD_TARGETS[@]}" -- "${BUILD_ARGS[@]+"${BUILD_ARGS[@]}"}"
```
hebasto, line 29: From the CMake [docs](https://cmake.org/cmake/help/v3.12/manual/cmake.1.html#build-tool-mode):
> If `<jobs>` is omitted the native build tool’s default number is used.

I don't think this change is an improvement, as it might reduce build parallelism.

maflcko: It could also lead to over-parallelism and lead to OOM?

PR #354, opened 2026-08-20
cmake: make Threads package optional

ryanofsky:
Make `find_package(Threads)` optional because there are platforms where this package may not be required, and because `find_package(Threads REQUIRED)` errors obscure more detailed error messages and make issues harder to debug.

There errors can happen on different platforms with different cmake policy settings. See commit message for details.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,13 @@ endif()
 
 include("cmake/compat_find.cmake")
 
-find_package(Threads REQUIRED)
+add_library(mpdeps INTERFACE)
+
+find_package(Threads)
+if(Threads_FOUND)
+  target_link_libraries(mpdeps INTERFACE Threads::Threads)
+endif()
+
 find_package(CapnProto 0.9 NO_MODULE)
  if(NOT CapnProto_FOUND)
    message(FATAL_ERROR
@@ -206,7 +212,7 @@ target_link_libraries(mpgen PRIVATE CapnProto::capnp)
 target_link_libraries(mpgen PRIVATE CapnProto::capnpc)
 target_link_libraries(mpgen PRIVATE CapnProto::kj)
 target_link_libraries(mpgen PRIVATE CapnProto::kj-async)
-target_link_libraries(mpgen PRIVATE Threads::Threads)
+target_link_libraries(mpgen PRIVATE mpdeps)
 # Use CAPNP_EXECUTABLE/CAPNPC_CXX_EXECUTABLE cmake variables set by
 # CapnProtoConfig.cmake rather than reading the CapnProto::capnp_tool
 # IMPORTED_LOCATION directly. On Debian/Ubuntu, IMPORTED_LOCATION is broken
--- a/cmake/pthread_checks.cmake
+++ b/cmake/pthread_checks.cmake
@@ -9,7 +9,9 @@ include(CMakePushCheckState)
 include(CheckCXXSourceCompiles)
 
 cmake_push_check_state()
-set(CMAKE_REQUIRED_LIBRARIES Threads::Threads)
+if(TARGET Threads::Threads)
+  set(CMAKE_REQUIRED_LIBRARIES Threads::Threads)
+endif()
 check_cxx_source_compiles("
   #include <pthread.h>
   int main(int argc, char** argv)
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -9,20 +9,20 @@ add_executable(mpcalculator
 )
 target_capnp_sources(mpcalculator ${CMAKE_CURRENT_SOURCE_DIR} init.capnp calculator.capnp printer.capnp)
 target_include_directories(mpcalculator PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
-target_link_libraries(mpcalculator PRIVATE Threads::Threads)
+target_link_libraries(mpcalculator PRIVATE mpdeps)
 
 add_executable(mpprinter
   printer.cpp
 )
 target_capnp_sources(mpprinter ${CMAKE_CURRENT_SOURCE_DIR} init.capnp calculator.capnp printer.capnp)
 target_include_directories(mpprinter PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
-target_link_libraries(mpprinter PRIVATE Threads::Threads)
+target_link_libraries(mpprinter PRIVATE mpdeps)
 
 add_executable(mpexample
   example.cpp
 )
 target_capnp_sources(mpexample ${CMAKE_CURRENT_SOURCE_DIR} init.capnp calculator.capnp printer.capnp)
 target_include_directories(mpexample PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
-target_link_libraries(mpexample PRIVATE Threads::Threads)
+target_link_libraries(mpexample PRIVATE mpdeps)
 
 add_custom_target(mpexamples DEPENDS mpexample mpcalculator mpprinter)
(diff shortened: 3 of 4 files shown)


CMakeLists.txt
```diff
@@ -12,7 +12,13 @@ endif()
 
 include("cmake/compat_find.cmake")
 
-find_package(Threads REQUIRED)
+add_library(mpdeps INTERFACE)
+
+find_package(Threads)
+if(Threads_FOUND)
+  target_link_libraries(mpdeps INTERFACE Threads::Threads)
+endif()
```
hebasto, line 20: > Make `find_package(Threads)` optional because there are platforms where this package may not be required...

But the code does not look platform-dependent.

According to the [docs](https://cmake.org/cmake/help/latest/module/FindThreads.html), `FindThreads` is capable of figuring out:
> ... if the thread functions are provided by the system libraries and no special flags are needed to use them.

Therefore, I don't see this change is necessary.

ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/354#discussion_r3838424965

> According to the [docs](https://cmake.org/cmake/help/latest/module/FindThreads.html), `FindThreads` is capable of figuring out

The docs are wrong. The REQUIRED keyword here hides useful error messages and replaces them with misleading "Could NOT find Threads" messages. This has happened to me on multiple occasions as described in the commit message and misled me and wasted my time.

If you think there is an advantage `REQUIRED` provides here, it would be good to know what it is. The only thing I see REQUIRED doing here is replacing real errors with fake and misleading ones. It seems clear to me `find_package(Threads)` is a helpful way to add thread dependencies and `find_package(Threads REQUIRED)` is a footgun.

re: https://github.com/bitcoin-core/libmultiprocess/pull/354#issuecomment-5366034482

> This won't happen in reality on current master

That's a bold prediction! It does not happen on current master, but it happened to me twice on different occasions changing the policy version and can easily happen with changes to platforms, toolchain files or any of the spaghetti code module files distributed with cmake.

purpleKarrot: > The docs are wrong.

Is there something we can report upstream? Can you share the situation and output of the suppressed diagnostics?

hebasto (commented):
I'm still not convinced that trading code correctness for easier debugging is an improvement.

In the following example, the build fails during the linking stage because it doesn't configure the thread library properly:
```console
$ cmake -B build -DCMAKE_CXX_COMPILER=g++-10
-- The CXX compiler identification is GNU 10.5.0
-- Check for working CXX compiler: /usr/bin/g++-10
-- Check for working CXX compiler: /usr/bin/g++-10 -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for C++ include pthread.h
-- Looking for C++ include pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - not found
-- Check if compiler accepts -pthread
-- Check if compiler accepts -pthread - no
-- Could NOT find Threads (missing: Threads_FOUND) 
-- Performing Test HAVE_PTHREAD_GETNAME_NP
-- Performing Test HAVE_PTHREAD_GETNAME_NP - Failed
-- Performing Test HAVE_PTHREAD_SETNAME_NP_3ARG
-- Performing Test HAVE_PTHREAD_SETNAME_NP_3ARG - Failed
-- Performing Test HAVE_PTHREAD_THREADID_NP
-- Performing Test HAVE_PTHREAD_THREADID_NP - Failed
-- Performing Test HAVE_PTHREAD_GETTHREADID_NP
-- Performing Test HAVE_PTHREAD_GETTHREADID_NP - Failed
-- Configuring done
-- Generating done
-- Build files have been written to: /libmultiprocess/build
$ cmake --build build          
Scanning dependencies of target mputil
[ 14%] Building CXX object CMakeFiles/mputil.dir/src/mp/util.cpp.o
[ 14%] Built target mputil
Scanning dependencies of target mpgen
[ 28%] Building CXX object CMakeFiles/mpgen.dir/src/mp/gen.cpp.o
[ 42%] Linking CXX executable mpgen
/usr/bin/ld: cannot find -lpthread
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/mpgen.dir/build.make:92: mpgen] Error 1
make[1]: *** [CMakeFiles/Makefile2:216: CMakeFiles/mpgen.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

CMakeLists.txt
ryanofsky: Thanks for the replies! I will close this PR because as obnoxious as `find_package(Threads REQUIRED)` behavior is for producing misleading error and hiding useful debug information, the `REQUIRED` keyword actually does seem to be used in bitcoin core and leveldb, so maybe it is not as much as a problem for others as it has been for me. And there could be other ways of avoiding harms like #352, or maybe moving the `find_package` call later so more useful errors can be triggered earlier. When I first made this change a year ago as part of #209, I didn't know techniques for debugging `find_package` errors and found this very confusing, but now I won't have this problem anymore.

Sorry for pushing back earlier. I pushed back because review comments were arguing against removing `REQUIRED` without acknowledging the harms it caused or pointing to any benefits it offered. But thinking about it more, one real benefit that `REQUIRED` offers is more predictability and less variance between builds. For example with 45c7dca6a55c2892d2b1b177123a3dea71d15090, if libmultiprocess is built on two openbsd systems, and one has `clang-scan-deps` installed and one doesn't, `find_package` may succeed on one system and fail on the other causing the library to be built with slightly different threading flags even if both builds succeed and are usable.

I still think 45c7dca6a55c2892d2b1b177123a3dea71d15090 is a correct change, just that benefits may not outweigh this cost. It is correct because the cmake [`FindThreads`](https://github.com/Kitware/CMake/blob/master/Modules/FindThreads.cmake) module is not a hard dependency of libmultiprocess (unlike Cap'n Proto). `FindThreads` is just a convenience module provided with cmake to help with portability and it is perfectly possible to write multithreaded code without it, avoiding its warts and bugs.

---

re: purpleKarrot https://github.com/bitcoin-core/libmultiprocess/pull/354#discussion_r3860478031

> > The docs are wrong.
>
> Is there something we can report upstream? Can you share the situation and output of the suppressed diagnostics?

I encountered these errors a year ago but you can see the error output in 45c7dca6a55c2892d2b1b177123a3dea71d15090 commit message and the `CMAKE_CXX_SCAN_FOR_MODULES` bug is reported upstream https://gitlab.kitware.com/cmake/cmake/-/work_items/27228. I described the bad `CMAKE_TRY_COMPILE_NO_PLATFORM_VARIABLES`/  `FindThreads` interaction in https://github.com/bitcoin-core/libmultiprocess/pull/163#issuecomment-3298814132 with steps to reproduce and you actually replied to that and seemed to think it was the same underlying bug.

---

re: maflcko https://github.com/bitcoin-core/libmultiprocess/pull/354#issuecomment-5421805354

> Ok, what I wanted to say was that this doesn't practically happen in reality. Someone changing the policy isn't something a real end-user does, but more a dev-only thing.

I don't think this is true because these errors depend on individual platform configurations. The same policies can work on one platform and cause problems on other platforms and this is exactly what we seen with `FindThreads` where the bug isn't even really triggered by the platform, but just by what packages are currently installed. Better error output is better for everyone and is not just a temporary development need.

---

re: hebasto https://github.com/bitcoin-core/libmultiprocess/pull/354#pullrequestreview-5029651423

> In the following example, the build fails during the linking stage because it doesn't configure the thread library properly:

That example seems to be artificial (setup is not described) unlike the real examples in the commit message, and even in that case `cannot find -lpthread` seems like a much better error than `Could NOT find Threads (missing: Threads_FOUND)` would be because it points to a clear cause.

PR #355, opened 2026-08-20
cmake: Disable automatic C++20 module dependency scanning

ryanofsky:
Set `CMAKE_CXX_SCAN_FOR_MODULES` off because this project doesn't use C++ modules, and they are enabled by default in newer versions of cmake, which can lead to failures trying to invoke the `clang-scan-deps` tool if it is not installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,11 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
   set(CMAKE_CXX_STANDARD_REQUIRED YES)
 endif()
 
+# Disable automatic C++20 module dependency scanning.
+# CMake >=3.28 tries to use `clang-scan-deps` by default, which may not
+# be installed on all platforms. We don't use named modules, so turn this off.
+set(CMAKE_CXX_SCAN_FOR_MODULES OFF)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -10,6 +10,11 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
   set(CMAKE_CXX_STANDARD_REQUIRED YES)
 endif()
 
+# Disable automatic C++20 module dependency scanning.
+# CMake >=3.28 tries to use `clang-scan-deps` by default, which may not
+# be installed on all platforms. We don't use named modules, so turn this off.
+set(CMAKE_CXX_SCAN_FOR_MODULES OFF)
```
hebasto, line 16: ~I tend to agree with @purpleKarrot and other CMake developers that a project should avoid forcibly setting the `CMAKE_*` variables as much as possible.~

As far as I understand this is a workaround for an upstream bug reported [here](https://gitlab.kitware.com/cmake/cmake/-/work_items/27228). 

~I think the proper workaround is to modify the CI scripts without touching the project's CML files.~

~If "we don't use named modules", we shouldn't worry about them at all rather than explicitly disabling them.~

PR #356, opened 2026-08-26
ci: Remove hard-coded -j4 from sanitize config

maflcko:
Not needed after CMAKE_BUILD_PARALLEL_LEVEL is set to nproc.


hebasto (approved):
ACK fa30e2093c7574f2fb8fe8f2eecd550927a5ae47.

PR #358, opened 2026-08-26
refactor: Enable readability-container-contains

maflcko:
This is enabled downstream, so should be fine to enable here as well.


hebasto (approved):
ACK fa101113b5097f6fabc7eedf2799f76e4503fb5d.

PR #360, opened 2026-09-01
pull latest .clang-tidy from downstream

maflcko:
Seems fine to do after https://github.com/bitcoin-core/libmultiprocess/pull/358, given that it is passing CI


hebasto (approved):
ACK fa1db7a9e3ed4b42c84c38cbc795fd1818829778.

PR #336, opened 2026-08-07
proxy-io: Reference-count Connection objects

ryanofsky:
Use reference counting to manage Connection object lifetimes. This implements an old idea from https://github.com/bitcoin-core/libmultiprocess/issues/176#issuecomment-2945442119 and has two benefits:

- Makes it possible to wait for objects associated with a Connection to be freed, to support unclean shutdowns better. This was implemented in base PR #335 for server objects, and this PR extends it to treat client and server objects symmetrically.
- Allows dropping the cleanup handlers `ProxyClient` objects register with Connections, so `Connection` objects no longer need to store lists of `ProxyClient` objects and can just use use counts instead.

---

**This is based on #335 + #365.** The non-base commits are:

- [`c092c18` proxy-io: manage Connection lifetime with shared_ptr](https://github.com/bitcoin-core/libmultiprocess/pull/336/commits/c092c18dc9df4789e1264b6223b6e61249309661)
- [`7635c12` proxy-io: keep client capability handles across disconnect](https://github.com/bitcoin-core/libmultiprocess/pull/336/commits/7635c126d67d5dac75f2d67dbc7fe9ab05c92375)
- [`162b7a6` doc: scope Connection cleanup callbacks to per-thread map entries](https://github.com/bitcoin-core/libmultiprocess/pull/336/commits/162b7a6ffeb218be39bd4ce34bd368d3f49b8752)
- [`f0a629d` proxy-io: make server object tracker a plain Connection member](https://github.com/bitcoin-core/libmultiprocess/pull/336/commits/f0a629d1f780118a69278cabd22285971e827021)
- [`6c7f1bf` ci: Check out bitcoin/bitcoin PR #35932 instead of master](https://github.com/bitcoin-core/libmultiprocess/pull/336/commits/6c7f1bfc86cb4d2c7c53d145eaf1a8122f64864d)

PR #361, opened 2026-09-04
proxy-io: Fix theoretical disconnect bugs

ryanofsky:
Two bugs in `Connection::onDisconnect` were pointed out in #335 review by @enirox001:

- [ListenConnections failing to accept new connections](https://github.com/bitcoin-core/libmultiprocess/pull/335#discussion_r3821865611) if max connections limit was reached and a disconnect of an existing connection was initiated locally, rather than remotely.
- [Race condition in Connection::onDisconnect](https://github.com/bitcoin-core/libmultiprocess/pull/335#discussion_r3821831654) implementation caused by using `evalLater` could allow an onDisconnect handler to run after a Connection object was destroyed.

A third "thread map teardown race" bug was also encountered in a new unit test introduced by #335. This PR fixes each bug in a separate commit and also includes documentation commits to help make the changes more understandable.

The fixed bugs are "theoretical" just in the sense that they haven't been seen in practice and were found in code review. The first ListenConnections bug can't currently happen in bitcoin core because it doesn't disconnect IPC clients except when it is shutting down, and it would not make sense to accept new connections. The race condition bugs have just existed for many years and not been seen previously.


include/mp/proxy-io.h
```diff
@@ -615,7 +621,7 @@ ProxyClientBase<Interface, Impl>::ProxyClientBase(typename Interface::Client cli
         m_context.loop->sync([&] {
             EventLoop& loop = *m_context.loop;
             Connection* connection = m_context.connection;
-            connection->onDisconnect([&loop, connection] {
+            connection->onRemoteDisconnect([&loop, connection] {
```
enirox001, line 624: In commit https://github.com/bitcoin-core/libmultiprocess/pull/361/changes/c5f093e48c0fd263fec82b84d83c6f8bb92abe39: _proxy-io: rename Connection disconnect handlers to reflect scope_

Good call on changing the name of the methods to match what they actually run, it a bit weird seeing a familiar name reused for a different method, but should become familiar soon enough 

In proxy.cpp there are some documentation that still references the previous method behaviours, here is a diff that would update them

```diff --git a/src/mp/proxy.cpp b/src/mp/proxy.cpp
index afb02ce..e77c836 100644
--- a/src/mp/proxy.cpp
+++ b/src/mp/proxy.cpp
@@ -111,7 +111,7 @@ Connection::~Connection() noexcept(false)
     // Connection destructor is always called on the event loop thread. If this
     // is a local disconnect, it will trigger I/O, so this needs to run on the
     // event loop thread, and if there was a remote disconnect, this is called
-    // by an onDisconnect callback directly from the event loop thread.
+    // by an onRemoteDisconnect callback directly from the event loop thread.
     assert(std::this_thread::get_id() == m_loop->m_thread_id);

     // Try to cancel any calls that may be executing.
@@ -187,14 +187,14 @@ Connection::~Connection() noexcept(false)
     // ProxyServer object destructors first, and then trigger an onDisconnect
     // callback.
     //
-    // On incoming side of the connection, the onDisconnect callback is written
+    // On incoming side of the connection, the onRemoteDisconnect callback is written
     // to delete the Connection object from the m_incoming_connections and call
-    // this destructor which calls Connection::disconnect.
+    // this destructor.
     //
     // On the outgoing side, the Connection object is owned by top level client
-    // object client, which onDisconnect handler doesn't have ready access to,
-    // so onDisconnect handler just calls Connection::disconnect directly
-    // instead.
+    // object client, which onRemoteDisconnect handler doesn't have ready access to,
+    // so onRemoteDisconnect handler just deletes the Connection object directly
+    // instead
     //
     // Either way disconnect code runs in the event loop thread and called both
     // on clean and unclean shutdowns. In unclean shutdown case when the
```

include/mp/proxy-io.h
```diff
@@ -899,10 +899,17 @@ void _Serve(EventLoop& loop, kj::Own<kj::AsyncIoStream>&& stream, InitImpl& init
...
-    it->onRemoteDisconnect([&loop, it, on_disconnect = std::forward<OnDisconnect>(on_disconnect)]() mutable {
+    // Run on_disconnect (e.g. the listener's active-connection counter
+    // decrement) on any disconnect. It is registered with onDisconnect rather
+    // than placed in the onRemoteDisconnect handler below because that handler
+    // only fires on a remote disconnect and is canceled when a connection is
+    // closed locally; if on_disconnect lived there, closing a connection
+    // locally would leave the listener's slot count stuck and stop it from
+    // accepting again.
+    it->onDisconnect(std::forward<OnDisconnect>(on_disconnect));
+    it->onRemoteDisconnect([&loop, it]() mutable {
```
enirox001, line 910: In commit https://github.com/bitcoin-core/libmultiprocess/pull/361/changes/bb21177965e363b32878f8258cc2bc1f9a8661f4: _proxy-io: fix listener stuck at capacity after a local disconnect
_

there is a documetation inconsitency though, it says in proxy.cpp in the `Connection::onDisconnect `method that "these callbacks only resets the connection pointers" but this commit adds a callback tthat also updates the listener count, so this description is a bit narrow

would suggest

```diff --git a/src/mp/proxy.cpp b/src/mp/proxy.cpp
index afb02ce..541948a 100644
--- a/src/mp/proxy.cpp
+++ b/src/mp/proxy.cpp
@@ -214,11 +214,10 @@ CleanupIt Connection::onDisconnect(std::function<void()> fn)
     // Add cleanup callbacks to the front of list, so sync cleanup functions run
     // in LIFO order. This is a good approach because sync cleanup functions are
     // added as client objects are created, and it is natural to clean up
-    // objects in the reverse order they were created. In practice, however,
-    // order should not be significant because the cleanup callbacks run
-    // synchronously in a single batch when the connection is broken, and they
-    // only reset the connection pointers in the client objects without actually
-    // deleting the client objects.
+    // objects in the reverse order they were created. The cleanup callbacks run
+    // synchronously on the event loop thread. They reset connection pointers in
+    // client objects without deleting the client objects, and update listener
+    // bookkeeping so freed connection slots can be reused.
     return m_sync_cleanup_fns.emplace(m_sync_cleanup_fns.begin(), std::move(fn));
 }
```

enirox001 (commented):
ACK bc98767dadb4d67b7cb6c87ca86696a7f888d715

Thnaks for opening this @ryanofsky, this does address the issues i mentioned, i agree these improvements here are more conservative as they have not been seen in practice, but would be nice to have it if we want to make disconnection more robust. 

Left some documentation nits

include/mp/proxy-io.h
```diff
@@ -422,9 +436,23 @@ struct Waiter
...
+    //! reverse order.
+    //!
+    //! Blocking rule: a thread other than the event loop thread must not hold
+    //! this mutex while calling EventLoop::sync() or EventLoop::post(). Those
+    //! calls block until the event loop thread runs the posted function, and
+    //! the event loop thread locks Waiter::m_mutex itself (in SetThread and
+    //! its disconnect callback, Waiter::post, and ~ProxyServer<Thread>), so
+    //! holding the mutex across the call could deadlock even though the two
+    //! mutexes would be locked in the permitted order. This is why
+    //! ~ThreadContext releases the mutex before destroying ProxyClient<Thread>
```
ryanofsky: >

include/mp/proxy-io.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/361#discussion_r3969587722

Thanks! I went though and updated all comments. I did wind up dropping the renames in this PR because it caused `test_bitcoin` to hang in https://github.com/bitcoin-core/libmultiprocess/actions/runs/33916092756, because bitcoin core tests are currently calling onDisconnect. So I moved the rename change to #336 and will be opening a new PR to simplify bitcoin core test code to use ConnectStream & ServeStream instead of low level onDisconnect code

include/mp/proxy-io.h
ryanofsky: re: https://github.com/bitcoin-core/libmultiprocess/pull/361#discussion_r3970151115

Nice suggestion, expanded the comment to mention the new change.

ryanofsky (commented):
Updated bc98767dadb4d67b7cb6c87ca86696a7f888d715 -> 0f5c633661c5068cfcb57dcdad21a1af6e6e275c ([`pr/ondis.1`](https://github.com/ryanofsky/libmultiprocess/commits/pr/ondis.1) -> [`pr/ondis.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/ondis.2), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/ondis.1..pr/ondis.2)) dropping renames because they cause problems downstream in `test_bitcoin`, making more documentation updates and adding a third bugfix here "thread map teardown race" that was originally in #335, since it fixes another bug that precedes #335

Updated 0f5c633661c5068cfcb57dcdad21a1af6e6e275c -> 040919b0b7b4855f2679bad80398cc5975d45155 ([`pr/ondis.2`](https://github.com/ryanofsky/libmultiprocess/commits/pr/ondis.2) -> [`pr/ondis.3`](https://github.com/ryanofsky/libmultiprocess/commits/pr/ondis.3), [compare](https://github.com/ryanofsky/libmultiprocess/compare/pr/ondis.2..pr/ondis.3)) to fix IWYU error https://github.com/bitcoin-core/libmultiprocess/actions/runs/34640736556/job/103399475344?pr=361

PR #362, opened 2026-09-09
Enable `-Wunused`

fanquake:
If merged, this would be used downstream in https://github.com/bitcoin/bitcoin/pull/36167.


maflcko:
Looks like this also exposes a build race in https://github.com/bitcoin-core/libmultiprocess/actions/runs/34482806488/job/102889597378?pr=362#step:6:384?

PR #364, opened 2026-09-10
util: Error out at compile time if KJ_NO_EXCEPTIONS is set

ryanofsky:
Cap'n Proto has a bug on netbsd where it incorrectly detects compiler does not support exceptions (reported and fixed in https://github.com/capnproto/capnproto/pull/2756) causing exceptions thrown from bitcoin core IPC methods not be caught and leading to the CI failure reported https://github.com/bitcoin/bitcoin/issues/36058.

Failure can be worked around by adding `-DKJ_NO_EXCEPTIONS=0`` to the build configuration so add an error message to detect when it would happen and suggest this.

### DIFF
--- a/include/mp/proxy-types.h
+++ b/include/mp/proxy-types.h
@@ -13,6 +13,12 @@
 #include <typeindex>
 #include <vector>
 
+// KJ exception support is required, or exceptions thrown in serverInvoke below go uncaught
+// and crash the process instead of being returned to the client as errors.
+#if KJ_NO_EXCEPTIONS
+#error "KJ_NO_EXCEPTIONS=1 is set but libmultiprocess requires support for exceptions. Please check build settings. On NetBSD you may also need to set -DKJ_NO_EXCEPTIONS=0 -DKJ_NO_RTTI=0 explicitly (https://github.com/capnproto/capnproto/pull/2756)"
+#endif
+
 namespace mp {
 
 template <typename Value>


include/mp/proxy-types.h
```diff
@@ -13,6 +13,12 @@
 #include <typeindex>
 #include <vector>
 
+// KJ exception support is required, or exceptions thrown in serverInvoke below go uncaught
+// and crash the process instead of being returned to the client as errors.
+#if KJ_NO_EXCEPTIONS
+#error "KJ_NO_EXCEPTIONS=1 is set but libmultiprocess requires support for exceptions. Please check build settings. On NetBSD you may also need to set -DKJ_NO_EXCEPTIONS=0 -DKJ_NO_RTTI=0 explicitly (https://github.com/capnproto/capnproto/pull/2756)"
```
hebasto, line 19: While the comment about `-DKJ_NO_RTTI=0` is technically correct, setting the flag isn't necessary in [practice](https://github.com/hebasto/bitcoin-core-nightly/blob/cec3b65eaf2e1e10e93bdb6ef13e5d358010ff9c/.github/workflows/netbsd.yml).

hebasto (approved):
ACK 4a1f400f8b7259db748b311c1441d375e68d76ca, I have reviewed the code and it looks OK.

include/mp/proxy-types.h
ryanofsky: > While the comment about `-DKJ_NO_RTTI=0` is technically correct, setting the flag isn't necessary in [practice](https://github.com/hebasto/bitcoin-core-nightly/blob/cec3b65eaf2e1e10e93bdb6ef13e5d358010ff9c/.github/workflows/netbsd.yml).

Thanks, I only saw the failing CI run in https://github.com/bitcoin/bitcoin/issues/36058 and wasn't sure what was needed to make CI pass. I believe KJ_NO_RTTI is good to set though to improve exception error messages.

ryanofsky (commented):
re: https://github.com/bitcoin-core/libmultiprocess/pull/364#issuecomment-5620341102

> Why not check earlier, at configure time?

This seems less reliable because the header is installed and can be compiled with different flags after being installed. I wouldn't object to another change adding a cmake check though if it would be helpful.

PR #365, opened 2026-09-11
proxy-io: Generalize ConnectStream / ServeStream for use in tests

ryanofsky:
Generalize `ConnectStream` and `ServeStream` functions to simplify test setup in libmultiprocess and bitcoin core code (in followup https://github.com/ryanofsky/bitcoin/commit/990f259a9dd60473d247218bf579c57893319ced). Also add an `incomingConnections` accessor so Bitcoin Core code does not need to directly access the incoming connections list to disconnect clients.

These changes help with #336 by getting rid of external code that would need to change with its changes to the `Connection` representation.


xyzconstant:
LGTM. ACK cb925202d26b6660211096b3377cce7e11cef9fe.

By the way, the llvm job failed with IWYU.

